Issue #50, opened 2018-10-27
The great refactor -- Refactor config to YAML, Add Docs, Revise exported fields.


Skarlso:
This is the first ticket tracking the first three points of the Great Refactor epic #49.

lastzero:
Concerning YAML config reader: I figured it is not really needed at this point, as the environment is configured via Docker env variables. Might be needed if installed without Docker (that's how we started). So I just left it like it is.

lastzero:
Config should use functions, not fields, to return values. It's on my personal to do list. Again, there should be good test coverage first.

lastzero:
For example, I started doing this so that not every single sub-directory needs to be defined in the config file (c.AssetsPath should be changed to c.assetsPath when refactoring is done):

```go
func (c *Config) GetAssetsPath() string {
	return c.AssetsPath
}

func (c *Config) GetTensorFlowModelPath() string {
	return c.GetAssetsPath() + "/tensorflow"
}
```

So now of course you could argue that all those paths could be properly populated upfront and the functions are not needed.

lastzero:
You suggest directly populating config via `json.Marshal()`? Is that a good idea, assuming you might want at least a little bit of abstraction and validation between the file and what you use internally? I see you can do it like that, I just want to think about it before we go that way (if that's what you suggest).

Skarlso:
So, Go isn't like that. Go is a very simplistic, C style language, with limited OOP capabilities. It love what's simple. Marshalling works in huge projects and in Go itself and it various other situations without making a getter setter. Getter setters are a reminiscence of Java, PHP and other languages who promote OOP. Go is more like C in this regard. As simple as possible don't think about it too much. :)

In the above example you provide GetAssetsPath would be a good thing only if it's doing something before accessing the property. Like, validation, or default setting, or something else. In this form is absolutely useless and clutters the code needlessly.

Let me ask you this. What would the abstraction be good for? What would it actually add in this situation? What do you expect to get out of a layer of indirection like that? 

If you want an abstraction just to have an abstraction you don't need the abstraction. :)

lastzero:
For me, Go is Oberon that looks like C. Using functions is not OOP.

The reason why I started to refactor this is that some config values had in fact be returned by functions like `GetTensorFlowModelPath()`, unless I want to put every single sub-directory in the config file (which also means the config file would change a lot and become big, fat and ugly).

[...]

PR #53, opened 2018-10-28
[DONE] Refactoring the photoprism package.

Skarlso:
This relates to #50.

Works still to be done:
- [x] Some files are left out yet and not finished I'm just sharing work
- [x] Revise the `util` file and throw it out
- [x] See other exported functions if they need to be exported
~- [ ] Revise the use of pointer structs instead of value. If they aren't changed there is no need for pointers to needlessly occupy heap.~ -- this would actually be a ton of work right now. I decided to leave that off to a different PR.


internal/photoprism/colors.go
```diff
@@ -1,12 +1,13 @@
 package photoprism
 
 import (
-	"github.com/RobCherry/vibrant"
-	"github.com/lucasb-eyer/go-colorful"
-	"golang.org/x/image/colornames"
 	"image"
 	"os"
 	"sort"
+
```
lastzero: Please make sure to run `make fmt` to avoid changes in the import order (should sort them alphabetically)

internal/photoprism/config.go
```diff
@@ -1,38 +1,45 @@
...
-	OriginalsPath  string
-	ImportPath     string
-	ExportPath     string
-	DarktableCli   string
-	DatabaseDriver string
-	DatabaseDsn    string
+	ServerPort     int    `yaml:"server-port"`
+	ServerMode     string `yaml:"server-mode"`
+	AssetsPath     string `yaml:"assets-path"`
+	ThumbnailsPath string `yaml:"thumbnails-path"`
```
lastzero: As discussed in #50, I think it's better to use functions in this particular case. I'll try to document that for others inline and/or in the developer guide. Thank you for the valuable feedback and suggestions!

internal/photoprism/colors.go
Skarlso: These import statements have been formatted by the Go standard, `goformat`. Fmt states this about it:

>gofmt follows a simple rule that works for all use cases: imports may be separated by
blank lines, and each set of blank lines is sorted independently.
goimports, which does more than just formatting, applies a different rule: it groups
imports according to whether they are standard imports, appengine imports, or other
imports.  It emits each group separately.  Then the gofmt algorithm kicks in and sorts
each group independently.
The goimports and gofmt programs do different jobs and it is not a goal that they
produce identical results.

Are you saying you don't want `goformat`? :)

Also, instead of a target your IDE should definitely take care of formatting rules as that's a much more convenient thing to have instead of running a make target.

lastzero: You mean `goimports`, not `goformat`? `make fmt` uses `go fmt`.

If `goimports` doesn't cause any issues (does it need installation? will GoLand remove the empty line again?), we can add it to the make target. Just want the imports to be the same for everyone.

lastzero: If everyone uses different IDE settings, merge conflicts and differently formatted code will be the result. That's why we use make. Of course if your IDE can do it the same, use it. Maybe I should also check my IDE settings... not a bad idea.

lastzero: Added it to `make fmt`, see 6a2af16d1cd556d4d25eb2c43c1cd43e9b9dad71 / #50

Skarlso: Interesting. I thought goland uses these tools by default. By the way that's why I like Go, formatting is a language standard and thus every tool from vimgo to goland should be applying them.

Yes, you are right. 😊 I meant goimports.

lastzero: No, you need to manually install `goimports` and enable it as a save hook. Should add that to the developer guide. Not a huge fan of extra tools in general: https://twitter.com/lastzero/status/1056992297135603712

Skarlso:
@lastzero I got the email but I can't see your comment any longer. :O

So, the util is usually anti pattern in a code. It's code no-one know where it should go. But these were usually functions that were used in either one place or actually belonged together like a file_checker. Which were two common functions that did file operations.

The rest was only used in one place and those belongs to that place. Also, some of these functions wasn't used at all, or only in the test which made no sense to me. It was dead code. So I deleted it. Like the random generator.

If you require it it's always under version control. :) That's why I removed it. :)

lastzero:
In final code it would be like that, I agree. The util file was a reminder to use a proper abstraction like spf13/afero later (and even that puts them in util.go...). Like many rough edges are actually reminders for something bigger, because they jump in your eye (like they did! see it worked...). Now the functions are distributed in different files :) But it's ok.

lastzero:
My focus will be to improve documentation and tests before I start adding new features. Very important when you work in a distributed team. When there are tests, you can always refactor later.

internal/photoprism/search.go
```diff
@@ -177,26 +182,31 @@ func (s *Search) Photos(form PhotoSearchForm) ([]PhotoSearchResult, error) {
...
+func (s *Search) FindFileByHash(fileHash string) (file models.File) {
 	s.db.Where("file_hash = ?", fileHash).First(&file)
 
 	return file
 }
 
-func (s *Search) FindPhotoById(photoId uint64) (photo Photo) {
-	s.db.Where("id = ?", photoId).First(&photo)
+// FindPhotoByID returns a Photo based on the ID.
+func (s *Search) FindPhotoByID(photoID uint64) (photo models.Photo) {
```
lastzero: Ah, didn't work because you changed the imports ;)

Skarlso:
@lastzero Cool stuff man! I take your view into account on the next changes! :) 👍

PR #73, opened 2018-11-24
Fixed chai tests

or29544:
Initial fix. Tests were comparing object instances instead of object properties.


lastzero:
frontend/tests/result.html should be removed from git (also add it to .gitignore please)

lastzero:
Why did you close it?

or29544:
Oops, must have been by mistake.

Issue #76, opened 2018-12-05
Photo like/dislike API must return OK (200) instead of Accepted (202)


lastzero:
DislikePhoto and LikePhoto should return `http.StatusOK` (200) instead of `http.StatusAccepted` (202) as requests are not handled asynchronously (StatusAccepted means the request was received, but the data will be changed later/asynchronously).

Code in the following two files needs to be changed:
- https://github.com/photoprism/photoprism/blob/develop/internal/api/photos.go
- https://github.com/photoprism/photoprism/blob/develop/internal/api/photos_test.go

PR #78, opened 2018-12-07
HTTP status code change

webrules:
This is the slight code change as mentioned in the issue https://github.com/photoprism/photoprism/issues/76 that 
DislikePhoto and LikePhoto should return http.StatusOK (200) instead of http.StatusAccepted (202)
I've executed "make test" without any problem.


lastzero:
Thank you 👍

PR #92, opened 2019-01-15
error chacks and minor api refactoring

vedhavyas:
- Add error checks
- return errors instead of log.Fatal
- minor refactor to apis


lastzero:
Nice, thank you 👍

Issue #82, opened 2018-12-19
Improve full-screen photo viewer (lightbox)


lastzero:
**As a user, I want to open photos in a full-screen overlay with lazy-loading and swipe support in order to view them quickly, easily and in best quality.**

Acceptance Criteria:
- [x] Use https://github.com/rap2hpoutre/vue-picture-swipe (unless you find a better solution)
- [x] Clicking / touching a tile should open the full-screen viewer with the image selected plus the others from the currently list (before/after)
- [x] Closing the viewer again must be easily possible on all devices (extra small close button is not user friendly, for example)
- [x] The viewer must look great on all devices from smartphones to Chrome, Firefox and Safari
- [x] The viewer should be able to display additional metadata (optional) and the photo title (must)
- [x] The viewer must be able to work completely full-screen (not just browser window size)
- [x] A thumbnail size that fits the screen size and resolution must be chosen by the viewer (like currently implemented, keep in mind Retina displays), see [Responsive Images](http://photoswipe.com/documentation/responsive-images.html)

ArmanFeyzi:
@lastzero I think [viewerjs](https://github.com/fengyuanchen/viewerjs) is the most powerful image viewer.
Here is a [demo](https://github.com/fengyuanchen/viewerjs) of Vue.js version.

**Features**
Supports 39 options
Supports 23 methods
Supports 9 events
Supports modal and inline modes
Supports touch
Supports move
Supports zoom
Supports rotation
Supports scale (flip)
Supports keyboard
Cross-browser support

lastzero:
@ArmanFeyzi Thanks for the hint! Strange that I didn't find it myself. Need to check how to integrate it with VueJS - somebody got experience with that?

skunert:
Can we close this?

PR #96, opened 2019-02-11
Integrate PhotoSwipe as photo viewer

skunert:
Hey,

in this PR I integrate PhotoSwipe as photo viewer as suggested in #82

I did not use the suggested vue wrapper for photoswipe since it did not offer the flexibility to modify the thumbnail style before click. So in the end I used the vanilla photoswipe library.

I know react but am new to vue, so if there are improvements, tell me.


lastzero:
Amazing, thank you very much! 👍

PR #103, opened 2019-04-10
Improve color indexing by using k-means clustering

skunert:
This is an implementation of the issue described in #7.

For the color indexing, I used the [prominentcolor](https://github.com/EdlinOrg/prominentcolor) library.
For the color palette, I used the material design palette as suggested in the open issue.

For now, I want to gather opinions if this solution is feasible, therefore I included a small frontend change that shows the colors in the details view, this frontend code should be removed before this PR is merged.


lastzero:
Amazing, thank you! Did you update the tests? We can merge this anytime you like 👍

skunert:
Did not update the tests yet, also need to remove the debugging code. Will udpate the PR soon :).

lastzero:
Before you do your final tests, make sure to merge develop and run `docker-compose build`. I've updated Go & TensorFlow.

skunert:
I rebased onto develop, added some tests and removed the debugging code. Until now, the color names where not really touched in these tests. I added some assertions.

From my site this is ready to merge.

lastzero:
I'll merge this, tests are green 🌱

lastzero:
This fails on develop:
```
Error Trace:	colors_slow_test.go:35
        	Error:      	Not equal: 
        	            	expected: []string{"black", "brown", "grey", "white"}
        	            	actual  : []string{"black", "brown", "grey"}
```

Any idea?

lastzero:
So the algorithm randomly returns different colors 🥳

AFK now... will have a look at it soonish.

skunert:
I will also take a look at it.

lastzero:
What happens if we just resize the image to 3x3 or 4x4 pixels and use those instead of random pixels of an 80x80 image? At least the result would always be the same and it should be good enough to detect a blue sky or a green forest.

PR #107, opened 2019-04-26
Fix broken tensorflow test

### LINKED ISSUE #104: Failing Tensorflow test TestTensorFlow_GetImageTags
When I execute `go test -tags slow -v ./internal/photoprism/...`, test `TestTensorFlow_GetImageTags` fails with the following message:

```
=== RUN   TestTensorFlow_GetImageTags
--- FAIL: TestTensorFlow_GetImageTags (2.59s)
    config.go:77: Initializing test data
    tensorflow_slow_test.go:33:
        	Error Trace:	tensorflow_slow_test.go:33
        	Error:      	Not equal:
        	            	expected: 0.1648176
        	            	actual  : 0.16481775
        	Test:       	TestTensorFlow_GetImageTags
```

I assume this is due to the recent tensorflow update. Project seems to be running fine on travis.

skunert:
This should fix #104. @lastzero In the issue you mentioned updating the travis config? What is there to be done? Do you mean that the travis build cache should be updated?

### DIFF
--- a/internal/photoprism/tensorflow_slow_test.go
+++ b/internal/photoprism/tensorflow_slow_test.go
@@ -4,6 +4,7 @@ package photoprism
 
 import (
 	"io/ioutil"
+	"math"
 	"testing"
 
 	"github.com/photoprism/photoprism/internal/test"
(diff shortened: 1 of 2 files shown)


lastzero:
I don't see this error and tests are broken on CI now (with the fix). Are you using the latest docker image for development? Maybe a different operating system or processor? Did you enable any hardware acceleration for TensorFlow?

lastzero:
Did you run `make all build test` or just `make test`?

lastzero:
We can also think about rounding. Maybe even use `int(Probability * 100)` which would make it easier to store and compare.

skunert:
I did run both commands, but encounter the same problem with `make all build test`. I just pushed a commit which introduces the rounding you suggested.

PR #114, opened 2019-06-01
add context

### LINKED ISSUE #112: Graceful shutdown of servers and other goroutines
**As a user, I want all goroutines to shut down gracefully so that I don't lose data when I quit PhotoPrism.**

See:
* https://github.com/photoprism/photoprism/blob/develop/internal/server/server.go#L40
* https://github.com/photoprism/photoprism/blob/develop/internal/tidb/server.go#L109

So far we have the web service and tidb running... in the future there might be goroutines for indexing / syncing as well... so we need a general approach that enables us to shut down all processes without losing data.

When this is done, we can start working on background processes for indexing / importing.

vedhavyas:
Closes #112

### DIFF
--- a/internal/api/api_test.go
+++ b/internal/api/api_test.go
@@ -13,9 +13,7 @@ func NewApiTest() (app *gin.Engine, router *gin.RouterGroup, conf *config.Config
 	conf = config.TestConfig()
 	gin.SetMode(gin.TestMode)
 	app = gin.New()
-
 	router = app.Group("/api/v1")
-
 	return app, router, conf
 }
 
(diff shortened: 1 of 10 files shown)


lastzero:
Thank you! Can't wait to test this at home. I have two short questions:

1. In `internal/tidb/server.go`line 105 `go runServer()` internally again uses two go funcs, so it looks double wrapped. Can we remove this if we assume the server doesn't use the MySQL X Protocol? We don't need it for our app.

2. Did you test if TiDB is still properly initialized if empty (no db / password)?

vedhavyas:
> In internal/tidb/server.goline 105 go runServer() internally again uses two go funcs, so it looks double wrapped.
Ahh, missed this.

> Did you test if TiDB is still properly initialized if empty (no db / password)?
I think yes, I just built the app and ran it.

lastzero:
Excellent work 👍 I'll merge this and test it again thoroughly while working on other issues...

PR #119, opened 2019-06-19
Feat/daemon

vedhavyas:
Run Photoprism in Daemon mode

### DIFF
--- a/internal/photoprism/openstreetmap.go
+++ b/internal/photoprism/openstreetmap.go
@@ -92,8 +92,8 @@ func (m *MediaFile) Location() (*models.Location, error) {
 	}
 
 	if len(openstreetmapLocation.Name) > 1 {
-		location.LocName = strings.ReplaceAll(openstreetmapLocation.Name, " - ", " / ")
-		location.LocName = util.Title(strings.TrimSpace(strings.ReplaceAll(location.LocName, "_", " ")))
+		location.LocName = strings.Replace(openstreetmapLocation.Name, " - ", " / ", -1)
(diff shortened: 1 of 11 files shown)


internal/photoprism/openstreetmap.go
```diff
@@ -106,11 +106,11 @@ func (m *MediaFile) Location() (*models.Location, error) {
 	location.LocCountryCode = strings.TrimSpace(openstreetmapLocation.Address.CountryCode)
 	location.LocDisplayName = strings.TrimSpace(openstreetmapLocation.DisplayName)
 
-	locationCategory := strings.TrimSpace(strings.ReplaceAll(openstreetmapLocation.Category, "_", " "))
+	locationCategory := strings.TrimSpace(strings.Replace(openstreetmapLocation.Category, "_", " ", -1))
 	location.LocCategory = locationCategory
 
 	if openstreetmapLocation.Type != "yes" && openstreetmapLocation.Type != "unclassified" {
-		locationType := strings.TrimSpace(strings.ReplaceAll(openstreetmapLocation.Type, "_", " "))
```
vedhavyas: ReplaceAll was added in 1.12. Anyone using Go version <1.12 will get build failure for develop

lastzero: Good to know! Shouldn't be an issue since we provide docker images containing all dependencies.

lastzero:
Looks good. Thank you very much! I'd change a few things to improve user experience... "daemon" might be too technical... maybe use "detach" instead? Or has that another, specific meaning I'm not aware of?

PR #120, opened 2019-06-22
(WIP) test on photos improvement

FrouxBY:
Hi, 
As part of my learning in Go and in this project, I try to improve the test,
This is a first shot for the photo API,
I have just a weird issue with the GetPhotos functions : 
When the test is done directly after a compilation, even if the database seems to be initialise (by AutoMigrate), the Search.Photo methods complain about a missing table.
`Error 1050: Table 'photos' already exists`

However, if this test is run after the photoprism package tests, which populate the database with a few example, then Search does not complain and the test pass. 

Despite this, does those test sounds ok ? And any idea for this gorm issues ?
Cheers


lastzero:
Amazing, thank you! I'll have a look ASAP, but I'm very busy right now - working on ARM64 / TensorFlow the last 5 days was a huge waste of time, but I really want to get PhotoPrism compiled...

Yes, database fixtures are still an issue and needs to be addressed fundamentally.

lastzero:
Didn't forget about this pull request! Just a bit tired after working on PhotoPrism for 10 hours today. Spent the last two weeks with compiling TensorFlow and helping others to do so.

Database fixtures and small (!) test images with metadata is something we also need for our acceptance tests, @graciousgrey will probably work on this. You can run our acceptance tests with `make acceptance` - server will start in the background (detached mode), so it should NOT already be running. Adding a `make testdata` target might make sense, need to think about it.

lastzero:
Also wondering if we should move API tests that expect state and test data to our acceptance test suite. API handler unit tests would be pretty basic then and would not require specific test data. Opinions?

FrouxBY:
Have been away, but back to continue those contribution,
I'm not sure wether API test might be move to acceptance, as unit testing here will test wether the api respond with a correct json or not (especially the json nested label structure). This should help to see later that a change in the database structure does not break the API.
This mean that API test could be done with minimal mock database and does not require a more complete database which include various kind of files, cameras, etc.

Am I correct or do i misunderstood the unit and acceptance test ?

lastzero:
Great that you're back :) We will implement a feature to reset the (test-)database to a defined state whenever needed, but it will take a couple of days as I'm busy with another project right now... your test will be a good example use case for this. In general, unit tests should require no state, so testing a database is always a bit awkward and complicated (if you want stable and fast tests at the same time).

FrouxBY:
I don't have time for now actually ,
and it will require a proper work with databases

Issue #131, opened 2019-08-18
Filename with brackets not supported


scilaci:
Hi guys,

I found a bug blocking the indexation process :'(

How to reproduce :

Rename the filename of a picture and use the brackets in the name:
- Ex: pictureFilename[1].jpg
 

Move the file in the Originals folder and try to reindex 

The indexation process is stopped by this error:

> photoprism_1  | time="2019-08-18T18:22:25Z" level=info msg="updated main jpg file \"2019/08/20190815_002342_D41083B2B2C2.jpg\""
photoprism_1  | [GIN] 2019/08/18 - 18:22:25 | 500 |  7.230782403s |    192.168.1.40 | POST     /api/v1/index
photoprism_1  | time="2019-08-18T18:22:25Z" level=info msg="/home/photoprism/Pictures/Originals/P1090443[1].jpg"
photoprism_1  |
photoprism_1  |
photoprism_1  | 2019/08/18 18:22:25 [Recovery] 2019/08/18 - 18:22:25 panic recovered:
photoprism_1  | POST /api/v1/index HTTP/1.1
....
photoprism_1  |
photoprism_1  |
photoprism_1  | runtime error: invalid memory address or nil pointer dereference
photoprism_1  | /usr/local/go/src/runtime/panic.go:82 (0xcc3f50)
photoprism_1  | /usr/local/go/src/runtime/signal_unix.go:390 (0xcc3d7f)
photoprism_1  | /go/src/github.com/photoprism/photoprism/internal/photoprism/mediafile.go:372 (0x1c9f5b6)
photoprism_1  | /go/src/github.com/photoprism/photoprism/internal/photoprism/mediafile.go:372 (0x1c9f5b1)
photoprism_1  | /go/src/github.com/photoprism/photoprism/internal/photoprism/indexer.go:111 (0x1c981e6)
photoprism_1  | /go/src/github.com/photoprism/photoprism/internal/photoprism/indexer.go:359 (0x1c9cbc0)
photoprism_1  | /go/src/github.com/photoprism/photoprism/internal/photoprism/indexer.go:397 (0x1cadbcf)
photoprism_1  | /usr/local/go/src/path/filepath/path.go:358 (0xd90c25)

Is a big problem when you are trying to index 50k of pictures and be blocked by one of these pictrues!

If you are tying to upload the picture from the UI and importing...
=> The file is never moved to Original folder and without error in the logfile

lastzero:
[...]

PR #134, opened 2019-08-27
Handle filenames with special characters in them #131

korimillivijay:
1. Catch and log unexpected errors due to failure of any individual files, but continue processing the remaining files
2. Escape filenames with special/meta characters so that they can match any related files without failures


lastzero:
Thank you so much!

PR #155, opened 2019-12-08
Frontend: fix lazy loading for albums

thielepaul:
just a really small fix to load only the images belonging to a specific album


lastzero:
Thank you, but what is  the bug? Seems to work for me...

Screenshot 2019-12-08 at 17 58 41

thielepaul:
When there are more than 60 photos in an album, the request to load more photos (triggered by scrolling down) does not contain the album id and therefore loads pictures that are not in the album.

PR #159, opened 2019-12-14
Select range of photos by pressing Shift

thielepaul:
photoprism-select-range


lastzero:
Amazing, thank you! Shouldn't the window.addEventListener also be removed when the component is destroyed?

thielepaul:
> Amazing, thank you! Shouldn't the window.addEventListener also be removed when the component is destroyed?

Yes, thanks!

frontend/src/component/p-photo-tiles.vue
```diff
@@ -70,6 +71,20 @@
             album: Object,
         },
         methods: {
+            selectRange(photo) {
+                var selection = this.$clipboard.getIds();
+                if (selection.length) {
+                    var lastAddedId = selection[selection.length - 1];
+                    var rangeStart = this.photos.findIndex((photo) => photo.getId() == lastAddedId);
```
lastzero: This might cause issues when you previously selected photos in a different view or in a different search result.

frontend/src/component/p-photo-tiles.vue
```diff
@@ -70,6 +71,20 @@
...
+                    var rangeEnd = this.photos.indexOf(photo);
+                    var photosToBeAdded = this.photos.slice(Math.min(rangeStart, rangeEnd), Math.max(rangeStart, rangeEnd) + 1);
+                    for (var photo of photosToBeAdded) {
```
lastzero: Better use `let` instead of `var` and a different variable name here (photo is a function argument already).

lastzero:
Do you have time to have a look at my comments? I'll otherwise merge this and refactor it if needed.

thielepaul:
> Do you have time to have a look at my comments? I'll otherwise merge this and refactor it if needed.

I agree with your comments. If you have time to solve them it would be nice! Should I give you write access to my fork so you can do it in the branch?

lastzero:
I can fix it on develop, only tested code is pushed to master (with a few exceptions we learn from).

lastzero:
Done ✅

PR #186, opened 2020-01-10
[Minor/Wording] Update photo.go url route docs

kelvinhammond:
Minor change.


lastzero:
Thanks! Probably not in use yet so it went unnoticed...

PR #191, opened 2020-01-14
Optimize EXIF integration

### LINKED ISSUE #172: Improve Exif data extraction from images
**As a user, I want fast Exif metadata extraction from images so that I can save time indexing large photo collections.**

See [Improve support for Exif metadata #13](https://github.com/photoprism/photoprism/issues/13),  https://github.com/photoprism/photoprism/issues/21#issuecomment-568534714 and https://github.com/photoprism/photoprism/issues/21#issuecomment-569498132:

> It looks like you're doing a brute-force search for the EXIF block:
> 
> https://github.com/photoprism/photoprism/blob/8e15c1d6fd4d1f82679d7b5f73dbe4c0823663ed/internal/photoprism/exif.go#L81
> 
> It would be quicker to try a context-specific method first (if a JPEG, try finding/parsing APP1 segment directly; if a PNG, try finding/parsing a "eXIf" chunk directly) and then fallback to the brute force method (only for file-formats that support them, currently JPG, TIF, XMP, PNG, and HEIF/HEIC). That said, is there a chance that you're doing this byte-by-byte search through formats that don't even support EXIF? I'd be worried about how much time we might be losing because at worst this can be very expensive and fruitless (like scanning RAW for EXIF?).

Acceptance Criteria:
- [ ] Exif extraction runs faster, ideally demonstrated by a Go test / benchmark
- [ ] Works with all file formats we support at the moment (or more)
- [ ] All currently supported Exif fields are supported (or more)
- [ ] Documentation / wiki updated (as needed)

Optional:
- [ ] Improve code quality and test coverage
- [x] Move (some) metadata code to a separate `meta` package if it makes sense
- [x] Take a look at `go-xmp` and figure out how it can be effectively used together with `go-exif`

dsoprea:
Use JPEG- and PNG-specific parsing if possible and brute-force search
only for others.

- Adopted go-exif/v2.

Fixes #172


dsoprea:
Not sure whether you want this on develop or master

dsoprea:
Comments on the acceptance criteria:

**"Exif extraction runs faster, ideally demonstrated by a Go test / benchmark"**

It works faster by nature of the design. Instead of an unconditional, byte-by-byte search, we now do a JPEG parse if a "jpg" or "jpeg" file-extension, a PNG parse if a "png" file-extension, and fallback to a brute-force search for any other extension. So, it'll be just as fast if every image that Photoprism sees is not a JPEG and not a PNG. Otherwise, it'll definitely, always be faster. QED

**"Works with all file formats we support at the moment (or more)"**

We optimized what was already there and provided a structure for adding more extension cases in the future. I think the scope of the issue should stop there. I recommend that new issues are created for file-formats are you want to immediately add handling for, now, and then we can add new issues in the future as newer formats see more priority.

**"All currently supported Exif fields are supported (or more)"**

I recommended that this be considered out-of-scope for this ticket. Enrichen the properties that we pull from EXIF is a different problem than image-format parsing optimization.

**"Documentation / wiki updated (as needed)"**

_What needs to be done here?_

lastzero:
Cool, thank you very much! Usually develop because we automatically build our demo and other docker images from master, so any bugs will hit users instantly and the latest code is in develop. Can you send this PR again there?

lastzero:
Regarding docs, we have this page: https://github.com/photoprism/photoprism/wiki/EXIF

You could add the comments here in this PR as documentation.

dsoprea:
@lastzero 

Done. See https://github.com/photoprism/photoprism/pull/192 .

PR #192, opened 2020-01-14
Optimize EXIF integration

### LINKED ISSUE #172: Improve Exif data extraction from images

dsoprea:
Use JPEG- and PNG-specific parsing if possible and brute-force search
only for others.

- Adopted go-exif/v2.

Fixes #172


dsoprea:
See comments in https://github.com/photoprism/photoprism/pull/191 .

lastzero:
I'll wait until CI is green and then merge this :)

PR #198, opened 2020-01-14
add dutch translation

mipmip:
proost


lastzero:
Cool, thank you! UI is still under development, so we'll need updates from time to time ;)

mipmip:
No problem. You can ping me when there's new stuff to translate to Dutch. I very much like you initiative.

Issue #187, opened 2020-01-11
[CLI]: Provide the option to skip unchanged files


linosgian:
As shown below in the API's `StartIndex` route, an option can be passed to skip indexing photos that haven't changed and thus reduce indexing time. 
https://github.com/photoprism/photoprism/blob/ba9b3cad72e402f51ee88e77b3c8a5705e117b75/internal/api/index.go#L89-L93

In the CLI's `index` command there seems to be no such option:
https://github.com/photoprism/photoprism/blob/ba9b3cad72e402f51ee88e77b3c8a5705e117b75/internal/commands/index.go#L48-L49 

I am syncing photos from multiple devices to a single directory, so I'd like to be able to create a cronjob to periodically index photos, but currently it takes too many resources to do it hourly. If I was able to skip unchanged files, it should be feasible.

lastzero:
You are absolutely right... should be done.

marcoamorales:
I'd like to help with this.. fairly new to the project though @lastzero if you could provide any pointers to find which files have been indexed already, I would really appreciate that :)

lastzero:
@marcoamorales You have to add command flags to https://github.com/photoprism/photoprism/blob/develop/internal/commands/index.go

See https://github.com/photoprism/photoprism/blob/develop/internal/commands/thumbnails.go for how to do that ("force" flag).

Main task is to think about what flags to add and how to name them. Probably "all" (short "a") makes sense. Code would look like:

```
	Flags: []cli.Flag{
		cli.BoolFlag{
			Name:  "all, a",
			Usage: "Re-index existing files",
		},
	},
```

lastzero:
Make sure to run tests before pushing / committing code: `make test-go`

Should be described in our Developer Guide (see Wiki). If anything is unclear, let us know!

lastzero:
Used the opportunity to improve our command help... will merge to master now.

If this solves your issue, you can close the ticket :)

lastzero:
Build is running: https://travis-ci.org/photoprism/photoprism/builds/640370213

I'll also start a build for arm64 (using our own hardware)

linosgian:
I've tested this one, it works perfectly! :)

PR #211, opened 2020-01-21
[CLI]: Add --skip-unchanged flag for index command #187

linosgian:
I took the liberty of taking a first try at this one as I need to create a cron out of it for my gallery.
I've not added a test for this yet, but I can look into verifying which mediafiles were skipped during indexing in `Index.Start`.

### Description
In order to provide the same flexibility as the API, expose the
`SkipUnchanged` flag contained in `IndexOptions`. Although the indexing
process is changing rapidly, and so re-indexing all photos might the
right way to go for now. This flag will be very useful, once indexing is
stable.

For more info refer to https://github.com/photoprism/photoprism/issues/187

### DIFF
--- a/internal/commands/index.go
+++ b/internal/commands/index.go
@@ -11,11 +11,19 @@ import (
 	"github.com/urfave/cli"
 )
 
-// Re-indexes all photos in originals directory (photo library)
+// indexes all photos in originals directory (photo library)
 var IndexCommand = cli.Command{
 	Name:   "index",
-	Usage:  "Re-indexes all originals",
+	Usage:  "indexes all originals",
+	Flags:  indexFlags,
 	Action: indexAction,
+}
+
+var indexFlags = []cli.Flag{
+	cli.BoolFlag{
+		Name:  "all, a",
+		Usage: "re-index all originals, including unchanged ones",
+	},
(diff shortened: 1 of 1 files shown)


lastzero:
Wondering if it wouldn't make sense to have this as default? That's why I suggested --all as flag and not --skip-unchanged

linosgian:
Hey,

As I said in the description above, I think that since the indexing process is not stable currently, it would make sense to automatically reindex everything by default. I'd be fine with either solution.

I'll do that shortly

lastzero:
@linosgian It is pretty stable, we didn't change much recently. Re-indexing happens automatically if you delete the database after schema changes. My assumption is that most users won't expect that index means "reindex everything" by default.

linosgian:
Alright, I've force pushed the `--all` flag instead. Let me know if there's anything more I should do.
I've signed the CLA already, but github doesn't seem to like the email I've signed off with.

lastzero:
Thank you! Celebrating my birthday now, will merge this asap. Maybe you need to change your local git config and set the email you use on GitHub?

linosgian:
Fixed. No worries, happy birthday! :)

PR #218, opened 2020-01-23
Backend: Add API endpoint for photo count per month

thielepaul:
This new endpoint allows the mobile app to have a draggable scrollbar without loading the metadata for all photos.

### DIFF
--- a/internal/api/photo_count_per_month.go
+++ b/internal/api/photo_count_per_month.go
@@ -10,9 +10,9 @@ import (
 	"github.com/gin-gonic/gin"
 )
 
-// GET /api/v1/photoCountPerMonth
+// GET /api/v1/moments/time
 func GetPhotoCountPerMonth(router *gin.RouterGroup, conf *config.Config) {
-	router.GET("/photoCountPerMonth", func(c *gin.Context) {
+	router.GET("/moments/time", func(c *gin.Context) {
 		if Unauthorized(c, conf) {
 			c.AbortWithStatusJSON(http.StatusUnauthorized, ErrUnauthorized)
 			return
@@ -22,7 +22,7 @@ func GetPhotoCountPerMonth(router *gin.RouterGroup, conf *config.Config) {
 
 		result, err := q.GetPhotoCountPerMonth()
 		if err != nil {
-			c.AbortWithStatusJSON(400, gin.H{"error": txt.UcFirst(err.Error())})
+			c.AbortWithStatusJSON(http.StatusInternalServerError, gin.H{"error": txt.UcFirst(err.Error())})
 			return
 		}
 
--- a/internal/api/photo_count_per_month.go
+++ b/internal/api/photo_count_per_month.go
@@ -1,0 +1,31 @@
+package api
+
+import (
+	"net/http"
+
+	"github.com/photoprism/photoprism/internal/config"
+	"github.com/photoprism/photoprism/internal/query"
+	"github.com/photoprism/photoprism/pkg/txt"
+
+	"github.com/gin-gonic/gin"
+)
+
+// GET /api/v1/photoCountPerMonth
+func GetPhotoCountPerMonth(router *gin.RouterGroup, conf *config.Config) {
+	router.GET("/photoCountPerMonth", func(c *gin.Context) {
+		if Unauthorized(c, conf) {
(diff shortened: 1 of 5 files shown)


thielepaul:
Bildschirmfoto_2020-01-23_19-11-47

lastzero:
I can merge this, but note that we'll probably change the URL in something like "/moments/time" as we also need it for the Web UI... additionally grouped by country, maybe like "/moments/country". Not sure where I have the query, but I've prepared it recently.

internal/api/photo_count_per_month.go
```diff
@@ -0,0 +1,31 @@
...
+		if Unauthorized(c, conf) {
+			c.AbortWithStatusJSON(http.StatusUnauthorized, ErrUnauthorized)
+			return
+		}
+
+		q := query.New(conf.OriginalsPath(), conf.Db())
+
+		result, err := q.GetPhotoCountPerMonth()
+		if err != nil {
+			c.AbortWithStatusJSON(400, gin.H{"error": txt.UcFirst(err.Error())})
```
lastzero: Use the http.* constant instead of an integer

internal/api/photo_count_per_month.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/photoprism/photoprism/internal/config"
+	"github.com/photoprism/photoprism/internal/query"
+	"github.com/photoprism/photoprism/pkg/txt"
+
+	"github.com/gin-gonic/gin"
+)
+
+// GET /api/v1/photoCountPerMonth
+func GetPhotoCountPerMonth(router *gin.RouterGroup, conf *config.Config) {
+	router.GET("/photoCountPerMonth", func(c *gin.Context) {
```
lastzero: Use "/moments/time" as URL

internal/query/photo_count_per_month.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+// GetPhotoCountPerMonth couts photos per month and year
+func (s *Repo) GetPhotoCountPerMonth() (results []PhotoCountResult, err error) {
+	q := s.db.NewScope(nil).DB()
+
+	q = q.Table("photos").
+		Select("photos.photo_year, photos.photo_month, COUNT(*) AS count").
+		Group("photos.photo_year, photos.photo_month").
+		Order("photos.photo_year DESC, photos.photo_month DESC")
```
lastzero: Looks good, but you might want to ignore deleted / archived photos for this?

internal/api/photo_count_per_month.go
thielepaul: good catch, would `http.StatusInternalServerError` be a good choice?

lastzero: Maybe yes, since this should never happen? Otherwise StatusBadRequest would be OK.

internal/api/photo_count_per_month.go
thielepaul: :heavy_check_mark:

internal/api/photo_count_per_month.go
thielepaul: :heavy_check_mark:

internal/query/photo_count_per_month.go
thielepaul: :heavy_check_mark:

internal/api/photo_count_per_month.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/photoprism/photoprism/internal/config"
+	"github.com/photoprism/photoprism/internal/query"
+	"github.com/photoprism/photoprism/pkg/txt"
+
+	"github.com/gin-gonic/gin"
+)
+
+// GET /api/v1/moments/time
+func GetPhotoCountPerMonth(router *gin.RouterGroup, conf *config.Config) {
```
lastzero: Can you also rename this to `GetMomentsTime()`? Looks good otherwise, probably need to add fixtures to properly test this with unit tests... we can do this later, so that you can focus on developing our mobile app :)

thielepaul: sound good :+1:

lastzero:
Cool, will be merged as soon as Travis CI is green 🥝

lastzero:
Also added you and @spaibs as team members to our org as you contribute a lot and maintain the mobile app.

Issue #221, opened 2020-01-24
Auth: Support encrypted passwords in addition to plain text 


lastzero:
**As an admin, I don't want to use a plain text password in my configuration so that my password can not leak.**

Should be easy to detect if the password is encrypted since Unix passwords typically start with a specific sequence like

`$1$EqkVUoQ2$4VLpJuZ.Q2wm6TAiyYt75.`

Acceptance Criteria:
- [ ] Setting a site password via PHOTOPRISM_ADMIN_PASSWORD also works with (standard unix) encrypted passwords and not just plain text
- [ ] Please add a unit test for this!

lastzero:
Example: https://medium.com/@jcox250/password-hash-salt-using-golang-b041dc94cb72

guysoft:
BTW, current workaround I use is https://github.com/jwilder/nginx-proxy with basic auth.

lastzero:
@guysoft Sure, but we want to get rid of this to simplify installation... should not require many lines of code :)

lastzero:
Should work, maybe somebody can test this with our latest Docker image?

guysoft:
@lastzero I have it with auto-updating (watchtower). Can you explain exactly what you need testing?

dennorske:
I am unable to get standard unix encryption working.
I encrypted it using: `makepasswd --crypt-md5 --clearfrom file`

On latest docker image, built 32 minutes ago from the docker repo.. does that include this change?

lastzero:
From what I see, only bcrypt is supported currently (should start with `$2`). Any specific reason you're using md5?

lastzero:
We're merging and building several times a day, so it should never take more than one or two days for you to get the latest code (unless everyone is on vacation).

dennorske:
My mistake, I don't know why i thought it was MD5. Could be I identified the example hash in this issue to be that. I now saw that the PR #231 has support for bcrypt.

I tested and it worked :+1:

lastzero:
Thank you very much!

FoxxMD:
To clarify: #231 is only for encrypting `admin-password` correct? This does not affect `sql-password` or `database-dsn`?

lastzero:
We can not hash the db password as the app needs to be able to access the database.

FoxxMD:
Thanks for the clarification 👍

PR #231, opened 2020-01-27
support encrypted password

agolebiowska:
This PR should be an answer to issue [#221](https://github.com/photoprism/photoprism/issues/221)

It only handles bcrypt. Maybe rest of the unix hash functions are needed too? http://man7.org/linux/man-pages/man3/crypt.3.html

If `utils.go` is wrong place to put this code or it should be implemented differently please let me know!


lastzero:
Thank you, I'll merge this! ✨

Might rename utils.go to auth.go, trying to avoid generic names

PR #238, opened 2020-02-02
Make aware that bcrypt is supported in dockerfile

dennorske:
Just quickly adding that bcrypt is supported in the dockerfile. I did the edit in the dockerfile linked over at  [docker hub](https://hub.docker.com/r/photoprism/photoprism) Anywhere else this could be stated?


lastzero:
Thanks! You have to escape $ in docker-compose.yml? Never needed it...

dennorske:
Yep escaping is necessary else it inteprets it as variables. https://docs.docker.com/compose/compose-file/#variable-substitution

PR #240, opened 2020-02-04
Do not print starting web server before starting web server

abrander:
Things can go wrong and abort in `server.Start()` before the webserver is
started. This commit will move the log message to where the webserver is
started to avoid confusing a user debugging a problem.

This was motivated by me trying to start the photoprism server and not realizing that the unrelated error `html/template: pattern matches no files:` was recovered by server.Start() *before* it got to `ListenAndServe()` thereby leading me to believe that a web server would indeed be listening when it was not.

This PR simply moves the `log.Infof()` to a point when we're sure we're actually trying to start the webserver.

### DIFF
--- a/internal/commands/start.go
+++ b/internal/commands/start.go
@@ -106,8 +106,6 @@ func startAction(ctx *cli.Context) error {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


lastzero:
Thanks, I'll merge this tomorrow. Note that we might compile assets in so that you only need the binary

PR #256, opened 2020-02-20
Serialize websocket writes with a mutex

### LINKED ISSUE #241: concurrent write to websocket connection
In some cases the user can trigger a `panic()` from `github.com/gorilla/websocket`.

I believe the culprit is `wsReader()` and `wsWriter()` which will execute is separate go routines and both write to the same underlying `websocket.Conn`.

The Gorilla documentation touches concurrency here:
https://godoc.org/github.com/gorilla/websocket#hdr-Concurrency

ifedorenko:
Fixes #241

Signed-off-by: Igor Fedorenko <igor@ifedorenko.com>

### DIFF
--- a/internal/api/websocket.go
+++ b/internal/api/websocket.go
@@ -33,7 +33,7 @@ var wsAuth = struct {
 	mutex         sync.RWMutex
 }{authenticated: make(map[string]bool)}
 
-func wsReader(ws *websocket.Conn, connId string, conf *config.Config) {
+func wsReader(ws *websocket.Conn, writeMutex *sync.Mutex, connId string, conf *config.Config) {
 	defer ws.Close()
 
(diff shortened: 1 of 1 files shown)


internal/api/websocket.go
```diff
@@ -131,6 +136,8 @@ func Websocket(router *gin.RouterGroup, conf *config.Config) {
 			return
 		}
 
+		var writeMutes sync.Mutex
```
lastzero: Should be writeMutex, not writeMutes :)

lastzero:
Thank you very much! Found a potential typo. If this is fixed, I'll merge it :)

ifedorenko:
Fixed. Thank you for the feedback.

Issue #113, opened 2019-05-30
Improve comments based on best practices from Effective Go


lastzero:
**As a developer, I want well written comments so that I can work more effectively.**

Somebody started rewriting our docs / comments but never signed the CLA, so we had to close the pull request #100. This should be an easy task, also for non-developers / juniors.

See https://golang.org/doc/effective_go.html#commentary for how to write good comments. Fixing comments has a fairly low priority for the core team right now because of time constraints, so any help is most welcome!

Bad: `// Returns true if file exists`
Good: `// Exists returns true if file exists`

Charliekenney23:
I'll take this

Charliekenney23:
@lastzero What is your preference on the http handler doc comments. i.e.
```go
// GET /api/v1/photos
//
// Query:
//   q:         string Query string
//   tags:      string Tags
//   cat:       string Category
//   country:   string Country code
//   camera:    int    Camera ID
//   order:     string Sort order
//   count:     int    Max result count (required)
//   offset:    int    Result offset
//   before:    date   Find photos taken before (format: "2006-01-02")
//   after:     date   Find photos taken after (format: "2006-01-02")
//   favorites: bool   Find favorites only
func GetPhotos(router *gin.RouterGroup, conf *config.Config) {
```

Should the top line instead read:`// GetPhotos handles GET /api/v1/photos`?

lastzero:
@Charliekenney23 Yes, why not... maybe you can figure out if there are any (other) best practices regarding the documentation of gin / http handlers / routes? This was a quick shot to at least offer some basic information to frontend developers.

lastzero:
It seems like there is a go version of Swagger that works with simple inline route docs:

```
// ServeAPI serves the API for this record store
func ServeAPI(host, basePath string, schemes []string) error {

    // swagger:route GET /pets pets users listPets
    //
    // Lists pets filtered by some parameters.
    // ...
```

See https://goswagger.io/use/spec/route.html 

So in theory you can change the comment and additionally use `// swagger:route` for the method and route - obviously you should also test and document if this works. If you don't have time for this, we can create a new ticket for somebody else.

Charliekenney23:
I can do it, but in a separate PR.

lastzero:
Comments are still far from perfect, so everyone is welcome to send pull requests (ideally small ones so that it doesn't take very long to do and review).

FrouxBY:
[...]

PR #257, opened 2020-02-20
Some more comment improvements

FrouxBY:
Continuing to improve the quality of code comments.  #113


lastzero:
Thank you! Please make sure to use "returns" instead of "return" (or "extends" etc), with an "s" end the end of the verb, when describing what a function does :)

FrouxBY:
You are right, I definitely need to revise my English grammar (this "s" have always been my nightmare).

PR #279, opened 2020-04-06
Skip hidden directories on indexing

### LINKED ISSUE #275: Skip indexing of hidden directories
Currently [indexing skips hidden files](https://github.com/photoprism/photoprism/blob/develop/internal/photoprism/index.go#L104) I'd like to extend this to also skip over hidden directories entirely. My exact usecase is that I'm syncing my phone camera directory via syncthing and have a directory called `.thumbnails` I want to skip over indexing. Would you be comfortable with this being the default or would it need to be hidden behind a flag? I'm happy to contribute this change.

rockwotj:
Fixes #275

### DIFF
--- a/internal/photoprism/index.go
+++ b/internal/photoprism/index.go
@@ -101,7 +101,13 @@ func (ind *Index) Start(options IndexOptions) map[string]bool {
 			return nil
 		}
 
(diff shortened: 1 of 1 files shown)


lastzero:
Thanks! Didn't notice this is necessary to skip a directory...

PR #354, opened 2020-06-09
Update to hide private folder in public mode.

hmalhi:
This way you can run docker process in both private and public and not share private images in public site.

### DIFF
--- a/frontend/src/component/p-navigation.vue
+++ b/frontend/src/component/p-navigation.vue
@@ -104,7 +104,7 @@
                     </v-list-tile-content>
                 </v-list-tile>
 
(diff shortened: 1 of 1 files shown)


lastzero:
Note that you can still use the API to find private photos, even if you remove the link from the navigation. We don't want to suggest strict protection in public mode (which was originally meant for development, testing and our demo) when there is none. Please don't use pubic mode for public galleries when you have strictly private content you need to hide. Private photos can be securely hidden to external users / guests as soon as we have link sharing implemented.

lastzero:
As you might have noticed, we've implemented link sharing. This provides an easy way to share albums with others (but without separate labels and maps view for now). In addition, we plan to protect private photos with a code so that not everyone can open them. But that's for later, after our initial release.

PR #381, opened 2020-07-06
Added animation for slide show.

rakesh-saride:
I added CSS animation to the slideshow. 
- `slide_show` class gets added when the slideshow starts and will be removed when the slideshow is paused.


lastzero:
Cool, thank you very much! Any browsers / devices this might cause issues with? What happens on touch screens if the user swipes during a slide show? I'll merge this now to test it in develop.

rakesh-saride:
I did not find any issues with the CSS transitions with touch operations on mobile.

PR #387, opened 2020-07-07
Parallelize Darktable

bobobo1618:
By adding the `--apply-custom-presets false` option to the `darktable-cli` command line, Darktable [doesn't read from `data.db`](https://github.com/darktable-org/darktable/blob/release-3.1.0/src/cli/main.c#L248) and so doesn't lock `data.db` and `darktable-cli` can be run in parallel.

It might be a good idea to make this optional but I thought this was a start.

### DIFF
--- a/internal/photoprism/convert.go
+++ b/internal/photoprism/convert.go
@@ -112,12 +112,12 @@ func (c *Convert) ConvertCommand(mf *MediaFile, jpegName string, xmpName string)
 			result = exec.Command(c.conf.SipsBin(), "-s", "format", "jpeg", "--out", jpegName, mf.FileName())
 		} else if c.conf.DarktableBin() != "" {
 			// Only one instance of darktable-cli allowed due to locking
-			useMutex = true
+			//useMutex = true
 
 			if xmpName != "" {
-				result = exec.Command(c.conf.DarktableBin(), mf.FileName(), xmpName, jpegName)
+				result = exec.Command(c.conf.DarktableBin(), "--apply-custom-presets", "false", mf.FileName(), xmpName, jpegName)
 			} else {
-				result = exec.Command(c.conf.DarktableBin(), mf.FileName(), jpegName)
+				result = exec.Command(c.conf.DarktableBin(), "--apply-custom-presets", "false", mf.FileName(), jpegName)
 			}
 		} else {
 			return nil, useMutex, fmt.Errorf("convert: no raw to jpeg converter installed (%s)", mf.Base(c.conf.Settings().Index.Group))


bobobo1618:
I've signed the CLA but I don't use a legitimate email address for my commits as it seems to be a surefire way to attract spam (like recruiters).

lastzero:
Can you send a copy with a short written agreement from your real address to hello@photoprism.org? Or print it, sign it and send the scan as PDF. It's just to avoid legal risks so that nobody sues us unexpectedly.

flokli:
https://github.com/darktable-org/darktable/issues/2890 might be related - but in any case, we should open a bug report with this issue, and try to get this fixed properly.

bobobo1618:
@flokli I think everything is WAI from the DT perspective. I think maybe they just need some documentation explaining that this flag is needed in order to run in parallel?

flokli:
Is there a specific bug about this? I mean, `darktable-cli` is explicitly for batch operations, I'd be very surprised if "can't run multiple conversion jobs in parallel" is a WAI.

bobobo1618:
Yes, it's been brought up in the past and the answer was that Darktable doesn't support multiple instances: https://github.com/darktable-org/darktable/issues/4081#issuecomment-573193622

It makes sense. They can either:

- Default to not loading the preset DB and end up outputting garbage to users who use presets (likely quite numerous) but supporting users who want to run many instances (likely not numerous)
- Default to loading the preset DB and give the expected output but cause issues when running in parallel. 

So from their side the behaviour is entirely WAI. But as this PR indicates, there is a flag that switches to the behaviour we might want if we're running in parallel, so perhaps that flag should be documented as having this effect.

flokli:
I don't know enough about darktable presets to judge whether they influence how images will be converted - if it matters, it would probably need to be a flag in photoprism, so users can choose what they prefer. At least the docker-based installations don't really share the darktable preset DB with the users' one anyways, right?

Regarding "WAI": Shouldn't reading custom presets only be a problem as soon as something tries to open the database read-write?

bobobo1618:
> I don't know enough about darktable presets to judge whether they influence how images will be converted

If the user employs a custom preset, they can change pretty much anything about how the image should be converted. If you ignore the preset with this flag, the output will be completely different to what the user would have seen and intended in Darktable.

> if it matters, it would probably need to be a flag in photoprism, so users can choose what they prefer. At least the docker-based installations don't really share the darktable preset DB with the users' one anyways, right?

The latter sentence, for me, makes the former a bit of a moot point. Currently, PhotoPrism doesn't have a standard/documented way for a user to allow `darktable-cli` access to their main `data.db`, so it seems a bit pointless to me at this point. That's something that should be added in a feature request later IMO.

> Regarding "WAI": Shouldn't reading custom presets only be a problem as soon as something tries to open the database read-write?

SQLite3 is cool with it but the DT developers deliberately added their own, separate DB lock to ensure that only one instance of DT can _access_ the DB at a time. DT is primarily a piece of GUI software and if you had multiple instances of it running simultaneously, it would lead to weird, inconsistent behaviour. I believe this is why they made this choice.

In theory they could add some exception for `darktable-cli` but since `darktable-cli` is just a CLI interface to the same backend the GUI uses, I think they're worried about causing the same problems you'd get with the GUI.

lastzero:
Implemented this along the way today, when working on albums. Was in a hurry, so no extra commit. Maybe you have a better idea for the config flag name.

PR #388, opened 2020-07-07
Add size cap parameter to Darktable to avoid huge JPEGs

bobobo1618:
By default, PhotoPrism leaves the `height` and `width` parameters of `darktable-cli` untouched. When the RAW images are quite large, the JPEGs derived from them can be too (when I gave PhotoPrism my RAWs it ended up writing ~260GB of JPEG previews).

This pull request adds a user-configurable cap that will be propagated through to `darktable-cli`, that can limit the maximum size of the image.

A nice side-effect is that this should also make `darktable-cli` slightly faster, if only because the JPEG encoder has to write a smaller image.

### DIFF
--- a/internal/commands/config.go
+++ b/internal/commands/config.go
@@ -106,6 +106,7 @@ func configAction(ctx *cli.Context) error {
 	fmt.Printf("%-25s %d\n", "thumb-limit", conf.ThumbLimit())
 	fmt.Printf("%-25s %s\n", "thumb-path", conf.ThumbPath())
 	fmt.Printf("%-25s %d\n", "jpeg-quality", conf.JpegQuality())
+	fmt.Printf("%-25s %d\n", "darktable-max-size", conf.DarktableMaxSize())
 
 	return nil
 }
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -254,3 +254,8 @@ func (c *Config) OriginalsLimit() int64 {
 	// Megabyte.
 	return c.params.OriginalsLimit * 1024 * 1024
 }
+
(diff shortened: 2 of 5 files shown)


bobobo1618:
An alternative to this and my other PR might be to just add an arbitrary `--darktable-cli-flags` option that allows the user to change the command freely.

I figure the rigidity should make it easier to use.

lastzero:
We might as well use the thumb max size as limit. If you happen to need a bigger size for a specific image, you can still create it manually without wasting gigabytes of storage. Thoughts?

bobobo1618:
Hmm, for me I think that isn't optimal but I might not understand everything that's going on well enough. My thinking is:

- I don't want to generate all the sizes of thumbnail because they consume _a lot_ of space (more than the Darktable previews), so I set the thumb max size quite low.
- I'm happy to generate any size of thumbnail on-demand because I have a beefy CPU that can handle it.
- I want the preview to accommodate all of the thumbnail sizes I'm likely to generate on-demand, as well as provide a reasonable-quality standin for the RAW file when I view the image full-size in the browser.

If we were to use the thumbnail max max size to cap the preview, that essentially brings the cap over to the on-demand thumbnails as well, which I'd prefer to avoid.

lastzero:
Added a size limit for darktable defaulting to 3840 pixels (4K Ultra HD)... and then noticed 3840 is not as popular as I thought. Apple (now) favors Retina 4K (4096×2304).

Ended up removing `fit_3840` in favor of `fit_4096` (Retina 4K, DCI 4K). Also added `fit_7680` for those who need it (Retina 6K, 8K Ultra HD). Now performing tests. When this is deployed, I'd appreciate some extra testing.

PR #389, opened 2020-07-07
Add basic french translation

FrouxBY:
This adds a basic French translation for most terms currently available,
More accurate and elegant translations will come with more usage of the software (e.g better wording, shorten wording for better fit the UI, ...)


lastzero:
Thank you very much! Note that we also have translations for our backend, but they don't use gettext (yet). See i18n package.

PR #429, opened 2020-07-29
add Chinese translation.

xiaket:
Please let me know if there're any issues. Thanks!


xiaket:
My apologies to traditional Chinese users, at the moment Photoprism has a policy to not to have dialects, so please don't feel offended by this PR. And I would venture to suggest Photoprism to bend that rule a little bit since traditional Chinese and simplified Chinese are different in many aspects.

lastzero:
Thank you so much! Note that we're on vacation, so adding dialect support is not possible right now. We'll revisit this later.

lastzero:
Merged to master: Fresh Docker builds available for testing 🇨🇳

xiaket:
Thanks mate! I'll have to wait cause I'm running photoprism on a raspberry Pi 4 as a family album service. Any chance I can contribute so we can build the arm64 image automatically when we merge to master?

lastzero:
Currently building it on an actual Pi 4. Taking care of this when migrating away from Teavis CI.

PR #433, opened 2020-07-30
Add (new) spanish translation

hialvaro:
Hola! Here's the new complete spanish translation. I think it's all right, if there's something missing I can fix it!


lastzero:
Thank you very much 🙏

hialvaro:
Spanish translation is not showing up when selecting "spanish" on the languages dropdown... :thinking: Is it a bug or there's something wrong with the translation?

lastzero:
That's because you've been using `es_ES` instead of just `es` in the .po file. Do you want to send a quick for or shall I change it later?

lastzero:
Fixed it :) Note that there is no Spanish translation for the backend yet!

PR #436, opened 2020-08-01
Disable proxy_buffering and increase client_max_body_size in nginx card

### LINKED ISSUE #427: Set client_max_body_size in nginx config
The example for the nginx websocket configuration should also include the setting for the max body size, otherwise the upload of larger pictures will fail.

`client_max_body_size 300M;` for example. Or the upload needs to be split up in chunks.

RAYs3T:
* This disables `proxy_buffering` (see: https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffering) meaning the content is served as soon as nginx itself gets it (might also speedup things). This is especially good for larger files (like pictures)
* and sets the `client_max_body_size` (see: https://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size) to 500M (even that might not be enough for uploading large videos, but the default is way to low to upload pictures via the UI  when using nginx.

closes #427

### DIFF
--- a/frontend/src/pages/help/websockets.vue
+++ b/frontend/src/pages/help/websockets.vue
@@ -47,13 +47,15 @@ ProxyRequests off
 http {
   server {
     server_mame example.com
-
+    client_max_body_size 500M;
+    
     location / {
       proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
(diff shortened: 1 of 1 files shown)


lastzero:
Thank you very much 💐

PR #451, opened 2020-08-08
wip: support building without tensorflow

tbhi:
Started trying to make photoprism build without tensorflow.  Open to comments/suggestions.


dfhns967:
i can help with making a python version soo it will be faster i guess and more light weight

lastzero:
We are closing this PR because we don't have time to complete it and there is no demand from our users. Thank you for your contribution!

Issue #437, opened 2020-08-01
Zip: Live Photo downloads should include all related media files


Chadea:
Title.

When downloading live photos (mine have jpg, heic, and the mov) it would be good if all 3 files were zipped for download. Currently it appears to just download the singular jpg/still image associated with that photo.

calucido:
Also, Firefox sometimes has issues asking permission to download multiple files at the same time.
If it's ok with the devs, I'd like to take this on as a first issue.

lastzero:
Sure, pull requests welcome :) still on vacation until next week

calucido:
@Chadea how are you uploading the live photos? Whenever I transfer them off my phone (to Photoprism or to my Mac), I find that they're converted to JPGs.

Chadea:
I’ve been copying them to the import folder over SMB using an app (PhotoSync) on my phone, I had to add the yaml and json sidecar option to my photoprism instance. That allows the small .movs to be linked. As detailed when I asked  https://github.com/photoprism/photoprism/issues/422

Chadea:
I should add, when I sync I have turned on jpg so I have a .jpg .heic .mov and photoprism now links them.

miguelarios:
I created this issue #934 that relates to this issue because I almost think it be nicer to let the user decide which version to pick from.

lastzero:
Would this be resolved when you change the settings as described at https://docs.photoprism.app/getting-started/config-files/settings/#file-downloads?

melyux:
I tried with those setting changes but no luck. Only getting a .heif when downloading Live Photos

> Download:
  Name: file
  Disabled: false
  Originals: true
  MediaRaw: true                                                      
  MediaSidecar: false

graciousgrey:
@melyux What files do you see on the files tab of your Live Photo? Are the related JPG and MOV files located in Originals or in Sidecar?

melyux:
The files tab shows:

- .heic file in Originals
- .mov file in Originals
- .heic.jpg file in Sidecar

graciousgrey:
What version are you running? I just tested the same scenario on my own setup:

Disabled: false
Originals: true
MediaRaw: true
MediaSidecar: false

Results in downloading the .heic and .mov file as those are the originals.

Disabled: false
Originals: false
MediaRaw: true
MediaSidecar: false

Results in downloading the .heic and .mov and .jpg file. 

[...]

Issue #449, opened 2020-08-07
Download: Select Files and Formats and configure defaults


RAYs3T:
## Download of multiple files
So right now whenever you download a single picture, it will download all the files assosiated with this specific one.
This leads to the behavior of downloading a raw+jpeg when you click on the download button.
In firefox for example this gives you two pop-ups to save the file.

In the most situations you only need one type of image. You either want the raw to edit the picture or just the version displayed for whatever.

## How about a dropdown?
What about adding a little drop down menu at the side of the button, where you can select what exact file you wish to download?
![image](https://user-images.githubusercontent.com/16596807/89622365-3b01fe80-d893-11ea-82e3-fe4da2504c4c.png)

## Make the default file type to download an option (later)
This also has great potential to be an option where the user could set its prefered download file type.
Tho I would NOT include this into this issue / PR. That might be something for later.

@lastzero If you're ok with that I would do the changes and throw it in a PR?

lastzero:
Sorry for the wait! You are right, need to improve this. Selecting a smaller (resampled) size should be possible as well. 

Adding a download dialog so that users may select specific file type(s) and size(s) seems necessary. However, this might complicate PhotoPrism for less experienced users as they'll feel unsure what to select.

MrRaph:
It's the only feature missing in PhotoPrism that lead me to find another software... :(

graciousgrey:
Good to know that this feature is important for some users. In the future we plan to have something like a public roadmap/feature voting, so that we gain insights which features are most wanted :)

shoeper:
> Sorry for the wait! You are right, need to improve this. Selecting a smaller (resampled) size should be possible as well.
> 
> Adding a download dialog so that users may select specific file type(s) and size(s) seems necessary. However, this might complicate PhotoPrism for less experienced users as they'll feel unsure what to select.

Maybe the proposed dropdown could contain very few carefully selected options (depending on object type and available formats) and for more there is an `advanced` option.

Another option could be to allow adding additional options in the user or system settings.

Too many options would also slow usage down for advanced users and many average users are even overwhelmed by any options. Probably it would be even better to have one default when clicking download without any further options. E.g. jpeg for images, gif (or mp4?) for moving pictures, mp4 for live pictures, mp4 for videos. Just click download and photoprism provides a good option.

lastzero:
In a first step, downloads have been refactored:
[...]

PR #452, opened 2020-08-08
Zip live photos before downloading

calucido:
Implements #437 for downloading photos individually, from the clipboard, and from shared albums.

### DIFF
--- a/frontend/src/component/photo/clipboard.vue
+++ b/frontend/src/component/photo/clipboard.vue
@@ -214,7 +214,6 @@
               for (let i = 0; i < this.selection.length; i++) {
                 new Photo().find(this.selection[i]).then(photo => {
                   for (let j = 0; j < photo.Files.length; j++) {
-                    window.console.log(photo.Files);
                     photoUIDs.push(photo.Files[j].PhotoUID);
                     if ((i === this.selection.length - 1) && (j === photo.Files.length - 1)) {
                       if (photoUIDs.length === 1) {
--- a/frontend/src/component/photo/clipboard.vue
+++ b/frontend/src/component/photo/clipboard.vue
@@ -127,6 +127,7 @@
 <script>
     import Api from "common/api";
     import Notify from "common/notify";
+    import Photo from "model/photo.js";
     import Event from "pubsub-js";
 
     export default {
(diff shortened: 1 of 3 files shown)


calucido:
Is the clipboard selection still supposed to be active after downloading? After a successful download, there is the line `this.expanded = false`, which looks to me like it is supposed to close the clipboard (as far as I can tell, it does nothing). If the clipboard should close, let me know and I'll add it to this PR.

frontend/src/component/photo/clipboard.vue
```diff
@@ -209,15 +210,25 @@
...
-                        this.onDownload(`/api/v1/zip/${r.data.filename}?t=${this.$config.downloadToken()}`);
-                    });
-                }
-
-                this.expanded = false;
+              let photoUIDs = [];
+              for (let i = 0; i < this.selection.length; i++) {
+                new Photo().find(this.selection[i]).then(photo => {
+                  for (let j = 0; j < photo.Files.length; j++) {
+                    window.console.log(photo.Files);
```
olegelantsev, line 217: Nitpick: maybe should be removed?

frontend/src/share/photo/clipboard.vue
```diff
@@ -83,15 +84,25 @@
...
-                        this.onDownload(`/api/v1/zip/${r.data.filename}?t=${this.$config.downloadToken()}`);
-                    });
-                }
-
-                this.expanded = false;
+              let photoUIDs = [];
+              for (let i = 0; i < this.selection.length; i++) {
+                new Photo().find(this.selection[i]).then(photo => {
+                  for (let j = 0; j < photo.Files.length; j++) {
+                    window.console.log(photo.Files);
```
olegelantsev, line 91: remove?

frontend/src/component/photo/clipboard.vue
calucido: yes to the log (that's left over from me debugging), but the for loops are necessary to download all the files; without them, only the main file is downloaded because only the main file is added to the clipboard.

olegelantsev: Indeed, I meant the log line

calucido: Good catch! All fixed.

lastzero:
Going to review this later today or tomorrow. Can you provide a short description what was changed and why? Maybe with a screenshot if there are UI changes. Thanks! :)

calucido:
Previously, when a live photo was downloaded, only a JPEG still image would be sent. This was a problem because live photos are a package of at least a JPEG/HEIC still image and a video file.
The change to `downloadAll()` in `frontend/src/model/photo.js` enables a single photo's files to be zipped when there are multiple files (there is no change in behavior in the case that there is only one file). This only had an effect on downloads initiated from the fullscreen photo viewer.
The identical changes in `download()` in `frontend/src/component/photo/clipboard.vue` and ` frontend/src/share/photo/clipboard.vue` iterate through each photo in the clipboard, adding the UIDs of all files from each photo to the request that is sent to the /zip API endpoint. This was needed because the clipboard only stores the photo ID; when a download is requested by that photo ID, only the primary file (in the case of a live photo, the still image) is sent.
There are no UI changes, but I was wondering if the clipboard was supposed to close after photo downloads.

lastzero:
Thanks! Spent the day getting Canon .cr3 up and running and upgrading all Docker images to Ubuntu 12.04. I'll try to test / merge it tomorrow. Thought about adding a download dialog that lets you select what files and sizes you want, but then didn't want to delay the release. Maybe this is a good compromise for the meantime.

lastzero:
Didn't forget about this :)

Came to the conclusion that the functionality should be implemented on the server side instead of checking for related files in the frontend as it might have incomplete information. What do you think?

In addition, we should refactor the ZIP download API to use a stream like already implemented for album downloads. It's a bit more complicated as a POST request is currently used for downloading multiple files instead of GET for downloading a single album. Instead, our API could store the selection in a cache using a token and return that token for requesting the actual ZIP file via GET.

In addition, a download dialog for selecting file type(s) and size(s) seems required, see #449.

calucido:
That makes sense. I just did it in the frontend because I'm more comfortable with JS than Go.
I saw the PR about streaming album downloads and I ran into the GET vs POST issue.
College is picking up for me so I don't have as much free time as over the summer, but I can get started on this expanded scope soon. I don't know where PhotoPrism stands in relation to a stable release, but if you're hoping it will be soon, maybe the dropdown could be a v1.1 thing. It seems (to me, at least) important but not a deal-breaker to be able to choose which files to download, while it's critical to be able to download more than the main file associated with a photo, even if that means getting extra files for now.

Issue #453, opened 2020-08-09
Show more columns in tile view on large screens


olegelantsev:
Hi,

I noticed that tiles are blurred on my FullHD screen. It seems that fronted scales the tile size depending on the page width and if page wider than ~1500px, tile size becomes bigger than default thumbnail size - 224px.

![photoprism-screenshot](https://user-images.githubusercontent.com/7139119/89729977-cd90d200-da3a-11ea-813e-24289d9ee781.png)

I see that thumbnails are also generated with higher resolution, so front end could use those.

lastzero:
That's not a retina / high dpi screen, right?

As a quick solution, you can use the Cards view which uses 500px tiles instead of 224px. The Mosaic view was built for smaller mobile screens. Making the thumb size dependent on the visible window size might not be ideal given the performance overhead when resizing (you have to reload all thumbs when over 1000 search results might be shown already).

olegelantsev:
Screen is 1920x1080. 

For now I set the default tile size in frontend to 500px in my branch. 

Is it possible to calculate max tile size in Mosaic view? Then switch between tile size wouldn't be needed.

lastzero:
Size is dynamic based on a responsive grid. Changes when you resize the window. We should show more tiles if the size is larger than 224px, not make them larger and/or replace already loaded images.

olegelantsev:
Why is the issue closed? The problem doesn't seem to be solved. I could help to fix it.

lastzero:
You are right, Theresa closed it because it was marked as question. Should actually create a new issue with a specific solution like showing more tiles (or convert this into a task).

Changing the resolution dynamically is not something we want to implement at this point for the reasons mentioned.

lastzero:
FYI: Vuetify has a 12 column grid system. How many columns do you see right now on your screen?

See https://v15.vuetifyjs.com/en/framework/grid

olegelantsev:
I have 6 columns. Thanks for the link.

lastzero:
In that case you probably only have to add the class `xl1` to the tile markup, see `.vue` component in `/frontend/src`

olegelantsev:
I struggle to serve the frontend after it is built. Is it possible to do that without docker? I am new to Vue (I saw vue-cli-service tool, but apparently it doesn't quite work with the frontend setup)

lastzero:
I can do it tomorrow. Have a great evening!

PR #463, opened 2020-08-13
Show 12 columns on desktop screens

olegelantsev:
Fix for issue #453 

Sets 12 columns on large screens (1264px < width < 1904px). Tile sizes on mosaic view don't exceed the default thumbnail size.


lastzero:
Thanks! In fact, you could have renamed `xl1` to `lg1` as keeping both doesn't make a difference...

Issue #424, opened 2020-07-26
Sharing: Add well visible "download as zip" button 


wiwie:
I would like my family to be able to download all photos from a share with them as a zip file.

lastzero:
Select the album in the overview and download using the context menu should work.

wiwie:
Okay. That works. But if someone is sent the share link of an album I think it's not very intuitive that they have to go back to the share/album list and download it from there. I think it's also not easy to discover that one can actually click on albums to get to that list. Would it be possible to make this somehow clearer in the UI? With an icon (e.g. a back arrow) or something similar?

lastzero:
Yes, that's possible. We tested some ideas and got stuck with what we have now. In addition, you can select all using a long or shift click and then download as zip without going back to the overview.

lastzero:
Maybe we can add a download button at the top where it's well visible...

wiwie:
That sounds great.

lastzero:
Added a download button to albums (not visible on very small screens like mobile phones):

<img width="689" alt="Screenshot 2020-08-12 at 05 41 29" src="https://user-images.githubusercontent.com/301686/89973083-c1398e80-dc5f-11ea-9a79-dc5e0ab5a816.png">

wiwie:
When downloading a larger album it takes quite a while from clicking to the download actually initiating. So long that I in fact clicked twice and was already wondering whether it wasn't working at all.

While I don't know what the best approach would be to improve on this, I could imagine that streaming the zip file rather than writing it in full and then sending it off would prevent the long delay. Not sure though whether it's possible to stream a zip file like that. Basically building it incrementally while streaming already.

lastzero:
That's true. For shared SaaS applications, you would typically implement this asynchronously and send an email notification. However, PhotoPrism typically isn't shared with other users and most of the time, it should be fast enough (more like seconds, not minutes or hours) when creating ZIP files.

Maybe we can improve our notifications and/or show a "please wait" message until the download starts...

wiwie:
I was thinking of something similar to this: https://github.com/scosman/zipstreamer

lastzero:
You're right, might be an option for us too. I was intentionally conservative for our initial (MVP) implementation, but in general you could probably stream the output like this:

https://stackoverflow.com/questions/46791169/create-serve-over-http-a-zip-file-without-writing-to-disk

lastzero:
Great example of how a small task ("add a button") becomes a big one 🥳

[...]

PR #492, opened 2020-09-06
Backend: stream album ZIP files

andir:
ZIP files generated for albums will no longer be written to disk prior
serving them to users. The zip file will be written directly into the
HTTP response.

This was done to mitigate the behavior described in https://github.com/photoprism/photoprism/issues/424#issuecomment-673071740 with the approach mentioned in https://github.com/photoprism/photoprism/issues/424#issuecomment-673289432.

### DIFF
--- a/internal/api/album.go
+++ b/internal/api/album.go
@@ -4,8 +4,6 @@ import (
 	"archive/zip"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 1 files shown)


lastzero:
Excellent, thank you! 👍

PR #509, opened 2020-09-23
Implement archive action in photo viewer

Morl99:
So I just gave it a try in order to better understand the codebase (and fight with WSL2, docker and alpine... ).
This implementation still has some todos:

1. The api does not seem to expose if a photo is currently archived or not. So it is not possible to correctly reflect the button. The correct implementation would be to behave as a toggle, just as the 'like' button does. Depending on the state, the action would alternate between 'archive' and 'restore'. Am I missing something, or do we need to add 'archived' to the API?
2. I used the Notify component, but I just can't seem to figure out where this message is displayed. Feedback for the user really is necessary here, otherwise he will be confused if hitting the button worked. 

What else could be missing?

### DIFF
--- a/frontend/src/component/photo/viewer.vue
+++ b/frontend/src/component/photo/viewer.vue
@@ -27,6 +27,11 @@
             <v-icon size="16" color="white">edit</v-icon>
           </button>
 
+          <button class="pswp__button action-archive hidden-shared-only" style="background: none;"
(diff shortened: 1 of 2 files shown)


graciousgrey:
Thank you for your pull request! At the moment we are super busy but we will come back to it as soon as we find time :)

Morl99:
I am closing this pr, as it looks like you are reimplementing the photo viewer with this feature in mind as well.

PR #522, opened 2020-10-01
Fix button class action-toggle-fullscreen

phil-davis:
This looks like a typo that might also be a problem. But I don't see where `action-toogle-fullscreen` or `action-toggle-fullscreen` is actually used by anything - so maybe it makes no difference?


graciousgrey:
Thank you, good catch :) Those classes are mainly for testing, but not all of them are in use.

PR #523, opened 2020-10-01
Fix minor typos

phil-davis:
Nothing exciting here! These would not have affected functionality.


graciousgrey:
Thank you for fixing :)

Issue #490, opened 2020-09-04
Parse JSON lists in metadata like "Subject"


totti4ever:
Hey,

I'm coming closer to a perfect world - which right now only seems reachable with Photoprism! :-)

----

So, I'm working with F-Stop Gallery on Android as it seems to be the only comfy solution to work with tags?! Files on my mobile are synced using Syncthing to my server, where Nextcloud (external SMB) and Photoprism (docker mount).

The tags from F-Stop Gallery are written as Subject into the JPG files and Photoprism shows them:
![image](https://user-images.githubusercontent.com/49901208/92273929-9653fa80-eeec-11ea-9f07-06b698fdd6d1.png)

1. How can I search for those subjects? From another [issue](https://github.com/photoprism/photoprism/issues/419#issuecomment-663968341) I understand that the problem might be the JSON format? Fix would be that you guys extend the indexing process to understand and separate those subjects?

2. I read somewhere that labels are somewhat special and are automatically matched to keywords - if existing keywords exist?! Does this somehow also work with subjects as there are the tags I put onto the pictures on my phone? Or is the only solution to do some magic with exiftool and transfer the subjects into exif-keywords?

3. When an exif-title exists, I see from my experiments that Photoprism uses it - and if it doesn't exist it tries to create one, right? If I edit a title in Photoprism the only solution to also see it on my Phone would be to also sync the sidecar file?

Thanks a lot for this great piece of software - by far the best approach I've seen so far for my personal needs. Will be happily donate some money once I got everything running here (although I know it'd only a minor help - but I'm no coder :-( )

lastzero:
Parsing the JSON should actually not be that difficult :) We're willing to accept pull requests, or I'll do it where I have a little bit of time... still pretty busy with our backend / geodata infrastructure.

We mainly do full-text search plus some more specific fields, see https://docs.photoprism.org/user-guide/organize/search/

Subjects are not indexed yet in a special way, full-text search should work of course. Keywords will be matched, so that you can see files with known keywords in Labels. It doesn't make sense to sync both ways, because then labels and keywords would be the same:

- Labels are general categories, like "cat", "dog", or "boat"
- Keywords can be any words like "kitten", "cat", "cats",... you typically don't want to have individual labels for all of those as they all mean about the same

We use titles if we find them. Even from file names, unless it seems extremely long, broken, numeric etc. Note that there is no title in Exif, just a description field that may be used like a title. Titles exists in Dublin Core (DC) which is part of XMP.

[...]

PR #526, opened 2020-10-02
Added conversion of json array metadata in the Subject field

Morl99:
So this is the first try at #490. I cannot get the test to work, because no matter what I do, the details.Subject field is always an empty string. I might have a misconception about which metadata field gets parsed into Subject. I tried xmp:subject. Or maybe I need to do something else in the test code, to get the xmp data parsed correctly. 

I am happy for any pointers about this code.


Morl99:
Fixed with a different approach

PR #585, opened 2020-11-07
Correction for french correction

jean-louis67:


lastzero:
@graciousgrey Can you check where the merge conflicts are coming from?

graciousgrey:
Thank you for those improvements! 
It seems that this merge conflict appeared because your branch is not up to date with our current develop branch. 
I took over your changes and pushed them to directly to develop. So no need to fix something here.
Regarding your request of having  "Monochromes" and "Privates" in the translation.pot - I put this task on our to do list, but it will take  a while because other things have higher priority right now :)

PR #607, opened 2020-11-14
Fix French translations of "scan"

alexislefebvre:


alexislefebvre:
It has been fixed in #591

PR #631, opened 2020-11-21
Allow convert in readonly mode

### LINKED ISSUE #630: Unlock conversion in read-only mode
Today the [documentation ](https://docs.photoprism.org/developer-guide/library/converting/#read-only-mode)states that:

> Conversion is currently disabled in read-only mode because creating JPEGs from RAWs is a pretty expensive operation and it makes sense to store JPEGs right next to their RAW files so that they can be found and used again in the future.

But if we enable `convert` in the config file, PhotoPrism will generate thumbnails for non JPEG images and videos, and store them in the sidecar directory without any problem.

So this limitation looks outdated.

alexislefebvre:
I haven't tested but it should do the trick.

Fixes #630

### DIFF
--- a/frontend/src/pages/settings/general.vue
+++ b/frontend/src/pages/settings/general.vue
@@ -45,7 +45,7 @@
             <v-flex xs12 sm6 lg3 class="px-2 pb-2 pt-2">
               <v-checkbox
                       @change="onChange"
-                      :disabled="busy || readonly"
+                      :disabled="busy"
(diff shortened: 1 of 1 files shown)


lastzero:
Restarted the build.

lastzero:
Excellent catch, thank you!

alexislefebvre:
I pulled last Docker image and it looks like it worked. 🎉

PR #632, opened 2020-11-21
changed translation of 'in'

jean-louis67:
_in 4,608 ms_ must be translated: _en 4,608 ms_ (and not _dans 4,608 ms_)
I just hope 'in' is not used somewhere else with an other meaning.


lastzero:
Merci beaucoup!

PR #634, opened 2020-11-22
Move advance search arrow next to search bar

eMerzh:
Hey, 
here is a teeny tiny PR, to dip my toes in the water...

I just discovered photoprism, really nice job, hope you guys can keep up the good work and grow this open ecosystem 👍 
so, back to this PR... 
I found that it's kinda weird that the expand search button is THAT far from the search field, 

here is a proposal to make it closer, to respect the "locality" in UX...

What do you think?

PIC:
image

### DIFF
--- a/frontend/src/component/photo/toolbar.vue
+++ b/frontend/src/component/photo/toolbar.vue
@@ -14,6 +14,10 @@
                     @click:clear="clearQuery"
                     @keyup.enter.native="filterChange"
       ></v-text-field>
+      <v-btn icon class="p-expand-search" :title="$gettext('Expand Search')"
+             @click.stop="searchExpanded = !searchExpanded">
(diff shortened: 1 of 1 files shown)


graciousgrey:
Thank you for this pull request - we will have a look at it after our first release :)

isleshocky77:
The drop-down arrow is only loosely associated with the search.. It's more about the view layer with sorting and other filtering. And is more of a drop-down for that whole top bar which is why _I think_ it makes sense where it is; however, I would like to see a help button next to the search where you put this arrow which had a pop-up with advanced search commands. See the [search help page](https://docs.photoprism.app/user-guide/organize/search/#search-filter-overview) for what I'm referring to.

lastzero:
We are closing this PR because it does not match our user interface guidelines. Thank you for your contribution!

PR #662, opened 2020-11-27
fix(i18n): exclude false positive i18n extract

eMerzh:
vm.js should not be include in pot extract, 
let's ignore it


lastzero:
Thank you very much, and sorry for the wait! I've just applied this manually as there was a conflict after updating translations earlier.

PR #670, opened 2020-12-01
Added Slovak Language

Pheggas:
Added sk.mo ; sk.po
Edited translations.pot and options.js to show Slovak language options


graciousgrey:
Thank you! We try to merge it on the weekend :)

lastzero:
Still trying to understand why `frontend/src/locales/options.js` is shown as new / changed...

Pheggas:
> Still trying to understand why `frontend/src/locales/options.js` is shown as new / changed...

Well i don't know but i added only that Slovak translation option.

alexislefebvre:
It looks like the file https://github.com/photoprism/photoprism/blob/develop/frontend/src/options/options.js was copy-pasted in https://github.com/photoprism/photoprism/tree/develop/frontend/src/locales

lastzero:
@Pheggas I've just added your translation manually to avoid conflicts... thank you very much, and sorry for the wait!

Note that we've added a few more strings in the meantime, after optimizing the indexer and adding settings for stacks:

staking-file-settings

You're welcome to update the existing translations with a new pull request.

The backend needs to be translated as well:

https://docs.photoprism.org/developer-guide/translations/

PR #679, opened 2020-12-05
fr.po updated according to translation.pot

jean-louis67:


lastzero:
Merci! Pull request build checks fixed as well, all green :four_leaf_clover:

PR #704, opened 2020-12-13
French version update

jean-louis67:
Welcome to translation of _Slovak_ and _Unstacked_!


lastzero:
Merci! Should we keep Unstacked or use Single anyway? Hard to decide... or maybe "Don't stack" 😂

Issue #244, opened 2020-02-05
API: Provide Counts of Photos and Albums 


thielepaul:
The API should provide the count of photos and albums:
* [ ] total number of photos (available in `clientConfig.count.Photos`, but not exposed via the API)
* [x] number of photos per month (provided by `/moments/time` endpoint)
* [x] number of photos per album (provided by `/albums` endpoint
* [ ] number of archived photos
* [ ] number of albums (available in `clientConfig.count.Albums`, but not exposed via the API)

Photo counts should be consistent with the number of photos found via `/photos`:
* [ ] number of photos in albums should not include archived photos
* [ ] fix: if a duplicate is placed in the `Originals` folder and then indexed it will be included in the counts but not found via the `/photos` endpoint (similar behavior occurs for related photos)
* [ ] fix: if a photo is deleted from disk, it is still included in the count but not found via the `/photos` endpoint

graciousgrey:
What's the current status? Do you need help from us?

thielepaul:
The current status, as far as I know it, is as indicated by the checkmarks. I do not expect that I will work on PhotoPrism project in the near future, as I recently started a new full-time job and my spare time is limited.

If you think this issue is important, feel free to work on it, I currently do not plan to do so.

lastzero:
@thielepaul That's sad to hear, but of course your job has priority. Will the mobile app continue to be developed or shall we find new maintainers?

We were actually planning to get in contact with you (both) in preparation of our stable release. That's also why we just went through all open issues.

thielepaul:
Yes, that does also include the mobile app, so it would be good to find new maintainers for it.

I still believe that there is a need for a self-hosted open source alternative to Google Photos, so thank you for working on PhotoPrism!
I am also looking forward to check out the stable release of PhotoPrism :)

lastzero:
You're always welcome to contribute when you have time. Thanks for all the work you put into this and sorry for not reaching out earlier!

PR #723, opened 2020-12-19
Photo Search API count & offset

thielepaul:
This PR changes two things in the photo search backend:
- add a parameter `format=count` to get the complete number of results for a specific query
- fix `offset` and `count` parameters if `merged=true`

Both changes are necessary to lazy-load the photos in the mobile app. The current behaviour of the backend leads to this bug: https://github.com/photoprism/photoprism-mobile/issues/78
The newly added parameter will be necessary to filter photos in the app or implement a view of the moments as seen in the web backend.

If someone can help improving the query such that it is a subquery (without the variables `photos` and `ids`), this would be great! The current implementation is working fine as far as I can see, but I believe it should be possible with a subquery and this would probably lead to better performance.

This should fix most of the todos in https://github.com/photoprism/photoprism/issues/244.


lastzero:
We can take a look at the API together, after the release is done 🎄

thielepaul:
Closing in favor of #995

PR #775, opened 2020-12-31
Upgrading to tensorflow 2.x

thomaseizinger:
This is a draft PR for now primarily to explore and discuss what is needed to upgrade to tensorflow 2.x.

I've identified the following things so far:

- I think there isn't yet an official release of go-tensorflow for the 2.x version? All of the 2.x versions [here](https://pkg.go.dev/github.com/tensorflow/tensorflow@v2.4.0+incompatible/tensorflow/go?tab=versions) have an `+incompatible` tag. One can still depend on these but I wonder what that exactly means?
- The most recent [pre-built](https://www.tensorflow.org/install/lang_c) tensorflow C library is 2.3.1 but the go package for this version [fails to install](https://github.com/tensorflow/tensorflow/issues/41808). Hence I am using the go package for 2.1.0 but version 2.3.1 of the C library.

### Open questions

- I think CI is using the latest development image from docker hub and hence not picking up my changes to the Dockerfile. As such, I don't think it is actually linking against tensorflow 2.3.1. Should I be sending a separate PR for updating the development image?
- The tests seem to pass already by just upgrading to the latest go package. Maybe we can:
1. Merge this without the Dockerfile changes
2. Update the docker images to have the new tensorflow version
3. Use the new docker images in CI

Thinking about it, we should probably first release new versions of the docker images and then upgrade the tensorflow go dependency + docker image in one patch.

### DIFF
--- a/docker/development/Dockerfile
+++ b/docker/development/Dockerfile
@@ -80,7 +80,7 @@ RUN sh -c "echo 'deb http://download.opensuse.org/repositories/home:/pmdpalma:/U
 ENV LD_LIBRARY_PATH /root/.local/lib:/usr/local/lib:/usr/lib:/lib
 ENV TF_CPP_MIN_LOG_LEVEL 0
 RUN curl -L \
-   "https://dl.photoprism.org/tensorflow/linux/libtensorflow-linux-cpu-1.15.2.tar.gz" | \
+   "https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-cpu-linux-x86_64-2.3.1.tar.gz" | \
    tar -C "/usr" -xz
 RUN ldconfig
 
--- a/go.mod
+++ b/go.mod
@@ -50,7 +50,7 @@ require (
 	github.com/sirupsen/logrus v1.7.0
 	github.com/stretchr/testify v1.6.1
 	github.com/studio-b12/gowebdav v0.0.0-20200929080739-bdacfab94796
-	github.com/tensorflow/tensorflow v1.15.2
+	github.com/tensorflow/tensorflow v2.1.0+incompatible
 	github.com/tidwall/gjson v1.6.4
 	github.com/tidwall/match v1.0.2 // indirect
 	github.com/ulule/deepcopier v0.0.0-20200430083143-45decc6639b6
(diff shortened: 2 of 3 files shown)


lastzero:
Upgrading probably doesn't deliver any value at this point. In fact, we need to test a lot which mainly is work. Unclear what the outcome will be. Maybe a bit faster. We could use more / other models, which would be another couple of weeks for testing all labels, finding the right thresholds,... probably need to compile TensorFlow on our own again, based on recent experience. Haven't found anyone capable of doing to properly... worked 7 days (and nights) to get it done, extremely bad documentation. Nothing I'm eager to do on New Year's Eve (or after, while I'm trying to recover).

lastzero:
Note that I've contributed the patch that makes TensorFlow compile on ARM64, so I know a little bit about it. You may use their binaries, but they require AVX now - so won't work on many processors, even some Xeon won't work. In the past they didn't compile with AVX saying this is useless - now they offer no other version. Don't ask me why. Need to compile our own versions.

thomaseizinger:
> Haven't found anyone capable of doing to properly... worked 7 days (and nights) to get it done, extremely bad documentation. Nothing I'm eager to do on New Year's Eve (or after, while I'm trying to recover).

I haven't looked as much into it but from what I've been exposed to, I can sense that it is quite painful. There is no rush in getting this done, I just wanted to open a draft PR to show that I am working on it :)

The reason I am interested in upgrading to tensorflow 2.x is because I intend to run photoprism without the docker setup and instead managed by systemd on Manjaro on a raspberrypi 4.

Manjaro ships tensorflow in its repositories but only the 2.x version. As such, I was hoping on simplifying the setup here by basically only needing to install `go` to compile it myself locally and rely on the shared library being installed through the package manager.

> You may use their binaries, but they require AVX now - so won't work on many processors, even some Xeon won't work. In the past they didn't compile with AVX saying this is useless - now they offer no other version. Don't ask me why. Need to compile our own versions.

This is interesting, thanks for the insight! From what I can see, Manjaro [ships](https://discover.manjaro.org/packages/) with four different versions:

- normal
- with CPU optimizations (I guess that is the AVX stuff you are talking about)
- with CUDA
- with CPU optimizations + CUDA

lastzero:
First step will be to figure out what compiler to use with what C / C++ standard settings.

thomaseizinger:
> First step will be to figure out what compiler to use with what C / C++ standard settings.

Got it. I am looking to into updating the tensorflow docker image first.

lastzero:
Note that we need libtensorflow for macOS and ARM64 too, not just Linux x88_64.

Petru948 (approved):

thomaseizinger:
Going to close this because the first step will be updating the docker, not the code.

PR #816, opened 2021-01-04
Update pt_BR translation

OliveiraHermogenes:


lastzero:
Thank you very much! Next time, you don't need to add translations.json and translations.pot - will be created automatically :)

OliveiraHermogenes:
> Thank you very much! Next time, you don't need to add translations.json and translations.pot - will be created automatically :)

Yes, you're right. It's actually in the docs somewhere. I will try to keep that in mind. Thanks.

PR #833, opened 2021-01-06
Add a link to photoprism-mobile in readme

benmccann:
It took me a really long time to find that `photoprism-mobile` exists, which was a key requirement for me. It's not mentioned anywhere in the docs or readme currently. Hopefully this will help others discover it as well

Is it still true that you can add it to your home screen? I don't see the PWA add button on demo.photoprism.org


lastzero:
As the app is developed by contributors and we didn't have time to fully test it (there are in fact compatibility issues with our API), we advice end users to go for our PWA + PhotoSync instead for now. Once the native app is tested & fully working, we may add a link in our `README.md`. We pinned it on our [org page](https://github.com/photoprism/) so that advanced users and developers can find it.

Issue #826, opened 2021-01-06
Remove use of string-strip-html


benmccann:
I see a function [`addCaptionHTMLFn` in the source code](https://github.com/photoprism/photoprism/blob/9ec23813070f66412460e15d13966fa209765d10/frontend/src/common/viewer.js#L117). If I `grep` for it I find it only the declaration of it and don't see anything using it

```
$ grep -r addCaptionHTMLFn src/
src/common/viewer.js:      addCaptionHTMLFn: function (item, captionEl /*, isFake */) {
```

I think if it is unused we could remove `string-strip-html` which would get rid of 100kb. If `addCaptionHTMLFn` is used it'd still be nice to see if we can get rid of `string-strip-html` since it's rather heavy for what it does. Maybe we could just use `textContent` instead of `innerHTML`?

lastzero:
Without having time to take a detailed look: Think it is used for generating the HTML for the image titles / descriptions at the bottom of the lightbox?

benmccann:
Yeah, I think you're right. It looks like this is passed as an option to PhotoSwipe where it's used and that's why I didn't see it called anywhere in the code.

I do think we could probably use `textContent` instead of `innerHTML` here and get rid of `string-strip-html`. I've change the title of this issue accordingly

lastzero:
We use HTML here so that it's possible to add an optional icon / link for playing related videos - it's an ugly solution, but the only way I found without modifying PhotoSwipe.

benmccann:
We should be able to write our own method to sanitize the user input without using `string-strip-html` and still set the result to `innerHTML`.

Here are two options that would be lighter than having to import the `string-strip-html` library:

```
function sanitize(text) {
  return text.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;").replace(/"/g, "&quot;")
}
```

```
function sanitize(text) {
  var tmp = document.createElement('div');
  tmp.textContent = text;
  return tmp.innerHTML;
};
```

lastzero:
As long as there is no risk of code / html injection, sure :+1:

PR #836, opened 2021-01-06
Remove use of string-strip-html

benmccann:
Closes https://github.com/photoprism/photoprism/issues/826.

This removes a dozen npm packages from the frontend bundle by using our own `encodeHTML` function [as suggested in the OWASP cheat sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#rule-1-html-encode-before-inserting-untrusted-data-into-html-element-content).


lastzero:
Merging this now. Guess you're sure there is no risk of introducing security issues? :monocle_face:

Issue #825, opened 2021-01-06
Moment.js / time zone picker suggestions


benmccann:
15% of the bundle appears to be Moment.js, which is only used for getting time zone names in a single autocomplete. Some ideas on how we might be able to reduce the size here in order of preference:

- This select is barely usable since it has about 600 entries. I'm not sure what most of these options are. When I googled how many time zones there are [the first result says there's 38](https://www.timeanddate.com/time/current-number-time-zones.html). But the dropdown has 167 starting just with "America/". Can we find a shorter list to use?
- If we want this many time zones then it might be best if it were empty and the user typed in the time zone. Typically autocomplete suggestions will be loaded from the server as the user types in that case
- I wonder if this autocomplete component could dynamically import `moment-timezone` rather than statically importing it. The autocomplete could potentially be disabled until `moment-timezone` loads
- I wonder if there's a way to use `moment-timezone` without `moment` itself. It's only used for the list of time zone names and we don't need the actual date library

filiptronicek:
I used this in a project once, it's called TZDB and should be quite suited for this use-case. 
[@vvo/tzdb](https://www.npmjs.com/package/@vvo/tzdb)

PR #839, opened 2021-01-06
Remove moment.js from build

benmccann:
Closes https://github.com/photoprism/photoprism/issues/825. As suggested by @filiptronicek in that ticket, this was probably the easiest/fastest fix.

`webpack-bundle-analyzer` shows the size reduced from 5.97MB to 5.12MB


filiptronicek (approved):

lastzero:
Let's hope this is fully compatible with the zones our backend uses. Needs to be tested well.

benmccann:
You can see the full list here: https://github.com/vvo/tzdb/blob/master/time-zones-names.json

This appears to be a subset of the time zone names that Moment provided because Moment provided time zone names in many different formats whereas this list is more standardized

PR #841, opened 2021-01-07
Implement frontend code splitting

benmccann:
This reduces the size of the main bundle from 5.97MB to 4.95MB. Especially impactful is lazy loading the places page because `mapbox-gl` is very heavy. Arguably you might want to leave the photos page in the main bundle since that's the homepage, so I'm happy to leave that one in if you'd prefer


lastzero:
Wondering if splitting makes sense or if this means maintaining two frontends: one complete package for the mobile PWA and basically the same app just splitted for regular Web browsers at home?

lastzero:
The basic photos view should be included in any case. Also it would be good to know if this change might cause issues with caching so that modules from different builds may get mixed up in the browser?

lastzero:
Login should always be loaded as well.

benmccann:
I've updated this to include the Photos and Login in the main bundle.

I do need to fix this to work with caching still, which I was just about to mention as your wrote your comment :smile: It doesn't appear to me like caching is enabled on the web server right now, so at the moment that won't be an issue, but if we were to enable caching this wouldn't yet work with it. I'll look into that now

I'm not sure I understand your comment about the PWA. As far as I understand, PhotoPrism's "PWA" is just the same website with a bookmark on the home screen. I believe the same bundle is loaded on both desktop and mobile. Code splitting works just fine on mobile and there shouldn't be any impact. If you're talking about a future PWA that works in offline mode code splitting shouldn't be an issue there either. In that case the manifest can specify the list of JS files so that they're downloaded at install time. You can use something like `workbox-webpack-plugin` in that case

lastzero:
The PWA worker is for offline use and completely pointless in our case, as the index and images are stored on the server. I can implement one for you, but I'd rather invest my time in more productive ways ;)

Taking a look at caching soon.

benmccann:
Yeah, I agree the service worker is pointless. I wasn't suggesting we should use one. They're also a complete pain to deal with from my limited experience

benmccann:
I just updated this so that there shouldn't be any issues if caching is enabled. This PR is good on my end

lastzero:
Thank you very much! 🐳  

We'll test & maybe merge this later so that it doesn't block us from releasing important short-term fixes and improvements.

From what I understood, your original performance issues are gone since you enabled caching in the browser?

benmccann:
Yes, performance is better now (with the possible exception of the Vue rendering). This will still be nice for the first page load after each deployment of a new version

lastzero:
Currently refactoring the mosaic view using pure CSS as a proof-of-concept. Let's see if this helps the rendering.

benmccann:
@lastzero is there anything you need me to test because of the "please-test" label or is that there for you?

lastzero:
Yes, please test was for me... doing my best to get this done... emails stacking up in my inbox... just released another update... Friday is my birthday. Need a short break! 😉

StephenBrown2:
> Friday is my birthday. Need a short break! 😉

Happy early birthday! 🎂 🥳

benmccann:
+1 Hope you get to enjoy the day! :cake:

lastzero:
Sorry we couldn't make any progress with this so far. Currently working on Face Recognition because everyone is waiting for this.

PR #853, opened 2021-01-07
Update package-lock to fully remove moment

benmccann:
Both `Chart.js` and `moment-timezone` had dependencies on `moment`, so it wasn't fully removed with either of those PRs individually


lastzero:
:star: :star: :star: :star: :star:

PR #854, opened 2021-01-07
Update gitignore for static assets

benmccann:
It's helpful to me to be able to run `rm -rf assets/static/build/` while testing changes to the frontend. I can't easily do that right now because I end up deleting `assets/static/build/.gitignore`. I was hoping we could move the ignore of that directory into the main `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 /frontend/tests/screenshots
 /assets/nasnet
 /assets/nsfw
+/assets/static/build/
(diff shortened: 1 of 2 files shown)


lastzero:
I see your point. Need to be careful even with removing empty folders as this may cause unexpected permission or other issues, e.g. when a tool doesn't create the folder automatically for any reason.

lastzero:
Let's see if this works :)

PR #866, opened 2021-01-09
Migrate from deprecated slot-scoped syntax

benmccann:
I was having quite a bit of difficulty understanding some of the Vue documentation since we were using a deprecated syntax, so I wanted to update it to make it easier to follow. I tested this a little, but would appreciate if you could test it as well since you know the application better


lastzero:
It is shown like this in the Vuetify docs, which may now be harder to understand:

vuetify

From what I see, the hover element now also has the hover slot, unlike in the example above.

Does that make sense? I would expect this to cause issues, without having it tested yet.

benmccann:
Hmm. It looks like the Vuetify docs have migrated certain 2.6 syntax and not others. I'll close this to keep in sync with the docs

PR #867, opened 2021-01-09
Migrate templates containing both slot and slot-scope

benmccann:
The old syntax was deprecated in Vue 2.6. This makes it a lot easier to follow the Vue documentation

I tested that the page renders. Not sure if there's anything else I would need to test


lastzero:
Thank you! I'll merge this as well... working on acceptance tests right now, so bugs shouldn't go unnoticed 👀

PR #868, opened 2021-01-09
Migrate from deprecated slot syntax

benmccann:
This syntax appears to be one where Vuetify does use the new syntax in the docs at least in 2.0: https://vuetifyjs.com/en/getting-started/upgrade-guide/#activators


lastzero:
2.0 docs may be incompatible with 1.5, so we need to be careful and test well. Taking a look at this later, 6.30 am over here.

benmccann:
This is the syntax used in the 1.5 docs as well. See the example here: https://v15.vuetifyjs.com/en/components/lists#expansion-lists

lastzero:
Thanks a lot! Looks good to me 👍

PR #880, opened 2021-01-11
Fix lint errors in .vue files

benmccann:
Linting is setup for `.vue` files, but not actually running. This fixes almost all the errors. There's still two left that I might need your help with since I don't know Vue that well

You can make it run by changing the two commands in `frontend/package.json` to:
```
    "lint": "eslint src/ src/**/*.vue src/**/*.js *.js",
    "fmt": "eslint --fix src/**/*.vue src/**/*.js *.js .eslintrc.js",
```

### DIFF
--- a/frontend/src/component/footer.vue
+++ b/frontend/src/component/footer.vue
@@ -25,7 +25,7 @@
 
 <script>
 export default {
-  name: 'p-about-footer',
+  name: 'PAboutFooter',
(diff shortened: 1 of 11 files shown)


lastzero:
Thanks! Yes, I'm careful with auto fixing as it previously left broken JS behind. Can't spend time investigating this.

PR #897, opened 2021-01-14
Allow changing the album's day via the REST API

kvalev:


internal/form/album_test.go
```diff
@@ -35,6 +41,9 @@ func TestNewAlbum(t *testing.T) {
 		assert.Equal(t, "test notes", r.AlbumNotes)
 		assert.Equal(t, "newest", r.AlbumOrder)
 		assert.Equal(t, "default", r.AlbumTemplate)
+		assert.Equal(t, 2050, r.AlbumYear)
+		assert.Equal(t, 20, r.AlbumMonth)
+		assert.Equal(t, 37, r.AlbumDay)
```
alexislefebvre, line 46: What is day 37 of month 20?

kvalev: Just wanted to highlight that there are no sanity checks for the month and day values. If its bothering you, I can set them to something more plausible.

alexislefebvre: It will probably be rejected when persisting date in the database, but that's not a concern in this context. Thanks for your answer.

alexislefebvre (approved):

internal/form/album_test.go
kvalev: Well, surprisingly, it isn't - I did few tests and have few albums with `month > 12` right now.

lastzero:
Note that the dates may be overwritten by PhotoPrism and are not supposed to be changes by the user. Why set a date when it can be determined automatically based on the photos that are in the album? I know, some users want to configure and set everything manually, but most users don't have the time or need for this ;)

kvalev:
Yeah, I noticed that the indexing operation resets the dates for all albums, but given that the other date-related properties (year and month) can be changed via the REST API, I thought this PR causes no harm. Right now this is a workaround for the problem, that it's not possible to configure how the album's date should be computed, see https://github.com/photoprism/photoprism/discussions/894

lastzero:
When the date is overwritten anyways, this is not a workaround - or only for a very short period of time. In fact, we'd need to change a lot more so that PhotoPrism knows the date was set manually and does not change it anymore - like for photo metadata fields.

kvalev:
As I said, its just a workaround. I cant seem to find the place where the album's date is calculated based on the media inside, if you can point me to the right spot, I can make it configurable, np. In any case, the current album REST API is inconsistent - either it should be possible to update all date-related properties (albeit temporarily) or update none.

lastzero:
I guess there's always things to improve with a complex software like this 😉 

It's just that we can not do everything at the same time, and need to set priorities. If it helps, we can remove AlbumYear and AlbumMonth from the form for now - this was implemented before the date was automatically set.

For everything else, we shouldn't go for a "quick fix" but think about a broader concept first. Might add yet another database column for the exact time, except that we didn't want to touch the database tables for a while. Even then, some users will still not be happy because of time zone differences. Setting everything manually is also not a solution. In fact we might have to implement "batch edit" for albums for usability reasons. First, we should do batch edit for photos though.

This is a never ending story 🥳

lastzero:
When it comes to input validation, we also know there is room to improve - but also that what we have is mostly good enough for now. I'm the author of multiple input validation libraries for other programming languages e.g. https://github.com/symlex/input-validation. Success tests should assume valid input values as they otherwise will start to fail all at once when validation is getting stricter over time.

lastzero:
This function in the query package updates the album year, month, and day. We've intentionally used multiple fields so that it's possible to set values to ANY, like for birthdays which don't depend on the year (hopefully).

```go
// UpdateAlbumDates updates album year, month and day based on indexed photo metadata.
func UpdateAlbumDates() error {
	switch DbDialect() {
	case MySQL:
		return UnscopedDb().Exec(`UPDATE albums
		INNER JOIN
			(SELECT photo_path, MAX(taken_at_local) AS taken_max
			 FROM photos WHERE taken_src = 'meta' AND photos.photo_quality >= 3 AND photos.deleted_at IS NULL
			 GROUP BY photo_path) AS p ON albums.album_path = p.photo_path
		SET albums.album_year = YEAR(taken_max), albums.album_month = MONTH(taken_max), albums.album_day = DAY(taken_max)
		WHERE albums.album_type = 'folder' AND albums.album_path IS NOT NULL AND p.taken_max IS NOT NULL`).Error
	default:
		return nil
	}
}
```

kvalev:
I see, its done in sql and not in go. Before I open a new PR, let me run my idea by you. I can also open an issue, instead of using this PR as a discussion thread.

My idea is to have a new setting (most likely a global one, it does not make sense to have it per album), which should determine how the album date should be computed:
- either latest photo (as it currently is), meaning use the `MAX` function
- or earliest photo, meaning use `MIN` function

Computing the min/max can also be done in go, so that its compatible with sqlite as well

lastzero:
That might work. Wouldn't use Go for this as this may become very inefficient. Might also be possible use the average date similar to what is suggested on https://stackoverflow.com/questions/1021947/how-can-i-make-an-average-of-dates-in-mysql:

```
   from_unixtime(
        avg(
            unix_timestamp(date_one)-unix_timestamp(date_two)
        )
    )
```

Note that adding more settings requires those to be translated and documented well, which is significantly more work than just adding a flag.

PR #899, opened 2021-01-14
Add few more places to the countries mapping

kvalev:


lastzero:
Excellent, thank you very much! ⭐️

PR #903, opened 2021-01-15
Estimate album country

kvalev:
Estimate the country for a given album, based on the photos in the album.

### DIFF
--- a/internal/entity/album.go
+++ b/internal/entity/album.go
@@ -285,9 +285,14 @@ func (m *Album) String() string {
 	return "[unknown album]"
 }
 
-// Checks if the album is of type moment.
+// IsMoment checks if the album is of type moment.
 func (m *Album) IsMoment() bool {
 	return m.AlbumType == AlbumMoment
+}
+
+// UnknownCountry checks if the album has an unknown country.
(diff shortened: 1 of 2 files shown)


kvalev:
@lastzero Before I spend too much time on that (and getting up to speed with golang and the codebase), let me know whether you agree in principal with this feature.

lastzero:
The idea is good. Should be implemented with SQL for performance reasons though, so that batch updates are possible. Looping through all photos using Go is very inefficient.

The issue here is that there may be photos from multiple countries, so you either need to weight them, set a threshold or only set a country if all photos are from the same country. Still somewhat unsolved in my brain, so not implemented yet as you can see :)

kvalev:
I see, implementing this is SQL will be a nightmare and quite messy, so I will close this PR. There is always the option to automate this using the REST API, if someone is eager to do so.

For the actual implementation, I did it the most conservative way possible - set the country, only if all (tagged) pictures are from the same country.

lastzero:
> I see, implementing this is SQL will be a nightmare and quite messy

Now you're exaggerating 😉

PR #908, opened 2021-01-16
Dockerfile: reduce number of layers

alexislefebvre:
We can use one Docker layer instead of 5:

```
7 /bin/sh -c echo 'Acquire::Retries "10";' 204 B
8 /bin/sh -c echo 'APT::Install-Recommends "false";' 217 B
9 /bin/sh -c echo 'APT::Install-Suggests "false";' 213 B
10 /bin/sh -c echo 'APT::Get::Assume-Yes "true";' 216 B
11 /bin/sh -c echo 'APT::Get::Fix-Missing "true";' 216 B
```

See the layers here:https://hub.docker.com/layers/photoprism/photoprism/20210111/images/sha256-1b5c2e4e618dcef9a50c29cf14f6197c1864d5626fea713fb37edd95ff9ac101?context=explore

### DIFF
--- a/docker/photoprism/Dockerfile
+++ b/docker/photoprism/Dockerfile
@@ -14,11 +14,11 @@ FROM ubuntu:20.04
 ENV DEBIAN_FRONTEND noninteractive
 
 # Configure apt-get
-RUN echo 'Acquire::Retries "10";' > /etc/apt/apt.conf.d/80retry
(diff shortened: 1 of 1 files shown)


lastzero:
Thanks, good idea! Remember to select develop as base for the PR next time 👍

PR #913, opened 2021-01-17
Separate dependencies and devDependencies

benmccann:
This makes it much easier to see what's being included in the application vs what's needed only for the build system and tests


docker-compose.drone.yml
```diff
@@ -10,7 +10,6 @@ services:
     depends_on:
       - photoprism-db
     volumes:
-      - "~/.cache/npm:/root/.cache/npm"
```
lastzero, line 13: Wouldn't this increase build times by a lot?

benmccann: I was able to remove that change

PR #918, opened 2021-01-18
Additional photo location unit tests

kvalev:
~I had a folder with non-geotagged photos, where the location was estimated based on the folder's name. After manually geo-tagging them, the location was not set after re-indexing the folder, because `estimated` locations are not allowed to be overwritten, which was probably not intended.~

Add few additional unit tests to make sure the photo location is set correctly.


lastzero:
Should use our source priority lookup table... See title and description.

lastzero:
Let us know if these changes work for you :)

kvalev:
Oh, I didnt see that you pushed a fixed already. My solution is largely the same, but it allows to overwrite the existing photo location (if for example you correct the geo information for an image directly in the exif data). And there are few more unit tests as well.

lastzero:
The location should be updated automatically after lat / lng have been changed. If this is not the case, that's a bug? Of course depending on the specific API / function you use for this. Make sure to also update the location source so that PhotoPrism knows this is a manual change. Otherwise, you may not be able to overwrite existing metadata.

internal/entity/photo.go
```diff
@@ -911,7 +911,7 @@ func (m *Photo) SetCoordinates(lat, lng float32, altitude int, source string) {
 		return
 	}
 
-	if SrcPriority[source] < SrcPriority[m.PlaceSrc] && m.HasLatLng() {
```
lastzero, line 914: `m.HasLatLng()` is there so that missing information can be added from sources that don't have priority - or `PlaceSrc` is set incorrectly, e.g. because people messed with their database.

kvalev: I dont know about the other users, but if I personally notice that some of my photos are geo-tagged incorrectly, I will not fix the location directly in PhotoPrism, but rather fix the geotag in the photo itself. With the existing implementation, the updated photo location will be ignored, as the photo was previously geotagged. This is basically the only reason why I removed the `m.HasLatLng()` check.

lastzero: This is strange as we fixed and tested this, probably before our release in December. Did you change the location manually in the UI as well? That won't be overwritten by changes in the file metadata.

The lookup table clearly shows that metadata has a much higher priority than estimates:

https://github.com/photoprism/photoprism/blob/develop/internal/entity/src.go#L22

kvalev: Nevermind, I cant read 🤦  I tested it with the latest `develop` and works as expected...

lastzero:
Thank you very much & sorry for the wait! :+1:

PR #924, opened 2021-01-19
Set cover image for album/calendar/folder/moment

### LINKED ISSUE #383: How to set the hero image for a collection
It looks like (apart from Stacks which have a 'primary' photo) it isn't possible to set the hero image for a collection of photos (e.g. Album, Calendar month 'Album', Moment, Folder).

Some ideas:

- Support the 'primary' concept that is used with Stacks.
- Use favourites? If multiple favourites, then either cycle through them (animated? or pick one at random when the collection is loaded for a particular view), or maybe render a 2x2, 3x3, 4x4 (or beyond?) mosaic.

kvalev:
Tasks:
- [x] backend changes
- [x] handle cover caching
- [x] UI changes
- [x] UI styling
- [ ] unit tests
- [ ] UI/integration/e2e tests
- [ ] translations

Closes #383


kvalev:
@lastzero I think I am mostly done, would you mind taking a look. Tests and translations are missing, tests should be doable, but I am not sure how to handle the translations? Should I wait until someone translates it into every supported language? Do I just translate it with DeepL?

kvalev:
@lastzero Did you had a chance to take a look?

lastzero:
Not yet, sorry. Might take a couple of days. Had to take care of a few bug reports and UX issues first (full log didn't fit on screen):

commits

graciousgrey:
@kvalev I checked out your branch (cover-image) to test. I see the button in the UI but the album cover is not updated for me and I get no error

kvalev:
If you refresh the page (browser reload, not pp reload) you should see the updated cover. For some reason pp does not fetch the updated covers for a page, which was already visited (e.g. going from `folders` to a specific folder/album, updating the cover there and then going back to `folders`).

graciousgrey:
Ok, I can confirm it works when reloading and emptying the cache of the browser.

Thanks for your work! 

I still see the following to dos:
* Album cover MUST update when clicking reload in PhotoPrism or directly
* The option in the context menu MUST only be shown inside of albums/moments/calendar/states/folders and originals and not in sections like search or video because those do not have a cover
* Nice to have: Set cover of a label as well

kvalev:
The first two are done. The third one should be fairly easy, but it requires extending the database schema and my understand is that you prefer not to make any database changes at this time.

graciousgrey:
Thank you! I will have a look later today or tomorrow :)

graciousgrey:
Functionality looks good :) @lastzero will have a look at the code as well as soon as possible, we need to find out why the build is failing

Thanks for your work!

kvalev:
> Functionality looks good :) @lastzero will have a look at the code as well as soon as possible, we need to find out why the build is failing

The failing build has been fixed

graciousgrey:
Thank you, unfortunately we won't find time in March. In the meantime you can send us an email to hello@photoprism.app so that we can send you instructions on how to enable sponsor/contributor features, if you like :)

lastzero:
We have thumb columns in our database for this now, so that no UID lookup is required anymore, plus you can set a custom cover if you change the src to any value <> ''

fehervalentin:
Hi! What about this PR? I am looking for this functionality too. What was the issue with the branch/commits?

> plus you can set a custom cover if you change the src to any value <> ''

@lastzero you mean directly in the database, right?

graciousgrey:
@q3flat 

The problem is missing time and resources.

We have high quality standards and merging pull requests requires a lot of work and time on our side.

- The feature has to be fully implemented - in the case of "set cover image", this means the ability to set cover images for all types of albums and people
- Translations need to be created
- Docs need to be updated

We would love to grow our team to be able to work faster but we can't even cover our own expenses yet --> https://github.com/sponsors/photoprism

To set a cover image in the database, you need to add the filehash of the file you want to have as cover in the "thumbs" column. Additionally you need to set "thumb_src" to "manual".

lastzero:
Check out the Release Notes if you wonder why we didn't have time yet:

https://docs.photoprism.org/release-notes/

fehervalentin:
Thanks for the responses! I understand that.

So technically PP is capable of this "feature" by now. It is great for me. Thanks!

(My previous nickname was q3flat)

Issue #832, opened 2021-01-06
iOS: Multi-select using long click doesn't work


adan89lion:
Hi,

I’ve recently switched my workflow from macOS to iPadOS, and encountered multi-selecting issues with Safari on mobile devices.

Step to reproduce:
1. Open PhotoPrism instance on iOS, with Safari or Chrome [1]
2. Open Review tab (or photo stream, or any album)
3. Select the first image
4. Long press the last image to make a selection
5. Doesn't select images, but shows out text selection frame

Expected behaviour:
Make a selection from the first selected image to the long pressed image.

What went wrong?
It calls out Safari’s text selector

Did it work before?
It works perfectly on Android’s Chromium based browsers (tested on Chrome, Brave and Samsung Internet) and Fennec. This issue does not apply to desktop based systems (since we use Shift and Control to select massively)

Other notes:
N/A

[1] Browsers on iOS / iPadOS are all using WebKit, so any browser will have the same issue.

lastzero:
Might be handling DOM / JS events differently. Needs to be debugged and fixed in this case.

lastzero:
Just ordered an iPad so that I can test this properly:

![ipad](https://user-images.githubusercontent.com/301686/104768328-c00c1080-576d-11eb-80af-4216bc545dc3.jpeg)

If you're not a sponsor yet, consider supporting us financially. Thank you! :heart:

alexislefebvre:
Have you been heard about Browserstack? :smiley: https://www.browserstack.com/list-of-browsers-and-platforms/live

lastzero:
On Fri, 15 Jan 2021, 22:28 Alexis Lefebvre <notifications@github.com> wrote:

> Have you been heard about Browserstack? 😃
> https://www.browserstack.com/list-of-browsers-and-platforms/live
>
Browserstack is not a UX testing tool, and also not completely free from
what I remember.

lastzero:
Browserstack is $348 per year minimum:

![Screenshot 2021-01-15 at 22 51 09](https://user-images.githubusercontent.com/301686/104782243-2ef46400-5784-11eb-8ce7-7bf430986085.png)

alexislefebvre:
(I removed my previous comment shortly after posting it, I thought it wasn't heplful, and you proved it)

lastzero:
[...]

PR #962, opened 2021-01-26
Better scrolling experience on mobile devices

kaixinguo360:
When the touch point is on the ❤️ button, scrolling will be blocked. This PR fixes this problem.

### DIFF
--- a/frontend/src/component/photo/cards.vue
+++ b/frontend/src/component/photo/cards.vue
@@ -56,9 +56,7 @@
 
             <v-btn :ripple="false" :depressed="false" class="input-open"
                    icon flat absolute
-                   @touchstart.stop.prevent="openPhoto(index, true)"
-                   @touchend.stop.prevent
-                   @touchmove.stop.prevent
+                   @touchend.stop
                    @click.stop.prevent="openPhoto(index, true)">
               <v-icon color="white" class="default-hidden action-raw" :title="$gettext('RAW')">photo_camera</v-icon>
               <v-icon color="white" class="default-hidden action-live" :title="$gettext('Live')">$vuetify.icons.live_photo</v-icon>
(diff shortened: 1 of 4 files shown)


lastzero:
Did you test on iOS? :)

kaixinguo360:
Sorry, no, because I only have an Android phone... :( But I have run `make test` and no new fail. I'm very sorry if this change has caused any problems on iOS

lastzero:
See https://github.com/photoprism/photoprism/issues/832

kaixinguo360:
Thanks for your reply. I tried Browserstack and found it's difficult to test long click event because of my poor network, so I'm not sure if the multi-selecting is still working on iOS (the other features are ok). I think I should borrow an iphone to test it tomorrow

kaixinguo360:
I just test it on my friend's iPad and found some bugs. Now I have fixed them. I believe there is no problem now.

lastzero:
Thanks! Taking a look at this ASAP, may take a while given the size of our todo list 😬

Marenz:
I had a similar problem on android touch screen but instead of hearting, it would open the picture while I was scrolling. Should I make an extra issue for that?

lastzero:
We will improve this anyway, just not short term as there are other issues - like face recognition - waiting. Please be patient.

lastzero:
In the meantime everyone is invited to test this PR on **iOS and Android** and let us know if this is good for you. I spent a ton of time and found no perfect solution for all devices and operating systems so far.

aberres:
> I had a similar problem on android touch screen but instead of hearting, it would open the picture while I was scrolling. Should I make an extra issue for that?

Came around while looking if an issue for this very problem exists.

On my wife's OnePlus 5 Photoprism with Chrome is pretty much unusable as scrolling either selects a favorite or opens the image under the cursor.

What is problematic from a developer's perspective: The very same Chrome version on my OnePlus 6T does not have the problem.

With Firefox I am running only into the problem initially described here.

When I find the time to build a Docker image with this change I can give it a spin.

lastzero:
I guess nobody had time to test this with different devices so far?

kaixinguo360:
I think it's time to close this PR... :(

heikomat:
> I think it's time to close this PR... :(

Why?

Scrolling feels very weird when it sometimes just doesn't work (when the scroll starts on an icon)

I know it's old, but what changed? the issue still persists

alexislefebvre:
It may be fixed by #2292

JulianEggers:
I also run into this issue and would suggest reopening it.

PR #983, opened 2021-01-31
Add Italian translation

LucaDiba:
Added translation for Italian language.
There is still something to be translated - I wasn't 100% sure about the context, I'll look for those sentences in PhotoPrism in the future to better understand what do they mean.


graciousgrey:
Thank you :)

PR #996, opened 2021-02-03
Add option to hide Video from sidebar

nesb0t:
**What it does do**: Hides "Videos" from the sidebar. Can be toggled through docker-compose or command line. 

**What it doesn't do**: Does not prevent accessing the `/videos` route if it is typed manually. Does not prevent PhotoPrism from processing videos or truly prevent access from a security-perspective. 

**Why**: I was creating a PhotoPrism to share with family members and didn't want them to be confused on why there were no videos. 

**Future plans**: 

- Adding an option to toggle this feature through the Setting UI.
- I also have plans to add options for hiding other sidebar options to overall clean it up if you're only using it for a certain use case like I am. 
- I may work on truly preventing access to videos if there is a desire for it. 

**Usage:**: 

- Command line: `./photoprism --disable-videos start`
- docker-compose: `PHOTOPRISM_DISABLE_VIDEOS: "true"`

videos-before-after


lastzero:
We could actually hide it automatically when there are no videos!?

benmccann:
>I also have plans to add options for hiding other sidebar options to overall clean it up if you're only using it for a certain use case like I am.

I think it'd be useful to consider everything at once in a holistic manner, so I'm curious what else you'd want to hide. It's possible that having lots of options might become difficult to maintain. I wonder if we should just create one option that lets you specify the menu layout. Right now the menu is hardcoded, but it could be turned into a vue component and we could read it from a json data structure. I potentially could take a stab at this, but am hoping my Vuetify 2 PR could be merged first so that I don't have to deal with merge hell :smile:

nesb0t:
@lastzero 
> We could actually hide it automatically when there are no videos!?

I think that's a great idea, and I see you already submitted a commit request for it. I think that's a great idea! I could still see a benefit to choosing to hide that section altogether though, even if you actually have videos. I could be thinking of using PhotoPrism in a way that isn't intended, but let's say for example you have two different PhotoPrism deployments for whatever reason (maybe one is public and one is private, or something like that), but they reference the exact same library of pics and videos. You might want to hide videos from one of those sites, but not in the other. It could also be useful when multi-user is an option. But again, it's just a thought. 🙂 

@benmccann 
> I think it'd be useful to consider everything at once in a holistic manner, so I'm curious what else you'd want to hide. It's possible that having lots of options might become difficult to maintain. My preference would be that we come up with a good solution and don't allow it to be configured at all. But if we do want configuration I wonder if we should just create one option that lets you specify the menu layout. Right now the menu is hardcoded, but it could be turned into a vue component and we could read it from a json data structure. I potentially could take a stab at this, but am hoping my Vuetify 2 PR could be merged first so that I don't have to deal with merge hell 😄

You're very right about how difficult it would/could become to maintain. For reference, this is actually how my site looks right now:

image

That's the only thing in the sidebar haha. Obviously at this point I could just hide the sidebar completely, but I wanted to leave an easy way for people to return back to the list of Albums. I think it's a great idea to create an option that lets you pick and choose the items that show in the menu. I have played around with Vue some in the past but overall I'm new to it. It's been pretty easy to pick up though. I don't expect I'll be submitting a PR any time soon with that component working, but I am going to start researching and playing and seeing if I can come up with something. Even if it doesn't make its way in to PhotoPrism, it's still a fun challenge. Thanks for the idea!

I totally understand about waiting on your PR to be merged haha.

nesb0t:
@lastzero I tested your commit and it worked as expected. No Videos visible in sidebar when there were none. Showed up when I uploaded a video. Went away again after removing the video. 👍

lastzero:
Thanks! Might be even better to support custom templates or CSS for special use cases when the standard UI doesn't fit.

lastzero:
There's a feature flag for this now so that the navigation doesn't "flicker" when you open the app.

PR #1027, opened 2021-02-13
Indexer status REST API

kvalev:


internal/api/index.go
```diff
@@ -16,6 +16,29 @@ import (
...
+		conf := service.Config()
+
+		if !conf.Settings().Features.Library {
+			AbortFeatureDisabled(c)
+			return
+		}
+
+		ind := service.Index()
+
+		c.JSON(http.StatusOK, gin.H{"running": ind.IsRunning()})
```
lastzero, line 38: I'd change "running" to "busy" so that it matches the backend name. True will also be returned when the importer is running, which blocks the main indexer as well.

internal/photoprism/index.go
```diff
@@ -52,6 +52,11 @@ func (ind *Index) thumbPath() string {
 	return ind.conf.ThumbPath()
 }
 
+// IsRunning checks whether the indexer is currently running
+func (ind *Index) IsRunning() bool {
```
lastzero, line 56: I'd rename `IsRunning()` to `Busy()` to avoid different names for the same flag.

lastzero:
Thank you very much! Are you using this for a custom CLI script?

internal/api/index.go
```diff
@@ -16,6 +16,29 @@ import (
 	"github.com/photoprism/photoprism/pkg/txt"
 )
 
+// GET /api/v1/index
+func IndexingStatus(router *gin.RouterGroup) {
+	router.GET("/index", func(c *gin.Context) {
+		s := Auth(SessionID(c), acl.ResourcePhotos, acl.ActionRead)
```
lastzero, line 22: I'd use `ActionUpdate` as guests, who can read photos, probably don't need access to this.

internal/api/index.go
```diff
@@ -16,6 +16,29 @@ import (
...
+
+		if s.Invalid() {
+			AbortUnauthorized(c)
+			return
+		}
+
+		conf := service.Config()
+
+		if !conf.Settings().Features.Library {
```
lastzero, line 31: Checking for `ActionUpdate` (see above) should be sufficient for our use case, you can remove this `if` block then.

kvalev:
> Thank you very much! Are you using this for a custom CLI script?

Yeah, I have a nightly script, which reindexes any new data and fixes few things - album date, sort order, country etc..
From time to time I have to reindex my library from scratch when I notice issues with my data. This takes few days, so in this period I want to disable the reindexing.

PR #1029, opened 2021-02-13
Support .mts and .mpo files

kvalev:


alexislefebvre:
Can your browser read these files?

lastzero:
Thanks! What are these file types used for?

kvalev:
`mts` is a video format and `mpo` is a stereoscopic image format, that contains 3D images

lastzero:
Can PhotoPrism display / index those? What happens if you add them?

kvalev:
Yeah it seems so. I guess the underlying framework supports both formats. Videos are detected and shown correctly. The `mpo` file is essentially a wrapper for two (or more) stereo images. Here only the first one is read, but that is good enough for me, as I dont think it is possible/easy to actually show a 3D image.

PR #1079, opened 2021-02-19
Add Norwegian translation

ohjohnsen:
Adding Norwegian translation for the frontend.  Was not able to build and test locally, so the translation is untested.


lastzero:
Thank you very much! `frontend/src/locales/nb.mo` should not be empty, but guess we can fit it :)

ohjohnsen:
Hmm..  I don't understand why the file is empty.  I see that in the "Files changed" tab in the PR that GitHub says the .mo file is an "Empty file".  But in both my local feature branch in my forked repo and when just downloading an export of the feature branch from GitHub, the nb.mo file should be around 32kB size.

I used Poedit to translate though, so it could be that I didn't do a step before committing the translation.  Please advice then and I will follow up.

lastzero:
norsk

ohjohnsen:
Awesome! :)

ohjohnsen:
Can you rename it to "Norsk (Bokmål)"?  We have two written Norwegian languages.  "Norsk (Bokmål)" and "Norsk (Nynorsk)".  Nynorsk is used extensively on the western part of Norway amongst other places.  My wife is from western Norway (we live in Bergen), and I've more or less recruited her to add a Nynorsk translation later on. :-)

lastzero:
Saw that and wasn't sure how to handle it... thanks for the hint!

lastzero:
Done!

lastzero:
Demo will be updated in ~30 min so that you can test it there as well :rainbow:

PR #1084, opened 2021-02-21
French translation update

jean-louis67:
Translated words: Italian, Norwegian, Korean


lastzero:
Merci!

FYI: We currently don't use the language names for anything since the native names are used for the settings dropdown.

jean-louis67:
Hello Michael

Yes I had noted this point.
However, I prefer to have a .po file corresponding in every way to the pot file. It's easier to manage.
And then maybe the language names will reappear in another view later!

Issue #85, opened 2018-12-21
Use pig.js for the gallery/image feed


Sohalt:
I think using https://github.com/schlosser/pig.js for the gallery would be really nice. It is similar to Google photos, providing a better user experience than a boring equal sized grid imo.

Sohalt:
Building the Google Photos Web UI – Google Design – Medium - https://medium.com/google-design/google-photos-45b714dfbed1 is a nice read on the Google photos grid.

lastzero:
Thank you for the hint... Looks nice! A contributor could implement that as additional view type 💎

rkkoszewski:
Hi, this sounds interesting, I might help to implement this once I've read trough the Contribution document (Is there any information regarding view types in there?). I actually was about to start a similar open source project like photoprism, but this seems to be a pretty solid project.

marcobeierer:
Hi, I'm also working on a closed-source photo application aimed at professional photographers.

I also had a look at pig.js and the medium article at the time I have implemented the same behavior for my application, but the pig.js demo seemed to be a bit laggy with 4k resolution and the solution in the article seemed unnecessarily complicated.

I just wanna let you know that it is relatively simple to implemented this behavior with pure CSS by using the flexbox technique. The basic concept is described in the following GitHub issue: https://github.com/xieranmaya/blog/issues/6

rkkoszewski:
@marcobeierer I agree with you, but I actually already tested both pig.js (slightly lagy), and also the same flexbox example you sent. Flexbox is way faster, but if you try to show 1000 - 10000 pictures or more at the same time it was even lagging even more (And thats on my desktop PC, I wonder about mobile). But I think a hybrid approach would be best, where flexbox handles the layout but the pictures get loaded/unloaded depending on the view, that should keep flexbox swift, altough if flexbox has too many nodes it also uses cpu quite heavily.

danielo515:
It definitely feels snappier than the current UIs. I think the current UI needs better handling onf lazy loading, but I think that is a topic for a different issue.

danielo515:
> I just wanna let you know that it is relatively simple to implemented this behavior with pure CSS by using the flexbox technique. The basic concept is described in the following GitHub issue: xieranmaya/blog#6

[...]

PR #1089, opened 2021-02-23
Add additional view type using pig.js

seond:
This PR is to create a new feature that was requested in #85. The ask of issue #85 was to use pig.js to create a feed that is a lot easier to scroll through all the images with better performance. The result turned out pretty nice. The page loads 1500+ images I have almost instantly and I can freely scroll to any position of the huge stack of pictures, then the page fluently lazy load the requested part of the page smoothly.

Check this demo to see how smooth it is.
https://drive.google.com/file/d/1Gku6aWRzShX4Ciz3xg68XHIvDCGv_Y8v/view?usp=sharing

Here's a few implementation highlights:
- The page replicates the existing `/photos` page in terms of 1. how it handles the filters/url params 2. all the actions on an opened image
- Instead of infinite scroll mechanism, the page loads all the images that subject to the applied filter based on the batch size. API calls are made consecutively instead of being made simultaneously.
- The page creates multiple chunk `div`s as many as the number of batches then converts each div into a `pig.js` object so it can prevent a dom object from being humongous.
- The only event that could be handled on a photo item was `onClick` due to the limitation `pig.js` has. The library does not provide any other event handler options. You can still like and the rest of actions by opening the image by clicking on it.

The page is accessible at `/instant` and I intentionally skipped adding the page in the menu to leave it on to the maintainers to decide how we want to utilize or even improve/develop this page in a long run.

This is my first contribution to PhotoPrism and I fell in love with the project as I was trying and working on it. Great work, and big thanks to all the efforts on this fantastic project.


lastzero:
Thank you very much for submitting this PR & sorry for the wait!

- Loading all search results is generally not a good idea as some users may have millions of pictures.
- Users trying different search filters / keywords are going to trigger many reloads.
- Results should automatically refresh after new files have been added / uploaded as well. 

Loading all pictures is very inefficient in this case. It's great in case the user scrolls very far, doesn't change search filters, and there are no index updates. A solution might be to disable those features on this page and maybe also reduce the amount of metadata if the view doesn't need it, similar to *Places*.

The pig.js library should be included via NPM (see https://github.com/photoprism/photoprism/blob/develop/frontend/package.json) and not directly added to our repository.

lastzero:
Thanks a lot for your contribution! I have to close this PR as there is still a lot of work to do and we haven't received any feedback for a long time. pig.js may be something we can use for a static export in the future though.

PR #1111, opened 2021-03-02
Create country moments

kvalev:
There are already moments for country per year (e.g. Germany 2020, France 2019), this PR adds moments (under `Places`) per country (e.g. Germany, France).

### DIFF
--- a/internal/config/client.go
+++ b/internal/config/client.go
@@ -324,7 +324,7 @@ func (c *Config) UserConfig() ClientConfig {
 
 	c.Db().
 		Table("albums").
-		Select("SUM(album_type = ?) AS albums, SUM(album_type = ?) AS moments, SUM(album_type = ?) AS months, SUM(album_type = ?) AS states, SUM(album_type = ?) AS folders", entity.AlbumDefault, entity.AlbumMoment, entity.AlbumMonth, entity.AlbumState, entity.AlbumFolder).
+		Select("SUM(album_type = ?) AS albums, SUM(album_type = ?) AS moments, SUM(album_type = ?) AS months, SUM(album_type = ?) AS states, SUM(album_type = ?) AS countries, SUM(album_type = ?) AS folders", entity.AlbumDefault, entity.AlbumMoment, entity.AlbumMonth, entity.AlbumState, entity.AlbumCountry, entity.AlbumFolder).
 		Where("deleted_at IS NULL AND (albums.album_type <> 'folder' OR albums.album_path IN (SELECT photos.photo_path FROM photos WHERE photos.deleted_at IS NULL))").
 		Take(&result.Count)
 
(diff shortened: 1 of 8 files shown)


internal/config/client.go
```diff
@@ -335,11 +335,6 @@ func (c *Config) UserConfig() ClientConfig {
 		Where("deleted_at IS NULL").
 		Take(&result.Count)
 
-	c.Db().
-		Table("countries").
-		Select("(COUNT(*) - 1) AS countries").
-		Take(&result.Count)
```
lastzero, line 341: Should not be removed / replaced as the country album count may be less. Moments will only be created if enough good photos exist for a certain filter, in this case a country code.

internal/query/counts.go
```diff
@@ -57,10 +57,6 @@ func (c *Counts) Refresh() {
 		Where("deleted_at IS NULL").
 		Take(c)
 
-	Db().Table("countries").
-		Select("(COUNT(*) - 1) AS countries").
-		Take(c)
```
lastzero, line 62: Same as above: Should not be removed / replaced as the country album count may be less. Moments will only be created if enough good photos exist for a certain filter, in this case a country code.

lastzero:
Thanks for your PR & sorry for the wait! 

My main concern is that it might cause confusion when the country albums list doesn't match the countries dropdown, for example because there are not enough (or no) good photos to create an album. Users previously complained about empty moments albums, so showing empty country albums as a "placeholder" is not ideal as well.

kvalev:
> Thanks for your PR & sorry for the wait!
> 
> My main concern is that it might cause confusion when the country albums list doesn't match the countries dropdown, for example because there are not enough (or no) good photos to create an album. Users previously complained about empty moments albums, so showing empty country albums as a "placeholder" is not ideal as well.

An easy solution for this is to set the `threshold` for creating the moment to 1. This way the countries will match up nicely and there would be no need to do the changes you suggested above. A downside is of course that this might lead to country albums with only few pictures, but I can live with that.

lastzero:
Yes, but then you have moments with just one photo. Moments are supposed to be clusters of photos. One photo is not a cluster / album. Removing the threshold is going to create a lot of noise for most users.

kvalev:
> Yes, but then you have moments with just one photo. Moments are supposed to be clusters of photos. One photo is not a cluster / album. Removing the threshold is going to create a lot of noise for most users.

So what would be the solution? Either keep the existing threshold and the number of countries is potentially different or reduce it to 1 and then there are moments with very few pictures. The third option will be to reject this PR, which I wouldnt mind as I have already merged it in my own fork.

kvalev:
Any decision on this?

lastzero:
We'd like to merge this, but still not sure about possible side effects and edge cases... don't want to introduce bugs. We'll come back to this when we had time to think (and test) it through.

PR #1129, opened 2021-03-10
Add the GeocodingApi option to allow switching between places and osm for reverse location lookups.

newswangerd:
- `GeocodingApi: places` configures the app to use the photoprism places api.
- `GeocodingApi: places` configures the app to use the Open Street Map api.


graciousgrey:
Thank you :) We will review your PR as soon as possible.

lastzero:
Sorry, but OSM doesn't allow using their API for production apps. We had a huge discussion. In addition, it's sometimes down and much slower - so not recommended and supported anymore from our side. We've used it previously for development as intended and allowed by their terms.

PR #1140, opened 2021-03-20
Reduce memory usage in indexer and importer

aaronjwood:
The empty struct type uses no storage. Correct me if I'm wrong but these areas seemed like a bit of a hot path, so thought it'd help to reduce some memory pressure here.

### DIFF
--- a/internal/photoprism/import_worker.go
+++ b/internal/photoprism/import_worker.go
@@ -152,7 +152,7 @@ func ImportWorker(jobs <-chan ImportJob) {
 				continue
 			}
 
-			done := make(map[string]bool)
+			done := make(map[string]*struct{})
 			ind := imp.index
 			sizeLimit := ind.conf.OriginalsLimit()
 
@@ -168,7 +168,7 @@ func ImportWorker(jobs <-chan ImportJob) {
 				res := ind.MediaFile(f, indexOpt, originalName)
 
 				log.Infof("import: %s main %s file %s", res, f.FileType(), txt.Quote(f.RelName(ind.originalsPath())))
-				done[f.FileName()] = true
+				done[f.FileName()] = &struct{}{}
 
 				if res.Success() {
 					if err := entity.AddPhotoToAlbums(res.PhotoUID, opt.Albums); err != nil {
@@ -186,11 +186,11 @@ func ImportWorker(jobs <-chan ImportJob) {
 					continue
 				}
 
-				if done[f.FileName()] {
+				if done[f.FileName()] != nil {
 					continue
 				}
 
-				done[f.FileName()] = true
+				done[f.FileName()] = &struct{}{}
 
 				// Enforce file size limit for originals.
 				if sizeLimit > 0 && f.FileSize() > sizeLimit {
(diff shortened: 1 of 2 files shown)


lastzero:
Thank you very much! Any estimate how much memory this may save? Side effects?

Not 100% sure how Go handles this internally. From my understanding, the compiler should be able to optimize it automatically, when your code essentially does the same...

My estimate is that the current implementation uses a couple of MB max. Should be insignificant compared to image resampling & classification. When users complain about memory or performance, it's typically because of that (which in turn depends on file size and TensorFlow).

aaronjwood:
> Thank you very much! Any estimate how much memory this may save? Side effects?
> 
> Not 100% sure how Go handles this internally. From my understanding, the compiler should be able to optimize it automatically, when your code essentially does the same...
> 
> My estimate is that the current implementation uses a couple of MB max. Should be insignificant compared to image resampling & classification. When users complain about memory or performance, it's typically because of that (which in turn depends on file size and TensorFlow). 

Good question, I'm not sure if the compiler optimizes this automatically. Would have to write some benchmarks to see.

A few MB from boolean usage seems like alot considering booleans are pretty small. I'm thinking this is not on the stack either considering it's a few MB. That would mean  this will contribute to GC pressure down the line.

lastzero:
> A few MB from boolean usage seems like alot considering booleans are pretty small. I'm thinking this is not on the stack either considering it's a few MB. That would mean this will contribute to GC pressure down the line.

That was a maximum estimate. I may be wrong, but don't think there is significant GC pressure from this and the main bottlenecks are related to opening / converting / resizing sometimes very large files (also requires fast I/O in addition to heaps of memory), and TensorFlow of course. We prefer doing more analysis / testing before taking action. Thanks for bringing this up!

internal/photoprism/import_worker.go
```diff
@@ -152,7 +152,7 @@ func ImportWorker(jobs <-chan ImportJob) {
 				continue
 			}
 
-			done := make(map[string]bool)
+			done := make(map[string]*struct{})
```
julienschmidt, line 155: `bool` size is max. 1 byte (can be optimized to 1 bit), while `*struct{}` is a pointer and would take 8 bytes on a 64bit system. What you probably intended is `map[string]struct{}`, which indeed has a size of 0.

julienschmidt (changes requested):
The idea is right, however the current implementation would actually make it worse. See my comments on the code.

internal/photoprism/import_worker.go
```diff
@@ -186,11 +186,11 @@ func ImportWorker(jobs <-chan ImportJob) {
 					continue
 				}
 
-				if done[f.FileName()] {
+				if done[f.FileName()] != nil {
```
julienschmidt, line 189:
```suggestion
				if _, ok := done[f.FileName()]; ok {
```
Checks if the key is set (there is supposed to be no value)

lastzero:
> The idea is right, however the current implementation would actually make it worse. See my comments on the code.

Premature optimization is the...

internal/photoprism/import_worker.go
aaronjwood: Yup you're right. Intention was to use the type that takes no memory.

aaronjwood:
> > The idea is right, however the current implementation would actually make it worse. See my comments on the code.
> 
> Premature optimization is the... 

I did not measure the impact of this change. If you feel this is not needed or if the impact is miniscule let's close it out.

PR #1141, opened 2021-03-20
Fix locking

aaronjwood:
Just a few small things I noticed :)


lastzero:
Excellent, good catch! Thank you very much 👍

PR #1142, opened 2021-03-20
Fix the arm32 docker build

kvalev:
Weirdly wget fails with an invalid certificate error in a virtualized arm32 environment (qemu), but not on the actual hardware. Replacing it with curl fixes the issue.


lastzero:
Thanks for improving this! 👍

PR #1144, opened 2021-03-20
Remove Niagara Falls from country mapping

kvalev:
The waterfalls are on a border, so they cannot be uniquely assigned to Canada (or USA).


graciousgrey:
Thank your for all your pull requests! We will review them as soon as possible - hopefully in April :)

lastzero:
From my point of view, CA is better than nothing as it's just an estimate anyway... from what I remember there are more tourists on that side. I'll merge this anyway since it seems to cause issues.

PR #1146, opened 2021-03-21
Use SHA-512/256 instead of SHA-1

aaronjwood:
A few reasons for this:

1. SHA-1 is completely broken and has been for a while https://shattered.io/
2. SHA-512 based functions are normally much faster than SHA-256 due to the use of 64-bit words (and I think it's safe to say that most people are using 64-bit chips these days)
3. SHA-512/256 is resistant to collision and length-extension problems https://en.wikipedia.org/wiki/SHA-2#Comparison_of_SHA_functions


internal/hub/config.go
```diff
@@ -67,7 +67,7 @@ func (c *Config) Sanitize() {
 	c.Key = strings.ToLower(c.Key)
 
 	if c.Secret != "" {
-		if c.Key != fmt.Sprintf("%x", sha1.Sum([]byte(c.Secret))) {
+		if c.Key != fmt.Sprintf("%x", sha512.New512_256().Sum([]byte(c.Secret))) {
```
aaronjwood, line 70: I think long term this area should move away from using hashing completely. Ideally you'd want to use something like PBKDF2 or, preferably, bcrypt/scrypt. A single round of hashing without a salt is too weak.

internal/hub/config.go
```diff
@@ -93,7 +93,10 @@ func (c *Config) DecodeSession() (Session, error) {
 	}
 
 	hash := sha256.New()
```
aaronjwood, line 95: Just curious, why is a SHA-256 hash getting encrypted with AES here? Why not just AES encrypt the raw data directly?

graciousgrey:
Thank you for all your PR's :) We will review them as soon as possible.

internal/hub/config.go
alerque: If secrets are being checked here you can't just swap out the hash mechanism with a different one without providing a migration path for previously stored secrets

internal/hub/config.go
alerque: There shouldn't be any point in hashing *and* encrypting anything unless it happens to be a piece of a larger data set. For example you might encrypt a user profile database that hase various user info plus have the password hash in that data set, hence encrypting the hash. But the point isn't to encrypt the hash (which would be pointless by itself) but to keep the data together as a set.

I don't know what's going on here, but it ought to be apparent what the use cases is, this shouldn't be an open question while your swapping out hash mec

internal/hub/config.go
aaronjwood: Hoping one of the author's can confirm my understanding of this area. If it is the case then I'd need some direction on providing a migration path as I'm not familiar with the majority of the codebase.

lastzero:
Thanks for your PR! Much appreciated.

According to this and other benchmarks, SHA-1 is significantly faster than SHA-512/256:

https://www.cryptopp.com/benchmarks.html

While SHA-1 can not be considered cryptographically secure (anymore), we use it for checksums and are aware it's unsuitable for cryptographic applications of any sort. Common file system implementation are even less secure and just use CRC32, which is even simpler to "hack".

Using a longer checksum as key may reduce database performance as well as less rows fit on a page. In addition, databases may use a shorter hash for indexing, thus the number of effective bits is less than it would look like.

The Git manual explains the differences pretty well: https://git-scm.com/docs/hash-function-transition/

internal/hub/config.go
lastzero: It's a proof the client knows the secret without transmitting it in plain text, as it is also used for signing requests later. True, using a salt would be even more secure, but it's good enough for distributing maps keys. That's also why SHA-256 is used here and not SHA-1 as for file checksums.

internal/hub/config.go
lastzero: We don't need higher security at the moment as the main use is distributing maps keys. We pay for them, so we wan't to avoid simple abuse (like script kids getting the keys in plain text using a perl script).

aaronjwood:
> Thanks for your PR! Much appreciated.
> 
> According to this and other benchmarks, SHA-1 is significantly faster than SHA-512/256:
> 
> https://www.cryptopp.com/benchmarks.html
> 
> While SHA-1 can not be considered cryptographically secure (anymore), we use it for checksums and are aware it's unsuitable for cryptographic applications of any sort. Common file system implementation are even less secure and just use CRC32, which is even simpler to "hack".
> 
> Using a longer checksum as key may reduce database performance as well as less rows fit on a page. In addition, databases may use a shorter hash for indexing, thus the number of effective bits is less than it would look like.
> 
> The Git manual explains the differences pretty well: https://git-scm.com/docs/hash-function-transition/ 

You're right about sha2 being slower overall than sha1, don't think you can do much there :)

Since you mentioned you're using this for real checksums why not use CRC32 if speed and/or hash size is what you're going for? It seems like the perfect candidate.

lastzero:
> Since you mentioned you're using this for real checksums why not use CRC32 if speed and/or hash size is what you're going for? It seems like the perfect candidate.

There's two main reasons:

1. It's too short for comparing millions of files, so a collision is much more likely compared to SHA-1. That makes it unsuitable for finding / avoiding duplicates and / or would reduce performance and / or increase complexity due to additional collision handling logic. We use CRC32 where suitable.
2. SHA-1 is well optimized and standardized across programming languages and operating systems, whereas there are multiple CRC32 polynomials: https://golang.org/pkg/hash/crc32/#pkg-constants

aaronjwood:
> > Since you mentioned you're using this for real checksums why not use CRC32 if speed and/or hash size is what you're going for? It seems like the perfect candidate.
> 
> There's two main reasons:
> 
> 1. It's too short for comparing millions of files, so a collision is much more likely compared to SHA-1. That makes it unsuitable for finding / avoiding duplicates and / or would reduce performance and / or increase complexity due to additional collision handling logic. We use CRC32 where suitable.
> 2. SHA-1 is well optimized and standardized across programming languages and operating systems, whereas there are multiple CRC32 polynomials: https://golang.org/pkg/hash/crc32/#pkg-constants
> 

Got it. I misunderstood, I thought you were saying earlier that sha1 was only being used for detecting errors and not for comparing contents.

PR #1150, opened 2021-03-22
Try to infer media date from filename

### LINKED ISSUE #1102: Occasionally Wonky Date Extraction
I've been doing some tests on various photos and I'm found that PP gets confused with date extraction sometimes. 

For example, this file seems to have no embedded "Date Taken" info:
```
root@da67aa80bf7b:/photoprism# exiftool "originals/2019/11 November/IMG-20191120-WA0001.jpg"
ExifTool Version Number         : 12.05
File Name                       : IMG-20191120-WA0001.jpg
Directory                       : originals/2019/11 November
File Size                       : 196 kB
File Modification Date/Time     : 2021:02:20 15:33:03+00:00
File Access Date/Time           : 2021:02:27 13:37:34+00:00
File Inode Change Date/Time     : 2021:02:21 20:06:19+00:00
File Permissions                : rwxr-x---
File Type                       : JPEG
File Type Extension             : jpg
MIME Type                       : image/jpeg
JFIF Version                    : 1.01
Resolution Unit                 : None
X Resolution                    : 1
Y Resolution                    : 1
Image Width                     : 1600
Image Height                    : 800
Encoding Process                : Progressive DCT, Huffman coding
Bits Per Sample                 : 8
Color Components                : 3
Y Cb Cr Sub Sampling            : YCbCr4:2:0 (2 2)
Image Size                      : 1600x800
Megapixels                      : 1.3
```

When importing it, PP sets the right month and year, but sets that date for it (and all the other images from whatsapp) as the 1st:

```
UID | FQOVAKH1725H5HXJ
Hash | 5b14f24c9c4ad6be334a693db167f677ff00a10d
Name | 2019/11 November/IMG-20191120-WA0001.jpg
Size | 1600 × 800, 0.2 MB
Type | JPG
Codec | JPEG
Primary | Yes
Aspect Ratio | 2
Main Color | Grey
Chroma | 8 / 100
Added | Feb 21, 2021, 7:09 AM in 1,832 ms
```

I wondered what source PP is using for date information when the exif data doesn't provide it? Is it running a regex on the filename? Or Possibly on the folder path?

I know that finding dates for photos that lack proper exif data is a can of worms, as they can come in all types and formats. Given that Whatsapp will be a major source of photos for some libraries, and given it seems to strip that info though, I wondered if we can extract it with a regex looking for a few standard date formats in the file name?

kvalev:
Add support for media files, where the taken date is part of the filename, for example `IMG-20191120-WA0001.jpg`.

closes #1102

### DIFF
--- a/pkg/txt/time.go
+++ b/pkg/txt/time.go
@@ -8,6 +8,7 @@ import (
 
 var DateRegexp = regexp.MustCompile("\\D\\d{4}[\\-_]\\d{2}[\\-_]\\d{2,}")
 var DatePathRegexp = regexp.MustCompile("\\D\\d{4}/\\d{1,2}/?\\d*")
+var DateFilenameRegexp = regexp.MustCompile("\\d{4}\\d{2}\\d{2}")
 var DateTimeRegexp = regexp.MustCompile("\\D\\d{4}[\\-_]\\d{2}[\\-_]\\d{2}.{1,4}\\d{2}\\D\\d{2}\\D\\d{2,}")
 var DateIntRegexp = regexp.MustCompile("\\d{1,4}")
 var YearRegexp = regexp.MustCompile("\\d{4,5}")
@@ -17,6 +18,10 @@ var IsDateTimeRegexp = regexp.MustCompile("\\d{4}[\\-_]?\\d{2}[\\-_]?\\d{2}.{1,4
 var (
 	YearMin = 1990
 	YearMax = time.Now().Year() + 3
+)
+
+var (
+	DefaultTime = time.Time{}
 )
 
 const (
@@ -36,12 +41,12 @@ const (
 func Time(s string) (result time.Time) {
 	defer func() {
 		if r := recover(); r != nil {
-			result = time.Time{}
+			result = DefaultTime
 		}
 	}()
 
 	if len(s) < 6 {
-		return time.Time{}
+		return DefaultTime
 	}
 
 	if !strings.HasPrefix(s, "/") {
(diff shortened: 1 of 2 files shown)


lastzero:
Thanks for your PR & sorry for the wait! We've intentionally removed **\\d{4}\\d{2}\\d{2}** matching prior to our stable release for two reasons:

- Files imported with PhotoPrism match to this, so it would look as if an actual date exists - even when the import file name was generated from the file mod time. A solution could be to use other common names for imported files, but it would break compatibility and may cause other issues. So not a quick win.
- All filenames containing large integers would match to this, even when they're completely random or serial numbers.

kvalev:
> Thanks for your PR & sorry for the wait! We've intentionally removed **\d{4}\d{2}\d{2}** matching prior to our stable release for two reasons:
> 
>     * Files imported with PhotoPrism match to this, so it would look as if an actual date exists - even when the import file name was generated from the file mod time. A solution could be to use other common names for imported files, but it would break compatibility and may cause other issues. So not a quick win.
> 
>     * All filenames containing large integers would match to this, even when they're completely random or serial numbers.

I cant say much about the first point, as I am not using the import function, but the second one is not quite correct, as there is a plausibility check for the year, month and day components. A collision can still happen, but the probability is minuscule.

lastzero:
It depends, some users might have a ton of serial numbers that much - even though we both don't. It's hard to say for sure.

Even if that's not a problem, the imported filenames would match and will increase the priority of dates that originally were just mod dates as they now look as if they are coming from the file name.

PR #1161, opened 2021-03-30
Metadata: Add support for XMP sidecar CreateDate and Keywords

graciousgrey:
TO DO:
- Currently the SetKeyword function in details.go overwrites existing keywords (e.g. colors) with the ones coming from xmp
- When running optimize, single keywords in the photo details are sanitized (e.g. Penguin becomes penguin) but this is not working in case there are multiple keywords in the details like "Krokus, Blume"
=> Once those are fixed the out-commented test in index_related_test should pass
- Add "Import from Apple" docs
- Add new fields to metadata docs

QUESTION:
- Could we also read the location data from the example xmps? If yes, create follow up ticket
- When is the data coming from xmp, written to the yml sidecars?


internal/meta/xmp.go
```diff
@@ -57,5 +57,14 @@ func (data *Data) XMP(fileName string) (err error) {
 		data.LensModel = doc.LensModel()
 	}
 
+	if takenAt := doc.DateCreated(); !takenAt.IsZero() {
+		data.TakenAt = takenAt
```
graciousgrey, line 61: How to set the local time here? With this implementation local time and UTC are the same?

lastzero:
Looks good to me now :rocket:

lastzero (approved):

PR #1192, opened 2021-04-14
Docker compose has been rolled into the main CLI, so the hyphen is deprecated.

ryanpcmcquen:
```% docker-compose up -d
Docker Compose is now in the Docker CLI, try `docker compose up`
```

This should probably be propagated to all docker-compose files, but I can only test on Mac, so I only edited this one.

### DIFF
--- a/docker/examples/macos/docker-compose.yml
+++ b/docker/examples/macos/docker-compose.yml
@@ -11,19 +11,19 @@ version: '3.5'
 # ------------------------------------------------------------------
 # DOCKER COMPOSE COMMAND REFERENCE
 # ------------------------------------------------------------------
-# Start    | docker-compose up -d
-# Stop     | docker-compose stop
-# Update   | docker-compose pull
-# Logs     | docker-compose logs --tail=25 -f
-# Terminal | docker-compose exec photoprism bash
-# Help     | docker-compose exec photoprism photoprism help
-# Config   | docker-compose exec photoprism photoprism config
-# Reset    | docker-compose exec photoprism photoprism reset
-# Backup   | docker-compose exec photoprism photoprism backup -a -i
-# Restore  | docker-compose exec photoprism photoprism restore -a -i
-# Index    | docker-compose exec photoprism photoprism index
-# Reindex  | docker-compose exec photoprism photoprism index -a
-# Import   | docker-compose exec photoprism photoprism import
+# Start    | docker compose up -d
+# Stop     | docker compose stop
+# Update   | docker compose pull
+# Logs     | docker compose logs --tail=25 -f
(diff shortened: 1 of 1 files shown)


StephenBrown2:
Can you specify that this is only for MacOS and Windows? https://github.com/docker/compose-cli/issues/1149#issuecomment-762278484, and for what version of the docker cli this is for? Looks like the message was added to docker-compose 1.28.6, but I'm still running `docker-compose version 1.28.5, build c4eb3a1f` on my Mac. I can't find mention of compose in the changelog for https://github.com/docker/cli/releases, though the docs mention it is only a tech preview: https://docs.docker.com/compose/cli-command/ and doesn't have full compatibility with `docker-compose`: https://docs.docker.com/compose/cli-command-compatibility/
> ❗ Important
> 
> The compose command in the Docker CLI is currently available as a Tech Preview. We recommend that you do not use this in production environments.

alexislefebvre:
I think it's too early, I use Ubuntu 20.04 and the [official release channel of Docker](https://docs.docker.com/engine/install/ubuntu/) and `docker compose` doesn't work yet:

```shell
$ docker compose
docker: 'compose' is not a docker command.
See 'docker --help'
$ docker --version
Docker version 20.10.6, build 370c289
```

Updating these commands would complicate the usage because users would have to rewrite every `docker compose` command to `docker-compose`.

ryanpcmcquen:
Hahahahaha:

https://github.com/lando/lando/issues/2962

lastzero:
Let's wait a tiny bit until everyone has the new release. If the info message bothers you, I'm sure there is a way to disable it in the meantime.

alexislefebvre:
Users should be able to use an alias in their shell from `docker-compose` to `docker compose` in order to use any version without seeing the message about the deprecation.

alexislefebvre:
I stumbled upon this PR again, and I saw that it targets only the documentation for MacOS. If the last release of Docker on MacOSdeprecate `docker-compose`, then this PR can be updated.

StephenBrown2:
Docker Compose 2.0 is still in beta, and breaks at least one `docker-compose.yml` file (the main one used for local dev at my work), so I would recommend not switching any documentation, and perhaps adding `docker-compose disable-v2` to the docs instead until it's verified that 2.0 works with this compose file.

lastzero:
@alexislefebvre  According to my tests, the "docker compose ..." subcommands are only optional if you explicitly install the "docker-compose-plugin". It may be that some distributions and Docker Desktop do this, but in general it does not seem to be the default.

lastzero:
I have changed this everywhere I found it. You are welcome to help me check if all changes are correct!

PR #1193, opened 2021-04-15
add usage text to describe argument to backup command

### LINKED ISSUE #1190: back up to the same file location every time
I have a cron job configured to run `docker-compose exec -T photoprism photoprism backup -a -i -f` every so often, but I'm noticing that the backups are stored in the `PHOTOPRISM_BACKUP_PATH` in a file named with the date. I use [restic](https://github.com/restic/restic) for remote backups on my system, and restic does periodic snapshots of files. It would make sense for `photoprism backup -f` to write to the same file every time, so that restic only stores the latest Photoprism backup in each snapshot and I don't have to periodically delete old Photoprism backups.

kylrth:
closes #1190

### DIFF
--- a/internal/commands/backup.go
+++ b/internal/commands/backup.go
@@ -23,8 +23,11 @@ import (
 
 // BackupCommand configures the backup cli command.
 var BackupCommand = cli.Command{
-	Name:   "backup",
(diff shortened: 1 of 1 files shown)


lastzero:
Thank you very much!

PR #1201, opened 2021-04-17
Add -overwrite_original option to exiftool

acranox:
Fixes #1200

Add -overwrite_original option to exiftool to avoid creating a copy of
the image file.
Adds -P option to preserve file modification time


lastzero:
Thank you very much!

PR #1202, opened 2021-04-17
Collapse the ENV instructions into fewer steps

acranox:
Docker allows setting multiple key/value pairs in one ENV instruction
https://docs.docker.com/engine/reference/builder/#env

I use ZFS as the storage driver, and building the docker image causes an intermediate ZFS clone for each step in the docker file, which takes a few seconds.  So each ENV instructions was adding a few seconds, which seemed like a lot for an environment variable.  This change reduces the number of Docker steps from 71 down to 31.  The overall performance improvement is trivial at 1-2 minutes, but I wanted to offer the Pull Request in case it seemed worthwhile.


alexislefebvre:
Nice! That would clean up these numerous layers visible on Docker Hub: https://hub.docker.com/layers/photoprism/photoprism/20210222/images/sha256-b3d95d91e99c8e672ad3d971c088dee068afe2ed14365ae266bc72af6f0d103f?context=explore

lastzero:
Excellent, wasn't aware of that... thank you!

PR #1222, opened 2021-04-23
Frontend: Copy to clipboard without HTTPS

haimkastner:
Allow copy content to the clipboard even if the "window.navigator.clipboard" doesn't support

### DIFF
--- a/frontend/src/common/util.js
+++ b/frontend/src/common/util.js
@@ -179,7 +179,7 @@ export default class Util {
       document.body.removeChild(clipboardElement);
 
       // Validate operation succeed
-      if(!succeed){
+      if (!succeed) {
         throw new Error('Failed copying to clipboard');
       }
     } else {
--- a/frontend/src/common/util.js
+++ b/frontend/src/common/util.js
@@ -147,9 +147,9 @@ export default class Util {
   }
 
   static async copyToMachineClipboard(text) {
-    if(window.navigator.clipboard) {
+    if (window.navigator.clipboard) {
       await window.navigator.clipboard.writeText(text);
-    } else if(document.execCommand) { 
+    } else if (document.execCommand) { 
       // Clipboard is available only in HTTPS pages. see https://web.dev/async-clipboard/
       // So if the the official 'clipboard' doesn't supported and the 'document.execCommand' is supported.
       // copy by a work-around by creating a textarea in the DOM and execute copy command from him.
@@ -173,7 +173,7 @@ export default class Util {
       clipboardElement.select();
       
       // Copy the selected textarea content
-      const succeed = document.execCommand('copy')
+      const succeed = document.execCommand('copy');
 
       // Remove the textarea from DOM
       document.body.removeChild(clipboardElement);
--- a/frontend/src/common/util.js
+++ b/frontend/src/common/util.js
@@ -150,7 +150,7 @@ export default class Util {
     if(window.navigator.clipboard) {
       await window.navigator.clipboard.writeText(text);
     } else { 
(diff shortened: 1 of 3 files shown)


frontend/src/common/util.js
```diff
@@ -145,4 +145,43 @@ export default class Util {
     console.log(`${label}: ${now.getTime() - start.getTime()}ms`);
     start = now;
   }
+
+  static async copyToMachineClipboard(text) {
+    if(window.navigator.clipboard) {
+      await window.navigator.clipboard.writeText(text);
+    } else { 
+      // Clapboard is available only in HTTPS pages. see https://web.dev/async-clipboard/
```
alexislefebvre, line 153:
```suggestion
      // Clipboard is available only in HTTPS pages. see https://web.dev/async-clipboard/
```

Fylax:
Out of curiosity: is there any particular reason to make a new `Util` method instead of going for a more classical `prototype` that prevents refactoring?

haimkastner:
Hi @Fylax,

I didn't fully understand what is your suggestion.

Do you mean to implement a `writeText` for `window.navigator.clipboard` if it does not exist? 

If so, I don't think it's a good idea to change the browser objects, and also where do you think we should implement it? in the `app.js` file?

Fylax:
Hi @haimkastner,
Indeed it is what I am asking.

I'm not stating is the correct solution, just wondering if it can come handy.

Usually I create a `Prototypes` module that I load in application initialization. Common examples are `string.replaceAll`.

This mechanism allows me to have a uniform API without having to put lot of logic to discriminate between different browser versions and/or HTTP vs HTTPS.

Why would you be agains implementing `writeText` inside `windows.navigator.clipboard`?

haimkastner:
Hi @Fylax,

It's only my thought, but I think that it's cool to extend an object prototype with some extra methods like your example, extending `replace-all` for the string prototype.

But to **override** existing object behavior that you don't own and you don't have full control over it such as the browser `window` object can make issues that you can't expect to, and In addition, in a case of bug/issue in the logic of the extension, no one can guess it's a problem in the prototype extensions that implemented somewhere in the app initialization since it's an official browser API.

Fylax:
Hi @haimkastner,

thank you for your clarification, which makes completely sense now.

For any mantainer stepping here, you can assume my comments as a review: so far, the code is good.

lastzero:
Wouldn't browsers likewise prevent copying from a textarea? Surprises me...

haimkastner:
Yes...
I tested it on Chrome version 90.0.4430.85 (Official Build) (64-bit) Windows 10.

Fylax:
@lastzero the `document.execCommand` API does not have any limit among HTTP/HTTPS, so it is expected to work out of the box.

Thinking of it, the only suggestion I might suggest to @haimkastner is adding a check to that API existence to avoid calling it and just throw an error if it doesn't.

Something like
```js
if(window.navigator.clipboard) {
  await window.navigator.clipboard.writeText(text);
} else if (document.execCommand) {
  // your code here
} else {
  throw new Error('Failed copying to clipboard')
}
```

Any thoughts on that?

haimkastner:
@Fylax that sounds great to me, I will implement it soon.

lastzero:
Thanks! We'll wait with merging :+1:

friendyou1:
I'm the Electronic person to reserve in a sad moment.

friendyou1 (commented):
Not worthy @Fylax

Fylax (approved):

graciousgrey:
Is this one ready to be merged ? :)

haimkastner:
Hi @graciousgrey , from my side, yes.

graciousgrey:
Thank you :)

PR #1243, opened 2021-04-28
Create codeql-analysis.yml

jfcg:
Add Github CodeQL security checks. In my fork under `/security/code-scanning` I see:

photoprism

Some of them, including zip slip, seem to be real security flaws.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,31 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ develop, release, master ]
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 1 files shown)


lastzero:
Thanks! We'll take a look. ZIP extraction is probably the test data from our own server secured by SSL. So yes, if you take over our server and a developer puls the test data, it may impact the files inside the Docker container in a possibly unwanted way. Could also be the TensorFlow model. Going to impact what's inside the Docker image in the same way. If an attacker takes over our complete infrastructure, I guess it's game over anyway...

lastzero:
Merged this... where / how can we see the results?

jfcg:
/security/code-scanning on your repo.

Cheers

lastzero:
Fixed two potential issues with str to int conversion and unzip file names, although likely not critical. Unzip is only used in tests so far. Paths can be configured by the user, so an admin could attack itself by changing paths different folders. Same for WebDAV URLs, which may indeed point to internal intranet services. Works as designed from my point of view.

lastzero:
Thank you very much! :partying_face:

PR #1257, opened 2021-05-03
Give the photoprism service the standard name photoprism 

zierbeek:
Add container_name to the docker-compose file, making it easier to setup other services who rely on the container name and making it easier for things like docker restart, docker log ,...

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,6 +17,7 @@ services:
   photoprism:
     build: .
     image: photoprism/photoprism:develop
+    container_name: photoprism
     security_opt:
       - seccomp:unconfined
       - apparmor:unconfined


lastzero:
Wound't this cause issues when running multiple instances on the same server?

zierbeek:
Hmm indeed, is there a case when people run multiple instances?

________________________________
Van: Michael Mayer ***@***.***>
Verzonden: dinsdag 4 mei 2021 11:23
Aan: photoprism/photoprism ***@***.***>
CC: zierbeek ***@***.***>; Author ***@***.***>
Onderwerp: Re: [photoprism/photoprism] Give the photoprism service the standard name photoprism (#1257)


Wound't this cause issues when running multiple instances on the same server?

lastzero:
> Hmm indeed, is there a case when people run multiple instances?

Yes, some users do this.

You can refer to the container via `docker-compose [command] photoprism` instead, container name is mainly useful when using the docker command directly from my experience.

lastzero:
You think we should still add it in this case? I'm afraid that it causes errors for users with more instances... going to ask for support and/or we need to add more docs...

zierbeek:
Yes perfect, for this case it could be good for using SWAG. Since it took me a while to fix it for swag. I Will add some info in the docs. Now it used docker rename.
________________________________
Van: Michael Mayer ***@***.***>
Verzonden: dinsdag 4 mei 2021 19:14
Aan: photoprism/photoprism ***@***.***>
CC: zierbeek ***@***.***>; Author ***@***.***>
Onderwerp: Re: [photoprism/photoprism] Give the photoprism service the standard name photoprism (#1257)


You think we should still add it in this case? I'm afraid that it causes errors for users with more instances... going to ask for support and/or we need to add more docs...

lastzero:
Alright, I'll close this then... thank you!

PR #1328, opened 2021-05-25
Docker multi-arch images

moximoti:
added multi-arch builds for development and photoprism docker images. didn't include arm32 as the build fails on my m1 mac. someone could try if the build works and enable it later. Dockerfile should be working for arm32.


docker/development/multiarch/Dockerfile
```diff
@@ -0,0 +1,163 @@
...
+RUN npm install --unsafe-perm=true --allow-root -g npm testcafe chromedriver && \
+    npm config set cache ~/.cache/npm
+
+# Install Go
+COPY ./scripts/go-bin-checksum.sh /tmp/
+ENV GOLANG_VERSION 1.16.4
+RUN set -eux; \
+    url=$(echo https://golang.org/dl/go${GOLANG_VERSION}.${TARGETPLATFORM}.tar.gz | sed -e 's@/v7@v6l@g' -e 's@linux/@linux-@g'); \
+    wget -O go.tgz "$url"; \
+    ./tmp/go-bin-checksum.sh ${TARGETARCH} | sha256sum -c -; \
```
lastzero, line 114: Instead of just getting the go archive checksum from `scripts/go-bin-checksum.sh` you could directly install Go with a `scripts/go-install.sh` script?

moximoti: integrated your proposal. additionally switched off chromedriver for arm32 builds. the multi-arch dockerfile should now work for this platform too.

lastzero:
Thanks! I'll merge and and test it with Drone CI. Looks like `scripts/go-bin-checksum.sh` can be removed?

kvalev:
Is the arm32 build working? I dont see an image on DockerHub.

lastzero:
So far, only the development image is available for arm v7. Compiling Go source for arm32 takes ages, so we don't build preview images, only stable releases. Unless someone wants to sponsor a 16 core build server at DigitalOcean that is.

kvalev:
There is already a precompiled golang version, which was being used in the Dockerfile I contributed, so there is no need to compile it yourself.

kvalev:
Apropos, there is no need for a custom DO build server, as you can do everything for free with GitHub actions (but slightly slower that your current build), as I am doing in my fork: https://github.com/kvalev/photoprism/blob/preview/.github/workflows/release.yaml

lastzero:
Didn't mean compiling Go itself, but our binaries. I expect it to be super slow, it's already slow with 4 new Xeon cores. You probably don't cross compile via GitHub Actions?

lastzero:
Maybe GitHub Actions can compile arm v7 natively? Anyhow, compiling ARM64 is more than 10x faster than ARM on our Intel CPUs. We're currently working on Face Detection / Recognition and also need a break from work, so no time to replace our existing build environment right now.

PR #1349, opened 2021-06-03
Fix unhandled error in Config.DecodeSession

garrettgu10:
Looking at the [source implementation of sha256.Write](https://golang.org/src/crypto/sha256/sha256.go), it's currently not possible for an error to be thrown by the function. However, I still think it's good form to handle this err here in case this changes in the future.


lastzero:
Thanks! Looks like it should do no harm :)

PR #1352, opened 2021-06-04
Update package-lock.json

benmccann:
When I run `npm install` it removes several packages that look to be unnecessary. I'm using npm 7.11.1. Perhaps this file was generated with an older version? Or the `package.json` was updated and the `package-lock.json` wasn't regenerated?

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1687,39 +1687,6 @@
         "node": ">=0.10.0"
       }
     },
-    "node_modules/@vue/reactivity": {
-      "version": "3.0.11",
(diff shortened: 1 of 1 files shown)


lastzero:
My NPM version is typically up-to-date as it always shows an info when there is a newer version.

The specific contents of `package-lock.json` may depend on weather you run `npm audit fix` after `npm update`?

lastzero:
Note that `npm audit fix` can't update all packages right now as there is a (server side) security issue with a webpack css plugin. The new version isn't compatible with our version of webpack. The new version of webpack isn't compatible with couple of other dependencies. One of the few items on our todo.

benmccann:
It looks like this has since been fixed by another commit

Issue #1354, opened 2021-06-05
UX: Configure default page for PWA


AliMickey:
Would be great to have an option to load a specific side bar menu. Example: load the album page when the app is launched.

AliMickey:
Clarification - this is for the mobile layout.

graciousgrey:
Do you mean the mobile app or photoprism as PWA?

AliMickey:
Photoprism as a PWA. Is there a mobile app? I thought PWA was the only and recommended way. 

Outside of mounting the WebDAV URL.

graciousgrey:
Yes, we recommend using PhotoPrism as PWA and PhotoSync for automatic sync. There is a community maintained app as well, but it is still in beta (https://github.com/photoprism/photoprism-mobile).

AliMickey:
Ah, that looks nice. I only need interfaces for my family to access the photo library. 

Any idea when the required API changes will be done? Or after the face recognition update?

Also I don't have experience in this framework, however i will be trying to implement this issue into the app. 
- Is there a way to sort folders? It seems to be random now, i would like it to be alphabetical if possible? Or needs implementing?

AliMickey:
Is there a set of instructions on how to build this from source? I tried running 'go run .' from cmd/photoprism but running into issues.

AliMickey:
Submitted a PR: https://github.com/photoprism/photoprism/pull/1361

PR #1361, opened 2021-06-07
Implement a default home page selector

AliMickey:
https://github.com/photoprism/photoprism/issues/1354

I have never used this framework before, apologies if there are any obvious mistakes. 

Its a fairly simple PR, please do consider merging as this would benefit my use case at home. 

Thanks


graciousgrey:
Thank you for your pull request, we will have a look after our vacation :)

frontend/src/options/options.js
```diff
@@ -236,6 +236,70 @@ export const Themes = () => [
...
+    value: "favorites",
+    disabled: false,
+  },
+  {
+    text: $gettext("Moments"),
+    value: "moments",
+    disabled: false,
+  },
+  {
+    text: $gettext("Calender"),
```
graciousgrey, line 267: Needs to be Calendar

frontend/src/options/options.js
```diff
@@ -236,6 +236,70 @@ export const Themes = () => [
...
+    value: "calender",
```
graciousgrey, line 268: calendar ;)

graciousgrey:
Setting the default page works :)
What we need to ensure is, that a user cannot set a default page that is disabled.
E.g. if a user disables private there should be no private option in the default page selector

AliMickey:
Sounds good,

I am unable to build for some reason, however I think that should be very simple. 

Here are screenshots of what I think will do the trick? I'm not sure if the settings features uri is correct. Please correct if wrong. 

Also, does the import have to be:

`import { config } from "../session";`

or

`import { config } from "session";`

image

AliMickey:
Do you need me to do this on my side so you can review it? I sponsored the project so I could get this through, as well as the native app API fix that's required.

lastzero:
Thank you very much! We can finish this if you're not in a hurry. Currently working on facial recognition. We have a small team, and doing our very best to get everything done as soon as possible.

AliMickey:
Thanks for an update, just wanted to make sure the issues were not forgotten.

platima:
Aaaah @AliMickey what are you doing?

AliMickey:
Ha! I was cleaning up my repos. and assumed this was a remnant one. 

Demonstrates the state of this project unfortunately, moving over to Immich at the moment. 

To the devs who might see this, I really do understand the immense workload and effort required by the team (which is still only the original couple despite recurring income, forgive me if this has changed) for a project of this magnitude, but unfortunately progress has slowed down. 

I wish you and the team the best of luck with this, I appreciate the hard work and for giving the community an alternative to Google Photos etc.

platima:
Hey @AliMickey looks like someone didn't like your comment, even if it did have a positive tone. Each to their own though and I've got the email notification of the comment if you ever need it.

Regarding your move though, I've gone the opposite way and moved from Immich to Photoprism purely for the tags/labels and the folder view, as Immich seems to have lost its tagging capability at some point, and has no folder view.

Just my personal preference really, but I have offered Theresa and Michael some funds to cover this change request!

lastzero:
@platima Honestly, I'm not sure why this slipped through - probably, yes, because of all the other work we have going on. Based on your interest, I had planned to get it implemented this week.

Ultimately, these discussions cause a lot of noise because we get an email notification every time (just like you), which distracts us from our work and slows things down for everyone (exactly what has been criticized with a "positive tone", yet it's not helpful in any way). I'm not making this up to annoy you, it's written right here in our README.md:

- https://github.com/photoprism/photoprism?tab=readme-ov-file#github-issues-%EF%B8%8F

**So I'm going to move ahead and close this discussion. If you have questions or want to provide personal feedback, you can do so via email. Thank you both for your support!**

Issue #945, opened 2021-01-22
Server-Timing headers


benmccann:
One thing I think that would be really cool is to have a `Server-Timing` header. This is a header that Chrome, Firefox, and Edge all support in order to show more information in the developer tools by debugging.

I was thinking that there'd be two pieces of information that would be generally helpful. One is how long it took to talk to the database and the other is how long it took to process the request in total. That would look like:

`Server-Timing: db;dur=201.098919, total;dur=301.271915;`

benmccann:
This could be an interesting Gin middleware to use to implement this: https://pkg.go.dev/github.com/p768lwy3/gin-server-timing
Or could use the underlying library it's based on if the Gin middleware turns out to have any issues with it: https://github.com/mitchellh/go-server-timing/

benmccann:
I've got this mostly working here: https://github.com/benmccann/photoprism/tree/server-timing

I've calculated the header. However, Gin isn't doing anything with it when I try to set it because it's trying to add a header after `Next` has been called, which Gin apparently doesn't allow for or at least not easily. Maybe someone who's more experienced with Gin could figure out this last bit

Issue #1351, opened 2021-06-04
Demo: Server is slow


benmccann:
It's taking me 25s just to load the JS on https://demo.photoprism.org/browse. Certainly some of that is that the file is a bit large at 759 kB and can be improved by things like https://github.com/photoprism/photoprism/pull/841, but even still it seems quite surprising that it'd take that long since it's just a static file and there's nothing being computed.

I can't really imagine what would make it take this long except maybe the demo server is overloaded? I'm curious what hosting service is used. I wonder if there's anything we can do to expose some metrics (possibly publicly) to make it clear if CPU, disk, or network is experiencing high utilization. Maybe something like [Prometheus Node Exporter](https://github.com/prometheus/node_exporter)?

![Screenshot from 2021-06-04 11-34-57](https://user-images.githubusercontent.com/322311/120852516-2af89100-c52f-11eb-9a1b-8367ab0561af.png)

lastzero:
Maybe there's an issue with the server / provider. Our upload to Docker Hub was also super slow, had to get a new server for this. Going to check this after our vacation. In general, PhotoPrism seems fast enough for most users and it's really super fast if the server is within the same country, even on a mobile network. I'm using it every day. No performance issues whatsoever.

lastzero:
System load wasn't high at all. Rebooted the server. Better now?

lastzero:
Slow routing to the US in particular seems possible too.

benmccann:
It took me 16s this time, so it doesn't look like the reboot helped much. I agree it's not likely to be an issue with the PhotoPrism code.

Slow network does seem like a leading contender. I'm not quite sure how to confirm that diagnosis. `ping` doesn't take terribly long. I think the default TCP window size on Ubuntu is large enough that there shouldn't be many round trips needed.

```
PING demo.photoprism.org (45.132.246.223) 56(84) bytes of data.
64 bytes from v2202007125883123470.happysrv.de (45.132.246.223): icmp_seq=1 ttl=46 time=163 ms
64 bytes from v2202007125883123470.happysrv.de (45.132.246.223): icmp_seq=2 ttl=46 time=165 ms
64 bytes from v2202007125883123470.happysrv.de (45.132.246.223): icmp_seq=3 ttl=46 time=227 ms
64 bytes from v2202007125883123470.happysrv.de (45.132.246.223): icmp_seq=4 ttl=46 time=165 ms
[...]

PR #1365, opened 2021-06-08
Add a couple Server-Timing headers

benmccann:
Closes https://github.com/photoprism/photoprism/issues/945

I added for two API calls right now that are hit on the initial page load. I'm not imagining we'd add this to all API routes, but over time can add to ones that we are particularly interested in

This may help provide some more information for https://github.com/photoprism/photoprism/issues/1351


graciousgrey:
Thank you :) We will have a look after our vacation!

benmccann:
@lastzero do you have any thoughts about this one?

lastzero:
We're currently trying to grow our team to get hold of all pull and feature requests. That reduces the time I can spend with development short-term though. Still busy answering emails.

We now have a KeyCDN account and a new bare metal server at a different provider which we can use to accelerate our demo and docs (TBD).

lastzero:
Had a quick look. Needs to be polished before we can merge it into develop. Maybe there's a way to abstract database timing, so that not every single statement needs to be wrapped.

Makes sense to only add those headers in debug mode!?

benmccann:
Thanks for taking a look!

It's a good idea to abstract the database timing. It looks like Gorm almost, but not quite supports doing that. I filed a feature request for it there https://github.com/go-gorm/gorm/issues/4493. For now, I've removed the database timing from this PR. I can work on adding it back as a follow up PR when Gorm supports it.

Before submitting this PR, I had also looked into whether it might be possible to do the overall timing in a middleware instead of instrumenting each call and unfortunately it's not. The issue there is that Gin won't let you write headers in middleware after `Next` is called to handle the request. I think basically the idea is that headers must be written in the response before the body is and `Next` is (or at least can be) writing the body. There might be some way of doing something really hacky where you buffer the whole body into a string or something, but basically I think Gin is written with the idea that all responses can be streamed and so I don't think there's any way to really improve this PR on that point.

I think it makes sense to always write these headers. It should be very cheap to calculate and include. When I worked at Google and LinkedIn we instrumented our servers in much heavier ways with distributed tracing tools and still served pages with very low response times. It was nice to have the info always available since you can never predict when an issue may occur.

lastzero:
I'll close this PR for now since the CLA is not signed (we've reset the signatures), there are merge conflicts, and @benmccann probably doesn't need it anymore. Thanks a lot for your contributions!

PR #1393, opened 2021-06-28
Fix Mapbox map RTL support - Fix #1391

### LINKED ISSUE #1391: Map in "Places" displays Hebrew in left-to-right order and not right-to-left
Recently installed photoprism via docker-compose in a debian host environment - working great and it's been slowly indexing the photos. I have many photos geotagged in Israel, and all the Hebrew names of cities, etc are displayed with the Hebrew letters in left-to-right order, and not the proper right-to-left. For example, Petah Tikva is displayed in English properly, but the name in Hebrew underneath is displaying as:
הווקת חתפ

Instead of:
פתח תקווה

See screenshot:
image

Tried changing the different map types but it's all pulling from the same API. Weirdly, if you load up OSM directly, on their site it's displaying the proper right-to-left Hebrew names:

image
https://www.openstreetmap.org/#map=15/32.0173/34.7597

Wondering what could be the cause of this bizarre display. I can confirm that the Arabic is properly displaying right-to-left, so not really sure why Hebrew is displaying in reverse....

Any ideas?

haimkastner:
Load Mapbox RTL plugin when facing RTL language on the map.
Load the plugin lazy, so it will not affect current other languages' performance.

### DIFF
--- a/frontend/src/pages/places.vue
+++ b/frontend/src/pages/places.vue
@@ -272,6 +272,14 @@ export default {
       }).catch(() => this.loading = false);
     },
     renderMap() {
+      // In case of facing RTL language on the map, this plugin will be fetched (only by lazy!) and will set RTL properly support.
+      // See mapbox documentation here https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-rtl-text/
+      mapboxgl.setRTLTextPlugin(
+        'https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-rtl-text/v0.2.3/mapbox-gl-rtl-text.js',
+        null,
+        true // Lazy load the plugin
+      );
+      
       this.map = new mapboxgl.Map(this.options);
 
       this.map.addControl(new mapboxgl.NavigationControl({showCompass: true}, 'top-right'));


lastzero:
Thank you very much!

frontend/src/pages/places.vue
```diff
@@ -272,6 +272,14 @@ export default {
       }).catch(() => this.loading = false);
     },
     renderMap() {
+      // In case of facing RTL language on the map, this plugin will be fetched (only by lazy!) and will set RTL properly support.
+      // See mapbox documentation here https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-rtl-text/
+      mapboxgl.setRTLTextPlugin(
+        'https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-rtl-text/v0.2.3/mapbox-gl-rtl-text.js',
```
alexislefebvre, line 278: I wonder if that file could be integrated in PhotoPrism? It would remove the dependency on an external URL that can fail.

haimkastner: The map content is fetched on run-time from an external API in any case, and integrating it into PhotoPrism will increase the frontend bundle.

alexislefebvre: Fair enough. Thanks.

lastzero: It's somewhat of a privacy and stability issue indeed, like when their server is down or blocked by a firewall. We could store a copy of that file in our static assets folder and load it from there?

https://github.com/photoprism/photoprism/tree/develop/assets/static

haimkastner: @lastzero you're right, see the #1398 PR with this improvement.

PR #1398, opened 2021-06-30
Serve map RTL plugin from internal static server

haimkastner:


lastzero:
Excellent, thank you! :+1:

Issue #22, opened 2018-10-08
People: Facial Recognition


lastzero:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%24900%20Rewarded-%237E24E3.svg" />](https://issuehunt.io/r/photoprism/photoprism/issues/22)

**As a user, I want to tag familiar human faces so that I can find all photos of these persons.**

PhotoPrism must be able to detect and recognize faces. There are existing libraries for this, so we should try those first. See [Face Recognition](https://github.com/photoprism/photoprism/wiki/Face-Recognition) in our Wiki.

Here is a complete howto: https://hackernoon.com/face-recognition-with-go-676a555b8a7e

Acceptance Criteria:
- [x] You SHOULD use [go-face](https://github.com/Kagami/go-face) (unless you find a better library)
- [x] A unit test demonstrating face recognition with real images and faces MUST be delivered
- [x] A user interface MAY be developed and gives extra karma
- [x] You MUST update our Docker images in `/docker` as needed (go-face requires dlib as external dependency); let us know when we shall push new images to [Docker Hub](https://hub.docker.com/u/photoprism)
- [x] [Documentation](https://github.com/photoprism/photoprism/wiki/Face-Recognition) MUST be updated so that other developers know everything to continue working on this task

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $900.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($40.00)
- [<img src='https://avatars2.githubusercontent.com/u/54749?v=4' alt='totempaaltj' width=24 height=24> totempaaltj](https://issuehunt.io/u/totempaaltj) ($30.00)
- [<img src='https://avatars.githubusercontent.com/u/64777347?v=4' alt='uluvil3gcs' width=24 height=24> uluvil3gcs](https://issuehunt.io/u/uluvil3gcs) ($22.00)
- [<img src='https://avatars3.githubusercontent.com/u/467291?v=4' alt='urbiops' width=24 height=24> urbiops](https://issuehunt.io/u/urbiops) ($20.00)
- [<img src='https://avatars1.githubusercontent.com/u/17299920?v=4' alt='luxumbra27' width=24 height=24> luxumbra27](https://issuehunt.io/u/luxumbra27) ($10.00)
[...]

PR #1406, opened 2021-07-03
Face recognition POC with Facenet model using TF

rahul003:
I put together a prototype for face recognition using the coordinates of faces already detected from Pigo. Ref: https://github.com/photoprism/photoprism/issues/22#issuecomment-871576686 

FaceNet seems to be the best open source pretrained model available for face recognition. I got the pretrained model checkpoint from [FaceNet's original repo](https://github.com/davidsandberg/facenet) and created a SavedModel. I can share the script used to create SavedModel if needed. I added this savedModel file to this PR. (The model used is  [20180402-114759](https://github.com/davidsandberg/facenet#pre-trained-models) ). Since PhotoPrism already links against TF, I thought this was the best way to proceed with face recognition. It seems to work quite well too on the test images in `face` folder. 

Flow here is the following
- Crop image using the coordinates and size of face returned from Pigo detector, to create a jpeg for each face
- Resize image to 160x160 as needed for the model
- Run inference using Facenet with this face image, which returns a (512,) embedding vector.
- Compute euclidian distance using these embedding vectors which gives us a score of similarity (lower the distance, more similar). Threshold needs to be adjusted based on real world data and the UX plan in Photoprism for face recognition. In the unit test I set a value for which most pairs were identified correctly (51/55 possible pairs of faces in images)
 
Let me know your thoughts on using this model and this approach for Face recognition. To proceed there needs to be some thought on the plan for comparing faces across the whole library of the user. There's of course normalization and other processing (whitening is one step done in Facenet repo which I haven't done yet) which could improve accuracy. Switching to an efficient implementation of Euclidian distance, vectorization etc can improve performance. 

- [x] P.S. CI is failing because the order of processing files seems different, I'll rework the test.

### DIFF
--- a/internal/classify/tensorflow.go
+++ b/internal/classify/tensorflow.go
@@ -1,4 +1,4 @@
-package face
+package classify
 
 import (
 	"bufio"
@@ -30,8 +30,8 @@ type TensorFlow struct {
 }
 
 // New returns new TensorFlow instance with Nasnet model.
-func New(modelName string, modelsPath string, disabled bool) *TensorFlow {
-	return &TensorFlow{modelsPath: modelsPath, disabled: disabled, modelName: modelName, modelTags: []string{"photoprism"}}
+func New(modelsPath string, disabled bool) *TensorFlow {
+	return &TensorFlow{modelsPath: modelsPath, disabled: disabled, modelName: "nasnet", modelTags: []string{"photoprism"}}
 }
 
 // Init initialises tensorflow models if not disabled
@@ -75,7 +75,7 @@ func (t *TensorFlow) Labels(img []byte) (result Labels, err error) {
 	}
 
 	// Create tensor from image.
-	tensor, err := t.createTensor(img, "jpeg", 224, 224)
+	tensor, err := t.createTensor(img, "jpeg")
 
 	if err != nil {
 		return nil, err
@@ -160,10 +160,7 @@ func (t *TensorFlow) loadModel() error {
 
 	t.model = model
 
-	if (t.modelName == "nasnet") {
-		return t.loadLabels(modelPath)	
-	}
-	return nil
+	return t.loadLabels(modelPath)
 }
 
 // bestLabels returns the best 5 labels (if enough high probability labels) from the prediction of the model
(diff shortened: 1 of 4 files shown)


lastzero:
Thank you very much! This sounds perfect... still busy with other tasks, doing my best to test your PR as soon as possible :+1:

lastzero:
Finally have a bit of time, after upgrading packages and improving our multi-arch Docker image earlier this week...

Guess we can merge this, but without `assets/facenet/saved_model.pb`. Should put that on `dl.photoprism.org` to reduce the git repo size - especially in case the model needs to be updated.

lastzero:
@rahul003 Face embeddings will now be stored in the markers table while indexing. To improve performance, it probably makes sense to cache face thumbs so that they don't need to be recreated every time. As your code comment suggests, it probably also makes sense to normalize them further for best results (whiten?). Currently reading the FaceNet paper to get a better understanding.

Are there any conditions under which FaceNet may return more than one embedding per face (file)? I've noticed the return value is a nested array.

lastzero:
From my understanding, it seems best to perform clustering & matching in a background worker after indexing, ideally when no other workers are active. Results can be stored using `Face` and `Person` entities, so that we can match different looking faces with the same person if needed.

I'd try to cluster embeddings using https://github.com/mpraski/clusters, unless someone knows a better library / approach.

rahul003:
> @rahul003 Face embeddings will now be stored in the markers table while indexing. To improve performance, it probably makes sense to cache face thumbs so that they don't need to be recreated every time. As your code comment suggests, it probably also makes sense to normalize them further for best results (whiten?). Currently reading the FaceNet paper to get a better understanding.

Nice! Thanks for following up on this. I see that changes from this are merged, so closing this PR.

 
> Are there any conditions under which FaceNet may return more than one embedding per face (file)? I've noticed the return value is a nested array.

I believe that it'll return more than one embedding if we send a batch of faces as opposed to one face.

rahul003:
> From my understanding, it seems best to perform clustering & matching in a background worker after indexing, ideally when no other workers are active. Results can be stored using `Face` and `Person` entities, so that we can match different looking faces with the same person if needed.
> 
> I'd try to cluster embeddings using https://github.com/mpraski/clusters, unless someone knows a better library / approach.

I agree. It can be intensive I think, best to separate our from initial indexing. I notice that Apple Photos also does it in the background later.

janstadt:
Hey this looks great! Im a bit late to the party but will this/photoprism (or facenet by proxy) support the coral usb accelerator for helping out the cpu for detection? That would be a great addition to this of the detection process is intensive.

lastzero:
It's possible, actually just other TensorFlow libs required. Nothing we need to implement in photoprism directly.

MasterGoofy:
Do you know when this feature will be released?

lastzero:
PhotoPrism doesn't require changes. You "just" need to install a version of the TensorFlow library with support for your hardware. We currently don't have time to compile a custom version of TensorFlow for everyone, so it's only for experienced users so to say. On the other hand, we also didn't reach our funding goal yet. Once that is the case, it's easier for us to help with installation and configuration. It's not that we don't want to.

Helvio88:
What is the status of this? Is there a way to hack-implement this? I have an idle Google Coral and my CPU is running 90%+ lol

graciousgrey:
>  Is there a way to hack-implement this?

You "just" need to install a version of the TensorFlow library with support for your hardware.

Issue #374, opened 2020-07-01
Feature Request: Make frontend a progressive web app (PWA)


wiwie:
As a user I want to be able to pin a photoprism instance from my mobile browser to the Homescreen of my phone and then use the shortcut as if photoprism was a native app. This functionality is called progressive web app (PWA).

As far as i know all that is needed is to embed a PWA manifest json file that defines how the PWA should behave and look like.

I know there is a mobile app in development, but I think adding this functionality to the frontend is little effort and would provide a valuable mobile experience of photoprism while photoprism-mobile is still behind the frontend's feature set. I think especially in the early stages of the project where things change more rapidly this would make sense.

lastzero:
Already done 🌈

wiwie:
Ok. I missed the you already include a manifest. But something seems wrong or missing, as it is not working for me on Android. Tried with chromium, Firefox and Firefox preview.

I've experienced before that browsers are quite sensitive to the contents of the manifest file. Amongst others, the dimensions of the referenced icon image files.

wiwie:
The nextcloud manifest file works for me, so maybe you could try including the icon sizes they do?



```
{"name":"Nextcloud","start_url":"https:\/\/xxx","icons":[{"src":"\/apps\/theming\/icon?v=0","type":"image\/png","sizes":"128x128"},{"src":"\/apps\/theming\/favicon?v=0","type":"image\/svg+xml","sizes":"16x16"}],"display":"standalone"}
```

lastzero:
Works for me on on Android and iOS, all I had to do is click "add to home screen". We're still working on updated icons and favicons that match our new logo.

wiwie:
Yes which is also what I do. For me though it doesn't add the app to the Homescreen as a PWA, just as a regular shortcut that just opens up a browser tab when I click on it.

lastzero:
Our demo opens without tab, full screen. It will run within Chrome in any case as it needs a browser. It's possible Google requires HTTPS to hide the tab / site info.

lastzero:
![Screenshot_20200702_094946](https://user-images.githubusercontent.com/301686/86331618-a2e87800-bc49-11ea-8140-436be75334b7.jpg)

wiwie:
You are right. It works with the demo installation, but not with my own.

My own instance has cert+https as well, so that can't be the problem. I'm using nginx as reverse proxy, maybe it has something to do with that. I'll look into it and report back in case I find anything.

[...]

PR #1410, opened 2021-07-05
Add crossorigin="use-credentials" to manifest links

tpansino:
While setting this up in Docker for the first time behind an [OAuth2-Proxy](https://oauth2-proxy.github.io/oauth2-proxy), I noticed that the PWA fails to offer an installation on latest Android Chrome.

It looks like this is a result of the `manifest.json` [being blocked by CORS](https://stackoverflow.com/questions/6294622/html5-manifest-cache-behind-basic-auth) unless the authentication credentials are passed. Sure enough, if I set an exception for `static/manifest.json`, the installation is offered, similar to what the user found in [this](https://github.com/photoprism/photoprism/issues/374#issuecomment-653042041) issue.

The [fix](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin#example_webmanifest_with_credentials) appears to be to add `crossorigin="use-credentials` to all of the manifest links. I found 3 instances, please correct me if I missed any or was overly zealous.

Also - I could use some guidance on how to test this locally before merging. I didn't see anything in the docs.

### DIFF
--- a/assets/templates/index.tmpl
+++ b/assets/templates/index.tmpl
@@ -27,7 +27,7 @@
 {{template "favicons.tmpl" .}}
 
   <link rel="stylesheet" href="/static/build/app.css?{{ .config.CSSHash }}">
-  <link rel="manifest" href="/static/manifest.json?{{ .config.ManifestHash }}">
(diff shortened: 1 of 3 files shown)


lastzero:
Excellent, thank you very much!

Issue #1301, opened 2021-05-19
Docker: Permission issues in new Ubuntu 21.04 base image


royaltongue:
Photoprism container was working perfectly fine this afternoon, and I noticed a few minutes ago that a WebDAV upload failed.

Checked the container logs and saw error:
`/entrypoint.sh: line 37: photoprism: command not found`

I've tried pulling the latest image and recreating through Portainer, but no luck

lastzero:
You think it's a problem with our update? Does it update automatically? Can you try an older version? Intel or ARM CPU?

royaltongue:
I don't have Watchtower running on this stack, so I'm not sure what caused it actually

What image tag should I use to get an older version?

Intel CPU

lastzero:
Here's a list of all available images (tagged by date, which is the version):

https://hub.docker.com/repository/docker/photoprism/photoprism

Previous version is `photoprism/photoprism:20210505`

royaltongue:
`20210505` spins up without any errors

lastzero:
Then it might be the new Ubuntu version which has other permission defaults :/ We're going to look into this! In the meantime, you might have to change permissions inside the Docker image manually in a terminal. Thee docs / docker-compose.yml file for how to open one.

satcom886:
I just ran into this too. Also x86 Linux.

lastzero:
Seems `/photoprism` wasn't world executable after upgrading Ubuntu. What user are you running the container as?

lastzero:
Started a preview build for testing: https://docs.photoprism.app/getting-started/updates/#development-preview

satcom886:
> Seems `/photoprism` wasn't world executable after upgrading Ubuntu. What user are you running the container as?

I created a user for running multimedia services (Transmission, Jellyfin, Photoprism, etc.) to avoid media permission issues. `user: 1002:1002`

lastzero:
So, is it working now?

royaltongue:
Yep, preview image is good

satcom886:
> So, is it working now?

Just had a chance to test, yes it works fine with the preview image.

lastzero:
Thanks for testing! :heart:

PR #1432, opened 2021-07-14
Improved Dockerfile with some enhancements

BetaAthe:
I've made some changes to the Dockerfile:

- I've merged all ENV variables into one ENV statement to make the Dockerfile more readable.
- Running apt-get clean in a different RUN command than apt-get update doesn't make sense. Merging this two saves space in this layer.
- There is no need to umask if you then change the folder permissions.
- The home folder can have a 755 permission, instead of 777, and still allow access to root, photoprism and a made-up uid.
- There is no need to chown to root and chmod to 755 on COPY as this is the default behavior.
- You do not need to own the /photoprism subfolders with the user photoprism if you give them 777 permissions.
- You do not need to do chmod and find | xarg chmod, as the folders are already made with the 777 permission set.
- The 2 points above make /photoprims/bin and /photoprism/assets read-only for any user. I think that a binary and its components don't need write permissions. I have tested photoprism with these permissions and it works perfectly.
- I have removed the "VOLUME /var/lib/photoprism" as this forces the user to mount this volume. This is something that not all users want in their instance so, I believe that, being an optional volume, it should not be declared with a VOLUME command. For example, in my deployment, a new unnamed volume is created each time the container is started.

All these changes reduce the total size of the image by 100-200 MB after decompression. The Docker image is still over a gigabyte in size and I think we should try to see if we are able to reduce its size a bit more (perhaps by removing some dependencies or changing the base OS).


lastzero:
Thank you very much for your PR and advice!

> * I've merged all ENV variables into one ENV statement to make the Dockerfile more readable.

We had it like that once, but there was a reason we split it again. Either it caused issues, or it made the dev and prod images less comparable in structure.

> * Running apt-get clean in a different RUN command than apt-get update doesn't make sense. Merging this two saves space in this layer.

Same here, was like that. However adding this line for our new multi-arch builds seems to have caused issues:

```
[ "$TARGETARCH" = "arm" ] || apt-get install darktable; \
```

Note that darktable is not available for ARMv7 / 32bit. Optimizing this more wasn't on the top of our todo list as multi-arch builds take ages to run (which we can't afford right now).

> * There is no need to umask if you then change the folder permissions.

It's for newly created files and folders (like when users play with the container or photoprism creates files), not for the ones created / added in the Dockerfile.

From my understanding this should have permanent effect as it sets the default?

Permission issues are the Nr 1 reason when users report issues, so it's supposed to minimize the time we need to spend providing support and explaining Linux.

> * The home folder can have a 755 permission, instead of 777, and still allow access to root, photoprism and a made-up uid.

Some users do weird things and were repeatedly running into permission issues. Added this after somebody reported a "bug". Any major downside of using 777?

> * There is no need to chown to root and chmod to 755 on COPY as this is the default behavior.

Also when not building as root or the original files have other permissions? Remember to have added this after problems came up. As we don't want to break anything in the many different environments our users run this, so we're pretty cautious. If it doesn't hurt, I'd rather keep it for now.

> * You do not need to own the /photoprism subfolders with the user photoprism if you give them 777 permissions.
> * You do not need to do chmod and find | xarg chmod, as the folders are already made with the 777 permission set.

Remember to have added this after a user reported problems in his environment, but can't remember specifics.

Wasn't like this originally and it usually isn't required, I agree.

> * The 2 points above make /photoprims/bin and /photoprism/assets read-only for any user. I think that a binary and its components don't need write permissions. I have tested photoprism with these permissions and it works perfectly.

It's writable because we took this over from the dev environment where we need to replace it with new builds. Also we're running our tests there. If the environments are too different, it's possible the tests are green but the production image is broken. Going for simplicity instead of optimization.

You think it's a major security issue to have a binary that's writable? Many other ways to execute code, especially when running as root.

> * I have removed the "VOLUME /var/lib/photoprism" as this forces the user to mount this volume. This is something that not all users want in their instance so, I believe that, being an optional volume, it should not be declared with a VOLUME command. For example, in my deployment, a new unnamed volume is created each time the container is started.

This is required as of now as our backend generates API keys for geocoding & maps. As you may have noticed satellite world maps are not part of the Docker image due to their size - it's an external service we pay for.

Permission issues are the number 1 reason PhotoPrism doesn't work. If the storage path isn't writable, it exits. Now some people have automatic restarts enabled in Docker. That means we get a new API key request every few seconds. That's why there is a backup storage volume so that PhotoPrism can remember the key when it restarts a few seconds later - even with wrong permissions. On this positive side, this simplifies setup and users don't need to provide personal information or manually request a key first.

Once we have a different / better process in place, we are happy to remove this. It's really not the most beautiful solution but works for now.

> All these changes reduce the total size of the image by 100-200 MB after decompression.

Will permission changes also reduce storage size? I assume this is mostly because of `apt-get clean`? It should really be running in the same layer, once the conditional package issue works properly for all architectures. We would love to get help with fixing and testing that.

> The Docker image is still over a gigabyte in size and I think we should try to see if we are able to reduce its size a bit more (perhaps by removing some dependencies or changing the base OS).

Sadly there are a ton of dependencies and also large TensorFlow models required for what we do. We don't have the resources to maintain a tailor made Linux distribution with custom builds. Aware this may somewhat reduce the image size.

BetaAthe:
Hi, I didn't found anything about the splitted ENV statement, but if being in two places is better for consistency, I agree to let it be this way.

This line `[ "$TARGETARCH" = "arm" ] || apt-get install darktable; \` shouldn't cause any trouble. I saw that in commit 6b7745e it was declared with `&&` at the end, so the logic was: If arm, do nothing, if not arm, install darktable and autoremove/clean/etc. This pull request ends with an `;`, making the logic be like: if arm, do nothing, if not arm install darktable. Then, autoremove/clean/etc.

I don't know where the issue could be as they are, effectively, two different commands.

> From my understanding this should have permanent effect as it sets the default?

I found this [blog post](https://codeyarns.com/tech/2017-07-21-how-to-set-umask-for-docker-container.html) about umask and Docker. It seems like setting umask in a RUN statement will only has its effect during the building phase. If you want to umask 0000 the photoprism directory, you will need to add that command to the entrypoint script.

> Any major downside of using 777?

In this context, I don't think so. The only difference between 777 and 755 is that with the latter a user cannot add a custom folder to /photoprism using a made-up uid. I can't understand why a user would do the former, let alone also do it with the latter. Although if you think that this might be a source of issue reports in the future, I agree to have it set to 777.

> Also when not building as root or the original files have other permissions?

The owner is always root by default, from [the Docker reference](https://docs.docker.com/engine/reference/builder/#copy):

> All new files and directories are created with a UID and GID of 0, unless the optional --chown flag specifies a given username, groupname, or UID/GID combination to request specific ownership of the copied content. 

But `--chmod` is not well documented, so I don't know if this applies always. I agree to set the `--chmod=755` as we can't be sure this is the default behavior.

I found this commit a0f3ab5 about this issue #1301 about the /photoprism subfolders with permission set 777.  I think this commit was a little overkill for the issue, as `chmod a+rwx /photoprism` should be enough. With these permissions, photoprism is still world executable.

> It's writable because we took this over from the dev environment where we need to replace it with new builds

I don't know enough of this project, so please correct me, but I don't think you should replace binaries with a custom uid. If you run the upgrade with root or the photoprism user, you must be able to replace the binary and its components with the 755 permission set. Then, you can run other tests with root, photoprism or other made-up uid only reading and running the tests. I don't think it's a major security issue, but I think it shouldn't be done.

> That's why there is a backup storage volume so that PhotoPrism can remember the key when it restarts a few seconds later

There's a thing or two I'm missing here, so please tell me if I'm wrong. I thought that /var/lib/photoprism was only for YAML backups, as I'm currently running the latest stable version and this directory is just empty. Also, when you restart a container, yo don't create a new one, you just restart it. This means that even if you don't have a volume for this path, the files it has are still inside the container. If a user don't restarts its container but recreates it, you are still going to lose whatever is in this directory, unless the user mounts this volume explicitly, as any time the container is recreated, a new volume is mounted for it.

> Will permission changes also reduce storage size?

Yes, check layer 19 of the [latest published Docker image](https://hub.docker.com/layers/photoprism/photoprism/latest/images/sha256-40e99d97615164f1a10c2a674301ced878030a7a79df1b8b40d70f096c4112a9?context=explore). Also, apt-get clean running in the same layer as apt-get update saves over 100 MB too.

lastzero:
Currently building new images based on your recommendations. To my surprise, removing `--chown=root:root` here caused a build error later (even though the docs say it's default):

```
# Copy scripts to /root/.local/bin
COPY --chown=root:root --chmod=755 /docker/scripts/*.sh /root/.local/bin/
```

I'll push my changes for you to review once dev and prod images are building without error.

Good catch with using `; \` instead of `& \` at the end of the conditional darktable installation! :+1:

lastzero:
> > That's why there is a backup storage volume so that PhotoPrism can remember the key when it restarts a few seconds later
> 
> There's a thing or two I'm missing here, so please tell me if I'm wrong. I thought that /var/lib/photoprism was only for YAML backups, as I'm currently running the latest stable version and this directory is just empty. Also, when you restart a container, yo don't create a new one, you just restart it.

It's a dedicated backup volume used for SQL database dumps as well, so that they don't get lost when upgrading PhotoPrism and/or there are issues with the storage path:

  https://github.com/photoprism/photoprism/blob/develop/internal/commands/backup.go#L98

Photo and Album YAML backups will be created in the storage path by default:

https://github.com/photoprism/photoprism/blob/871a5301f607ba23bfb29320e51920be0ab6913f/internal/config/fs.go#L364

So if you lose one volume, you should still have a backup. Wouldn't have implemented it like this for professional users.

If you don't like the anonymous backup volume you should be able to mount it to an external folder or named volume instead.

lastzero:
Development image was updated, and is a bit smaller now:
https://hub.docker.com/r/photoprism/development/tags?page=1&ordering=last_updated

lastzero:
New preview build has been started:
https://docs.photoprism.app/getting-started/updates/#development-preview

lastzero:
Preview build failed with:

```
#53 63.70 dpkg: error processing archive /tmp/apt-dpkg-install-L8fMxG
          /0-libapparmor1_3.0.0-0ubuntu7_arm64.deb (--unpack):
#53 63.70  dpkg-deb --control subprocess returned error exit status 1
#53 63.70 Error while loading /usr/sbin/dpkg-split: No such file or directory
#53 63.70 Error while loading /usr/sbin/dpkg-deb: No such file or directory
```

Seems related to buildx and qemu, which of course is installed on the server. Tested it yesterday. Moved `scripts/install-qemu.sh` to the main `.drone.yml` file, hope this helps. Help with figuring out the root cause would be much appreciated.

lastzero:
New build was successful: https://docs.photoprism.app/getting-started/updates/#development-preview

It's a bit smaller now (compressed) but not a lot... completely removing chmod may reduce the size further, but I'd rather not try this now as I've already spent a full development day with this, and our users are desperately waiting for new features.

BetaAthe:
> completely removing chmod may reduce the size further, but I'd rather not try this now as I've already spent a full development day with this

Understandable, but this is the thing that gives an extra 100 MB to the final image.

But yes, I think this is better than before, so thank you for your work!

lastzero:
Would it be ok to close this PR, and update our Developer Guide to capture the current state and collect ideas for improvement instead? See https://docs.photoprism.org/developer-guide/technologies/docker/

gerroon:
It seems like the develop tag which is referenced in the repo does not exist.

https://github.com/photoprism/photoprism/blob/develop/docker-compose.yml


```
docker pull photoprism/photoprism:develop
Error response from daemon: manifest for photoprism/photoprism:develop not found: manifest unknown: manifest unknown
```

lastzero:
It's only available in development environments, built  when you clone the repo and run `docker-compose up`

PR #1443, opened 2021-07-22
Fail entire build script if any command fails

thomaseizinger:
At the moment, the build script exits with 0 even if any of the commands fail.
Setting `-e` fixes this.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
+
 PHOTOPRISM_DATE=`date -u +%y%m%d`
 PHOTOPRISM_VERSION=`git describe --always`
 


lastzero:
Thanks! I'll merge this in the expectation you've tested it :)

thomaseizinger:
That was quick! Thank you :)

To be perfectly honest, I didn't test it with _this_ particular shell script :see_no_evil: 
But I've ran into a missing `set -e` so many times already that I pretty confident that this is what is missing.

Your comment did prompt me to also check _this_ particular shell script and yes, it also works here! Script correctly fails now if any of the commands fail.

For context: I've hit [this](https://github.com/golang/go/issues/43938) weird bug.

lastzero:
Should also add it then to the new buildx.sh script for multi-arch builds.

Issue #584, opened 2020-11-06
Labels: Show photo count


danielo515:
Hello.
Right now, I have a lot of labels, but very few photos per label.
It will be awesome if, before opening a label, I can see a count of how many images belong to it, so I get an idea if opening it is worth or not. Specially when the count it's 1, it's pointless to open such a label to see something you are already seeing.

Regards

lastzero:
Our API returns the count in the `PhotoCount` field, we just haven't found a nice way to display it yet:

![Screenshot from 2020-11-08 17-31-38](https://user-images.githubusercontent.com/301686/98470770-480ef080-21e8-11eb-970c-134492b9f24d.png)

danielo515:
This is how they do on QuMagie (some inspiration)
![image](https://user-images.githubusercontent.com/2270425/98580462-4a447e00-22c0-11eb-8b30-04da1d9be221.png)

hartraft:
I think the photocount would be nice to have as well. What would be a better preference? 
Option 1: (similar to how albums are shown and translated but more verbose)
![image](https://user-images.githubusercontent.com/788583/126811103-a8b96166-3e65-4a38-a9e6-55690a831cb4.png)

Option 2: (plain and simple number)
![image](https://user-images.githubusercontent.com/788583/126811191-24163cd2-40a6-44e6-85af-52da1ddb6177.png)

lastzero:
Note that label photo counts are currently not updated in real time:

https://github.com/photoprism/photoprism/blob/develop/internal/entity/photo_counts.go#L89

We had been using (automatically updated) views previously, but they were not compatible with all supported databases.

At the moment, counts can be reliably used to hide labels without any result. Counts > 1 may be imprecise at times e.g. while indexing. It may well lead to frustrated users filing bug reports because numbers don't match. Frequently happens for other counts as well. It's understandable as users are afraid of data loss, and bugs in general.

fensminger:
I updated photoprism to last preview. It is working.
![image](https://user-images.githubusercontent.com/1978956/134763225-3d8f0d1e-91dd-43de-8a3a-69d50df10465.png)

PR #1444, opened 2021-07-23
Add photocount to labels page

hartraft:
Simple change to add the photocount. Depends on implementation choice
Issue #584

### DIFF
--- a/frontend/src/pages/labels.vue
+++ b/frontend/src/pages/labels.vue
@@ -113,7 +113,7 @@
                     @save="onSave(label)"
                 >
                   <span v-if="label.Name" class="body-2 ma-0">
-                      {{ label.Name | capitalize }}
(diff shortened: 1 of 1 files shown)


lastzero:
Note that label photo counts are currently not updated in real time:

https://github.com/photoprism/photoprism/blob/develop/internal/entity/photo_counts.go#L89

We had been using (automatically updated) views previously, but they were not compatible with all supported databases.

At the moment, counts can be reliably used to hide labels without any result. Counts > 1 may be imprecise at times e.g. while indexing. It may well lead to frustrated users filing bug reports because numbers don't match. Frequently happens for other counts as well. It's understandable as users are afraid of data loss, and bugs in general.

graciousgrey:
@hartraft With the latest release we improved the counts and display them :).

lastzero:
Counts now are at least correct when using MariaDB or MySQL. They may still be too high when using sqlite, that's why we didn't show them previously (due to limited SQL support).

Issue #28, opened 2018-10-08
Stacks: Search + manually stack similar images


lastzero:
Perceptual hashing is already implemented, all we need is a user interface and maybe an easy way to merge multiple files into one photo.

rkkoszewski:
What about non-destructive merging? Where the main picture is shown as normal, but the similar pictures are still accessible from an option at the image. Just like the time-shift option in many phones, which take a burst of pictures when you make a picture and you can select the best one retrospectively while still preserving all alternatives.

lastzero:
@rkkoszewski Yes, this is how it should look like when it's done. The database has a table for photos and files for that reason, so you can have multiple files for a photo.

lastzero:
We've recently added a "Group by similarity" sort option to our photo search, it uses perceptual image hashes: https://demo.photoprism.org/photos?view=cards&order=similar&public=true&quality=3

spiritbro1:
@lastzero when i try that link the photo doesnt order by similiarity just look at this photo of a girl : 
![image](https://user-images.githubusercontent.com/62529025/99207132-bd358380-27ef-11eb-82f2-b579143976af.png)
this is the first photo of the girl on the right side
![image](https://user-images.githubusercontent.com/62529025/99207162-d0e0ea00-27ef-11eb-9d73-9608db728036.png)
this is the second photo way below from the first photo can you elaborate more on what do you mean by similiarity here or is the algorithm still wrong because the order i think still not right

timdonovanuk:
@graciousgrey am I overlooking something or is stacking not in the currently API (even though its clearly implemented)? Unstacking seems to be! I was planning to call it to stack a whole bunch of semi-duplicates I have.

lastzero:
Stacking happens based on the file name or metadata. There is currently no way to manually stack random files without renaming / moving them or changing their metadata in any way. For example you could set a unique image ID (Exif) or document ID (XMP) and then reindex to get files stacked.

timdonovanuk:
Ah righty, understood. Thanks @lastzero !

GlassedSilver:
I would definitely welcome an option to not stack based on perceptional hashing, at least not automatically. Maybe with a review process and manually selecting my favored pick.

**Sample:**
![Screenshot 2022-03-28 225922](https://user-images.githubusercontent.com/1912133/160486255-e48070d0-aa64-4ef4-86ba-47c07bd8b941.png)


[...]

PR #1457, opened 2021-08-05
Library: Add duplicate search with delete function

SNThrailkill:
Draft PR of initial implementation. Will definitely need polish and tweaks.

### DIFF
--- a/frontend/src/component/components.js
+++ b/frontend/src/component/components.js
@@ -38,6 +38,7 @@ import PPhotoToolbar from "./photo/toolbar.vue";
 import PPhotoCards from "./photo/cards.vue";
 import PPhotoMosaic from "./photo/mosaic.vue";
 import PPhotoList from "./photo/list.vue";
+import PPhotoDuplicates from "./photo/duplicates.vue";
 import PPhotoClipboard from "./photo/clipboard.vue";
 import PAlbumClipboard from "./album/clipboard.vue";
 import PAlbumToolbar from "./album/toolbar.vue";
@@ -58,6 +59,7 @@ components.install = (Vue) => {
   Vue.component("PPhotoCards", PPhotoCards);
   Vue.component("PPhotoMosaic", PPhotoMosaic);
   Vue.component("PPhotoList", PPhotoList);
+  Vue.component("PPhotoDuplicates", PPhotoDuplicates);
   Vue.component("PPhotoClipboard", PPhotoClipboard);
   Vue.component("PAlbumClipboard", PAlbumClipboard);
   Vue.component("PAlbumToolbar", PAlbumToolbar);
--- a/frontend/src/component/navigation.vue
+++ b/frontend/src/component/navigation.vue
@@ -332,6 +332,15 @@
             </v-list-tile-content>
           </v-list-tile>
 
+           <v-list-tile to="/library/duplicates" class="nav-hidden" @click.stop="">
+            <v-list-tile-content>
+              <v-list-tile-title :class="`menu-item ${rtl ? '--rtl' : ''}`">
+                <translate key="Duplicates">Duplicates</translate>
+                <span v-show="config.count.duplicates > 0" :class="`nav-count ${rtl ? '--rtl' : ''}`">{{ config.count.duplicates }}</span>
+              </v-list-tile-title>
+            </v-list-tile-content>
+          </v-list-tile>
+
           <v-list-tile to="/library/hidden" class="nav-hidden" @click.stop="">
             <v-list-tile-content>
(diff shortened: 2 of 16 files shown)


SNThrailkill:
Sample API Payload:
http://localhost:2342/api/v1/duplicates/5352b00121b026185ab1ab400077c655c8dba877

Sample API Response:
[
  {
  "Name":"2021/04/20210402_154516_0895C5F8 - Copy.jpeg",
  "Root":"/",
  "Hash":"5352b00121b026185ab1ab400077c655c8dba877",
  "Size":1802482,
  "ModTime":1622606866
  },{
  "Name":"2021/04/20210402_154516_0895C5F8.jpeg",
  "Root":"/",
  "Hash":"5352b00121b026185ab1ab400077c655c8dba877",
  "Size":1802482,
  "ModTime":1622606866
  }
]

SNThrailkill:
Sample of current UI implementation
sample

ark-:
Out of interest, what method is used to compare the photos and determine a match? I note you match a hash but I'm not familiar with how Photoprism determines this hash.

lastzero:
> Out of interest, what method is used to compare the photos and determine a match? I note you match a hash but I'm not familiar with how Photoprism determines this hash.

PhotoPrism uses [SHA-1](https://en.wikipedia.org/wiki/SHA-1) checksums to detect exact duplicates. While it's older / shorter than e.g. SHA-2, you don't need to be afraid of security issues as it's not used as cryptographic hash. In addition we index image properties like colors and contrast that could be used to find similar images.

ark-:
Thanks Michael. I wonder if (further down the line) there could be an image "similarity" algorithm. I know your thoughts on this are that metadata is better at determining a match than any RGB comparison but perhaps it could be a suitable fallback when no metadata is available (e.g. no EXIF data or useful filename). 

The https://github.com/qarmin/czkawka project has a fairly good image duplicate feature. However it is written in Rust and so is the library it uses https://github.com/abonander/img_hash.

lastzero:
You can already group / sort by similarity:

https://demo.photoprism.org/browse?view=cards&order=similar&public=true&quality=3

This was specifically built for linear sorting - not to monitor unlicensed image usage, or find derived works with a near 100% confidence. There are other algos for this. We'll look into it when our roadmap has space for this.

ark-:
Very interesting! Didn't notice that feature so thanks for pointing it out.

G2G2G2G:
There's ahash, phash, dhash, etc for "similar" images, regular cryptographic hashes will give you exact matches, if we want to add similar-like image search (maybe you check a box for exact matches vs similar) we can go deeper in the future

Some links:
https://stackoverflow.com/questions/75891/algorithm-for-finding-similar-images

https://github.com/JohannesBuchner/imagehash/blob/master/find_similar_images.py

http://www.phash.org/ (strict but not as strict as cryptographic hash) 

Unfortunately not many things written directly in Go, it is annoying to have to compile stuff per system etc.

graciousgrey:
@G2G2G2G implementing other algorithms than sha1/perceptual hashing to find exact duplicates or similar files won't be part of this issue. This is about being able to delete already detected duplicates via the UI. I guess the title "advanced duplicate handling" is a bit misleading.

Ideally compressed, cropped or edited images are already stacked e.g. in case they have been taken at the same place and time, have related filenames or share the same document id. Perceptual hashing is implemented as well and we have a open ticket for providing a UI for it: https://github.com/photoprism/photoprism/issues/28

We know this won't catch all similar images, especially in case files have no metadata. Implementing other algorithms is something we can do later, feel free to open a separate issue for this :)

stbenjam:
Are there plans to match and allow deletion of perceptual hash matches as well? I know they stack, but I ended up importing about half my library twice, and I really want the dupes gone.  The first import was with a piece of software that compressed, resized, converted, etc so sha's don't match.

graciousgrey:
@stbenjam In case your similar images are already stacked, you find a delete button on the edit dialog's file tab, see https://docs.photoprism.org/user-guide/organize/stacks/#remove-not-primary-files-permanently. We also have a ticket to display very similar images (based on percepual hashes) that have not been stacked to the user, so that they can be manually merged. In this case a user could decide to delete one instead of merging (https://github.com/photoprism/photoprism/issues/28)

lastzero:
@SNThrailkill My apologies for the long delay! Do you still want us to work with you on this PR? What's still missing (besides fixing conflicts) before it's ready for release?

lastzero:
I'm sorry to have to close this PR for now, as we cannot easily merge it and have not received any feedback. If you have time to get the changes to a point where we can merge them without further refactoring and testing, feel free to reopen the PR or create a new one. Thank you very much for your contribution!

Issue #153, opened 2019-12-08
Discover: Display a random set of pictures


tonydub:
Hi,

We could rediscover and enjoy memories you haven't seen in years by displaying a random "sample" of images with a new "Random" option among the possible choices in "Sort By" list.

lastzero:
Great idea! Since we still change the database and index a lot, we don't test with many pictures - but for large collections, a random view makes sense.

tonydub:
Glad to see how this project will evolve.

lastzero:
Our idea is to add a "Discover" area where you can find photos by similarity, by color, by season or just random. This is a very first draft:

![Screenshot 2019-12-29 at 15 00 27](https://user-images.githubusercontent.com/301686/71557845-2aabba80-2a4c-11ea-9371-a740cfc28051.png)

Adding a random option to the regular search might turn out to be tricky as it needs to work together with the other filters and with all database backends. When doing a database search, you want to make the result set as small as possible to fit everything in memory. So first selecting as much as possible just to randomly change position (basically removing results from the top) is not a good strategy. In fact you expect a different result every time you click the search button plus when you scroll down to load additional results, photos must not repeat. Maybe easier done in Go and then just tell the db to return the randomly selected IDs, although I didn't spend much time investigating other solutions just yet (given the amount of other work we have to do).

lastzero:
Added more tabs to get a better impression:

![Screenshot 2019-12-29 at 15 52 05](https://user-images.githubusercontent.com/301686/71558402-2e8f0b00-2a53-11ea-9a1c-5f961acd383e.png)

lastzero:
FYI: Discover area probably won't be part of the release as we received very little feedback and no pull requests so far.

JamesSwift:
Would it be possible to implement just the "random sort" option in search view as the discover feature is not going ahead? Even if it is "truly random" and not the "pseudo random" where we expect to always see different images. The idea of continually scrolling and seeing more and more random images would be great.

While assessing different solutions for self hosting photos I tried PhotoStructure. PhotoPrism is better and has a better roadmap IMO, but one thing I loved was that the home view of PhotoStructure defaulted to random. I found so many nice memories while trying it out.

graciousgrey:
@bernsel feel free to reach out via hello@photoprism.app

bernsel:
I have emailed you today.

lastzero:
Thanks! Sorry for the wait, replied today :)

[...]

PR #1477, opened 2021-08-16
Random photos sorting

kvalev:
related to #153

### DIFF
--- a/frontend/src/component/photo/toolbar.vue
+++ b/frontend/src/component/photo/toolbar.vue
@@ -193,6 +193,7 @@ export default {
           {value: 'name', text: this.$gettext('Sort by file name')},
           {value: 'similar', text: this.$gettext('Group by similarity')},
           {value: 'relevance', text: this.$gettext('Most relevant')},
+          {value: 'random', text: this.$gettext('Random')},
         ],
       },
     };
--- a/internal/entity/const.go
+++ b/internal/entity/const.go
@@ -9,6 +9,7 @@ const (
 	SortOrderSimilar   = "similar"
 	SortOrderRelevance = "relevance"
 	SortOrderEdited    = "edited"
+	SortOrderRandom    = "random"
 
 	// Unknown values:
 	UnknownYear  = -1
--- a/internal/query/photo_search.go
+++ b/internal/query/photo_search.go
@@ -27,10 +27,10 @@ func PhotoSearch(f form.PhotoSearch) (results PhotoResults, count int, err error
 	// Base query.
 	s = s.Table("photos").
 		Select(`photos.*, photos.id AS composite_id,
-		files.id AS file_id, files.file_uid, files.instance_id, files.file_primary, files.file_sidecar, 
-		files.file_portrait,files.file_video, files.file_missing, files.file_name, files.file_root, files.file_hash, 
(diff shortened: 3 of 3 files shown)


graciousgrey:
Thank you, we will have a look as soon as possible :)

lastzero:
How does this work with count and offset? Did you test it with MariaDB and SQLite?

Users typically expect results not to repeat even if they are random...

dror3go:
For MySQL, a random `seed` should be generated and passed to the `RAND()` function.
AFAIK, SQLite doesn't support a seed for its `RANDOM()` function.

alexislefebvre:
@kvalev why did you close this PR? You can leave it open, it will be treated at some point.

orisdorch:
Looks like this was perhaps closed due to https://github.com/kvalev/photoprism/pull/13 ?

lastzero:
Yes, we just don't have enough resources to work on all 200+ issues at the same time... Doing our best! 👍

lastzero:
Random sorting must work on all databases and without breaking stacks. So it turned out to be more work. Thanks for sponsoring us! 🌈

orisdorch:
You guys are doing awesome work - glad to contribute in some small way.. How many DBs are you supporting? It looks like the initial approach attempted to leverage each database's randomization function for the sort.. Would it make more sense to add a new field like photos.photo_randomid so you could simply sort by this field rather than using each database's randomization function? Then you could also start at a random number between 1 and max(photos.photo_randomid), so it wouldn't always start at photos.photo_randomid = 1..

It may be a pain to ensure uniqueness of randomized IDs - if you could randomly reorder a series of numbers between 1 and length(photos), then append them sequentially to photos.photo_uid (i.e. if a randomized list of 1 to 4 is '3,2,4,1', these values would sequentially associated with uids '1,2,3,4').. I imagine a randomizer job that would populate entries with null photos.photo_randomid after the indexer runs, perhaps with the possibility or re-randomizing everything as a seperate job.. 

Code I saw in the prior attempt at randomized sort:
```
	case entity.SortOrderRandom:
		switch DbDialect() {
		case MySQL:
			s = s.Order("RAND()")
		case SQLite:
			s = s.Order("RANDOM()")
		default:
			err := fmt.Errorf("unknown sql dialect %s", DbDialect())
			log.Errorf("photos: %s", err)
			return results, 0, err
		}
	default:
		s = s.Order("taken_at DESC, photos.photo_uid, files.file_primary DESC")
```

gerroon:
Is this implemented?

graciousgrey:
@gerroon No,  it was closed by the contributor before it was fully implemented and tested. The related issue is still open: #153 

- Docs need to be updated

lastzero:
Note that we have added support for a "random" sort order to our backend API, see https://github.com/photoprism/photoprism/commit/47defc861c17ca1348eb0193b9ea84ce2708f228.

kvalev:
> How does this work with count and offset? Did you test it with MariaDB and SQLite?
> 
> Users typically expect results not to repeat even if they are random...

For documentation purposes, I have to mention that the current implementation also suffers from this problem, so as a user dont be surprised if you see a bunch of repeating photos.

lastzero:
See:
- https://github.com/photoprism/photoprism/issues/153#issuecomment-1408480166

PR #1577, opened 2021-09-30
updated requirements

JanithKE:


lastzero:
Thank you! Does that read better for a native speaker? We actually recommend running PhotoPrism on a strong server with as many cores as possible, 2 is just the bare minimum. Same for memory.

README.md
```diff
@@ -42,7 +42,7 @@ Step-by-step installation instructions for our self-hosted [community edition](h
 on [docs.photoprism.org](https://docs.photoprism.org/getting-started/) -
 all you need is a Web browser and Docker to run the server. It is available for Mac, Linux, and Windows.
 
-We recommend hosting PhotoPrism on a server with **at least 2 cores** and **4 GB of memory**.
+We recommend hosting PhotoPrism on a server with ** 2 cores** and **4 GB memory respectively**.
```
alexislefebvre, line 45: Why remove _at least_? Why add _respectively_ here?

*respectively* is used to associate several lists: https://en.wiktionary.org/wiki/respectively There are only 2 items here, there's no point to use it here in my humble opinion.

alexislefebvre:
It looks like the commit was merged: 85beb0c9a42541300261901ecc790f7ee7aef46c

alexislefebvre:
This part was removed: https://github.com/photoprism/photoprism/commit/f1bf3d27b8d3aa63cc026c54ad0e3f61d3a60b38#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L45

PR #1611, opened 2021-10-08
Config: Allow bypassing low memory suggestion

petertrr:
After one of the recent releases I found out that tensorflow is disabled on my device (with 1.7 GiB of memory as reported by `free`). I understand the concern, but I wanted to try to run tensorflow anyway and see how bad it is. So I added a way to bypass this check with an environment variable. I named it `unsafe_` to emphasize that it's not stable.

It turned out that my device still can run image recognition with these resources, so I would really like to have a way to disable low memory check. I'm not very familiar with Go, so if my change is not properly implemented or can be enhanced, I'll be happy to do it.

Another note, when tensorflow is disabled because of low mem, a checkbox 'Disable tensorflow' on frontend is checked and cannot be unchecked.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -72,6 +72,11 @@ func init() {
 	TotalMem = memory.TotalMemory()
 	LowMem = TotalMem < MinMem
 
+	if (os.Getenv("UNSAFE_IGNORE_LOW_MEM") == "true") {
+		LowMem = false
+		log.Warnf("Ignoring memory requirements, because UNSAFE_IGNORE_LOW_MEM is set to true")
+	}
+
 	// Init public thumb sizes for use in client apps.
 	for i := len(thumb.DefaultSizes) - 1; i >= 0; i-- {
 		name := thumb.DefaultSizes[i]


lastzero:
What threshold would you be comfortable with? We received a ton of "bug" reports by people who didn't have enough memory and no swap configured.

petertrr:
> What threshold would you be comfortable with? We received a ton of "bug" reports by people who didn't have enough memory and no swap configured.

I understand your concern, so I suggested not to simply reduce the threshold, but to allow user to explicitly disable it. I think that current default value of 2 GiB is reasonable, but if I want to experiment with my setup, I now have no other option other that to reduce or turn off the threshold in code and recompile the entire app.

petertrr:
I'll close this PR as I see the feature is implemented in 09f50fc

lastzero:
Yes, I was just going to ask if you're happy with this? I've changed the variable name to `PHOTOPRISM_UNSAFE` as we've decided to prefix all our variables with `PHOTOPRISM_`. Also you didn't sign the CLA yet, so I wasn't allowed to merge the PR directly.

petertrr:
> Yes, I was just going to ask if you're happy with this? I've changed the variable name to `PHOTOPRISM_UNSAFE` as we've decided to prefix all our variables with `PHOTOPRISM_`. Also you didn't sign the CLA yet, so I wasn't allowed to merge the PR directly.

Yes, no problem at all. Thanks!

PR #1657, opened 2021-10-24
CI: Add GitHub action config for building Docker image

leozz37:
Implement CI for Docker image build

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,11 @@
+on: [push, pull_request]
+name: Docker
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


ntimo:
As far as I know the Photoprism Team uses Drone to build the docker image. And to run tests.

lastzero:
Right, we use Drone CI internally. It would be an option to use GitHub for testing pull requests from external contributors.

lastzero:
Thanks a lot for your contribution! I'll close the PR for now as we are building on our own 16 core Ryzen server. Cross compiling for all architectures is too slow otherwise.

PR #1688, opened 2021-10-30
Update docker-compose.yml

bigbagoshwag:


lastzero:
The default should stay 1000 as this the first regular user id on most Linux distributions (in the past it was 500, but never 997).

Issue #799, opened 2021-01-03
UX: Improve Accessibility / Enable Page Zoom


Jab2870:
Hi,

I see that commit b716935e259944b52f849f6e6497a9fee95e0c4d disables page
zooming. As someone who makes heavy use of this feature, would you consider
undoing the change? For people who have weak eyesight, this change could cause
significant issues.

The ability to increase the size of everything is, for me, a huge advantage that
web apps provide over native.

Here are a few resources that you might find helpful

https://www.a11yproject.com/posts/2013-01-14-never-use-maximum-scale/
http://www.javierusobiaga.com/blog/stop-using-the-viewport-tag-until-you-know-ho/

lastzero:
Someone else asked for this. Guess we'll add yet another config flag! Thanks for letting us know :)

lastzero:
@Jab2870 Are there additional accessibility features we may implement to make using PhotoPrism easier for you?

Jab2870:
Hi @lastzero,

I have noticed that when tabbing through the application, it is not always clear
which elements currently have focus. However, if you like, I will create a
separate issue for that.

lastzero:
Sounds like this is more work, also I'm not sure how much influence we have on this since we're using a UI framework.

Maybe you (or some one else) finds something in the docs: https://v15.vuetifyjs.com/en/getting-started/quick-start

lastzero:
Let us know if there are any obvious "quick wins", like the zoom. Might be able to implement those as part of this issue.

Jab2870:
I used to be a front end dev although I haven't done much with vuetify. I will
have a read through the docs and let you know, might be able to make some pull
requests for you as well although it probably won't be until next weekend now.

lastzero:
You may now change this in config/settings.yml - we'll also add Accessibility settings to our UI but in a later release as it requires translations & documentation.

See https://github.com/photoprism/photoprism/releases/tag/210104-7f9e806a

PR #1716, opened 2021-11-09
Search: Basic zoom feature for Mosaic result view

radsto:
For the MOSAIC view only, adds a basic zoom slider on the toolbar for the following tabs : 

- Search
- Videos
- Favorites
- Calendar albums
- Private

Hi there ! I thought that would be useful because thumbs are really tiny on my huge screen. Please be kind with me since it's my first PR.

Any comments appreciated :-)


graciousgrey:
We will have a look as soon as possible :). Note: This might take a few months.

Sohalt:
I don't think it makes sense to clutter the toolbar with a slider for this, since you'll rarely adjust the thumbnail size, i.e. you'll set it once to the desired size for a screen and then leave it at that.
I think a better solution would be to use CSS media queries to query the screen size and dpi and adjust the thumbnail size based on that. That way for every screen size there would be a (hopefully) sensible default without requiring any interaction on the user's part and without cluttering the toolbar.

inthreedee:
There are good reasons users might want to have control over the size of their thumbnails, for example accessibility reasons https://github.com/photoprism/photoprism/issues/799. Setting standard sizes based on screen resolution sounds like a nice clean solution, but ignores accessibility and personal preference.

I think a zoom slider would be a significant improvement to the photo viewing experience.

lastzero:
@heikomat From my understanding, this would interfere with the virtualization you have added? In that case, I don't see how and when we could merge this :/

lastzero:
I'm sorry to have to close this PR for now, as it seems we can't easily merge it. We also haven't received any feedback on the technical concerns outlined above. Should you have time to get the changes to a point where we can merge them without further refactoring and testing, feel free to reopen the PR or create a new one. Thank you very much for your contribution!

PR #1726, opened 2021-11-12
Places: Add Normalization Mappings for BW and RLP (Germany)

hutattedonmyarm:


lastzero:
Great, thank you! Didn't see those 👍

Issue #1664, opened 2021-10-25
Places: Normalize Location Names


lastzero:
Location metadata sometimes contains abbreviations for US states instead of their full name:

https://assets.ltkcontent.com/files/US-State-Abbreviations.pdf

This leads to different albums for the same state in https://demo.photoprism.org/states.

The obvious solution is to expand abbreviations and always use the full name.

kvalev:
Quick question - do I need to run a full rescan for the new names to be applied? And are existing "duplicate" state albums going to be merged?

lastzero:
Please wait with updating your production instance. It's not done yet! Found a few issues while testing...

lastzero:
In particular, the problem also exists for other countries. Not just the US. Doing our best to also add a command for updating location data only, see GitHub Discussions. Was already answered there today.

kvalev:
> Please wait with updating your production instance. It's not done yet! Found a few issues while testing...

Okay will postpone it for now. I guess you can remove the `please-test` label just in case anyone else decides to give it a go as well.

kvalev:
> In particular, the problem also exists for other countries. Not just the US. Doing our best to also add a command for updating location data only, see GitHub Discussions. Was already answered there today.

Yes, I encountered the same problem with other countries as well, for example ones that do not use latin script - in my case arabic or cyrillic. Sometimes the state is in latin sometimes in arabic/cyrillic.

lastzero:
Well, we are testing right now.... `please-test` doesn't mean there are no issues anymore! :wink:

lastzero:
@kvalev You are welcome to send PRs with additional mappings :point_down: 

https://github.com/photoprism/photoprism/blob/develop/pkg/txt/states.go

lastzero:
Started a new [Development Preview](https://docs.photoprism.org/release-notes/) build for testing (wait until it is green and has been uploaded to Docker Hub):

https://docs.photoprism.app/getting-started/updates/#development-preview

lastzero:
Next [Development Preview](https://docs.photoprism.org/release-notes/) comes with a `photoprism places update` command! 💐 

https://docs.photoprism.app/getting-started/updates/#development-preview

It's brand new and pretty much untested outside my development environment. Please report issues if you find any. Since this will send many requests per second to our backend API, we need to have an eye on performance / server load. Therefore, at this time, it is only for sponsors who help us finance the infrastructure.

kvalev:
[...]

PR #1729, opened 2021-11-12
Places: Add Normalization Mappings for NDS (Germany) #1664

jloewe:
NDS is commonly used as an abbreviation for Niedersachsen.
See [Wikipedia](https://de.wikipedia.org/wiki/Niedersachsen) or in [the title of this Website of the local government](https://www.mi.niedersachsen.de/startseite/).


graciousgrey:
Thank you :)

PR #1743, opened 2021-11-21
Retrieve missing state data from Overpass

kvalev:
In case the places API does not provide state information for a given s2 cell, use the Overpass API to retrieve it.

Related to #1664


kvalev:
Not needed anymore

PR #1746, opened 2021-11-21
Auth: 401 on invalid credentials

Fylax:
With this PR changes some HTTP response codes from hardcoded values to net/http module constants, to achieve improved readability and more uniform APIs.

Moreover, on invalid credentials sent to session API, the HTTP response code changed from 400 (Bad Request) to 401 (Unauthorized).
The rationale of this change is described in [RFC 7235, section 3.1](https://datatracker.ietf.org/doc/html/rfc7235#section-3.1), in particular on the sentence
> If the request included authentication credentials, then the 401 response indicates that authorization has been refused for those credentials.


lastzero:
FYI: We can probably take care of this soon! :partying_face:

lastzero:
We have almost completely rewritten the session management with the above commit. The response codes should now reflect the actual status, e.g. whether you just don't have authorization or whether the session has expired and you need to re-authenticate.

lastzero:
I am closing this as the original PR has conflicts and is probably no longer needed, meaning it would not be merged. Thank you for bringing this up!

el-tiuri:
Hey, this is still on the roadmap in the “development” category. I’m guessing this should be removed from there?

lastzero:
We move it to "preview" as soon as a new build with the ":preview" tag has been released and documented. Due to the many changes and the fact that we are continuously developing new code, this may take a little while.

PR #1804, opened 2021-12-11
People: Improve update performance on MariaDB

srett:
MariaDB/MySQL traditionally have performance issues for queries of
type UPDATE ... WHERE xxx IN (SELECT ...)

Instead, use JOINs which are much faster.

As discussed in [this thread](https://github.com/photoprism/photoprism/discussions/1768#discussioncomment-1759544), I open this PR after verifying that all tests still pass in the docker compose dev environment, and I've been running these changes on my private collection and nothing blew up so far. This is the first time I touched Go code so I hope it's not too ugly. :-)

### DIFF
--- a/internal/entity/face.go
+++ b/internal/entity/face.go
@@ -273,11 +273,19 @@ func (m *Face) RefreshPhotos() error {
 		return fmt.Errorf("empty face id")
 	}
 
-	update := fmt.Sprintf(
-		"UPDATE photos SET checked_at = NULL WHERE id IN (SELECT f.photo_id FROM files f JOIN %s m ON m.file_uid = f.file_uid WHERE m.face_id = ?)",
-		Marker{}.TableName())
+	var err error
+	switch DbDialect() {
+	case MySQL:
+		update := fmt.Sprintf(`UPDATE photos p JOIN files f ON f.photo_id = p.id JOIN %s m ON m.file_uid = f.file_uid
+			SET p.checked_at = NULL WHERE m.face_id = ?`, Marker{}.TableName())
+		err = UnscopedDb().Exec(update, m.ID).Error
+	default:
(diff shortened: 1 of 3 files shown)


internal/entity/face.go
```diff
@@ -273,9 +273,14 @@ func (m *Face) RefreshPhotos() error {
...
-		"UPDATE photos SET checked_at = NULL WHERE id IN (SELECT f.photo_id FROM files f JOIN %s m ON m.file_uid = f.file_uid WHERE m.face_id = ?)",
-		Marker{}.TableName())
+	var update string
+	switch DbDialect() {
+	case MySQL:
+		update = "UPDATE photos p JOIN files f ON f.photo_id = p.id JOIN %s m ON m.file_uid = f.file_uid SET p.checked_at = NULL WHERE m.face_id = ?"
+	default:
+		update = "UPDATE photos SET checked_at = NULL WHERE id IN (SELECT f.photo_id FROM files f JOIN %s m ON m.file_uid = f.file_uid WHERE m.face_id = ?)"
+	}
+	update = fmt.Sprintf(update, Marker{}.TableName())
```
lastzero: May lead to problems in that dialects can have different parameters or need them in a different order. So it's much easier for developers to make mistakes or they have to refactor backwards.

srett: You're right, that's an unnecessary footgun. I'll pull the `.Exec()` call up into the branches.
Related: I was wondering why the `markers` table name is pulled from the according class via `.TableName()`, but the others aren't.

lastzero: Because we used temporary table names during development, while the other table names did not change after the initial stable release.

srett: Updated the PR. Tests still pass. Wasn't sure about the err var, could have put return statements into the branches, but I think it can be confusing if a non-void function doesn't end in a return statement.

lastzero:
Thanks a lot! I'll merge this so everyone can test it using the Development Preview :+1:

PR #1807, opened 2021-12-12
Add test for file name and extract date time.

matglas:
I implemented the extraction of date time from the example of `20130518_142022_3D657EBD.jpg`.

There was alredy a test case but it was 'disabled' because the case was not implemented yet.
See #1102 for the conversation on the topic. I had this problem while importing my own photos.

### DIFF
--- a/pkg/txt/time.go
+++ b/pkg/txt/time.go
@@ -9,6 +9,7 @@ import (
 var DateRegexp = regexp.MustCompile("\\D\\d{4}[\\-_]\\d{2}[\\-_]\\d{2,}")
 var DatePathRegexp = regexp.MustCompile("\\D\\d{4}/\\d{1,2}/?\\d*")
 var DateTimeRegexp = regexp.MustCompile("\\D\\d{4}[\\-_]\\d{2}[\\-_]\\d{2}.{1,4}\\d{2}\\D\\d{2}\\D\\d{2,}")
+var DateTimeCompactRegex = regexp.MustCompile("\\D([1-9][0-9]{3})(\\d{2})(\\d{2})[-_T](\\d{2})(\\d{2})(\\d{2})[-_]")
 var DateIntRegexp = regexp.MustCompile("\\d{1,4}")
 var YearRegexp = regexp.MustCompile("\\d{4,5}")
 var IsDateRegexp = regexp.MustCompile("\\d{4}[\\-_]?\\d{2}[\\-_]?\\d{2}")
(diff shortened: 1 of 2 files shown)


lastzero:
This is intentional because we do not want to extract time from canonical names that PhotoPrism generates itself:

It would cause a self-reference / feedback loop when re-indexing so that all "taken at" times are taken from the filenames and none from the file system.

matglas:
I see. But I am still a bit confused because does it mean that if there is no metadata the file date is preferred? Is that logical? Schould it not be the name that give more detail if there? And will it indeed create a loop? Where will that happen? Trying to understand and help if I can.

lastzero:
A specific timestamp in the file NAME has a higher priority than the file MODTIME, which is more volatile. Doesn't that make sense?

dexif:
> I implemented the extraction of date time from the example of `20130518_142022_3D657EBD.jpg`.
> 
> There was alredy a test case but it was 'disabled' because the case was not implemented yet. See #1102 for the conversation on the topic. I had this problem while importing my own photos.

That's what I need! Most of my photos (~15k) after import were undated. :(

lastzero:
What do you need? Why undated? PhotoPrism doesn't remove any dates.

dexif:
> What do you need? Why undated? PhotoPrism doesn't remove any dates.

Oh, sorry, it's my fault. :( I looked at the wrong files (no valid dates in filename and exif)

dexif:
But for example. I imported ~ 5k files from Google Photos and now they look like this:

Name | 2014/08/20140814_180759_8C8B6C78.jpg
Original Name | album/p8144388.jpg
...
Added | Dec 17, 2021, 2:40 PM in 365 ms

In this case, the date is not filled on details tab. I would like the date on details tab to be auto-filled on import for files in which these fields are not filled.

lastzero:
Where should the date come from? Does the file have metadata? The name doesn't contain a date as it seems.

dexif:
In fact, I don't know where photoprism got the date of the photo. However, the file name has the correct date and time. I suspect that the data is taken from JSON that is provided in the Google Photos archive. Unfortunately, I have no archive at the moment. To check if this is so, I will have to download gigabytes of archives again :)

graciousgrey:
@dexif PhotoPrism does read dates ("photoTakenTime") coming from google photos json and sets those as takenat date. To find out what's the issue with your file we need examples. Could you send us an example photo + related google photos json where this is not working? You can download both files from your originals directory on the files tab of the edit dialog.

dexif:
@graciousgrey Unfortunately my files are deleted after import. I have requested a new export to Google Photos, and as soon as it is ready, I will send you a JSON file and one photo.

lastzero:
Open a GitHub Discussion for this please 👍

lastzero:
Unfortunately, I have to close this. It would create a self-reference when re-indexing imported files, since it matches the automatically generated, canonical import file names. Thank you for your contribution! :heart:

PR #1873, opened 2022-01-02
XMP: Add support for reading fstop:favorite flag

akirk:
I use the app [F-Stop on Android](https://www.fstopapp.com/) to [tag photos as favorite](https://www.fstopapp.com/tutorial/) (with a heart symbol). Additionally, F-Stop can be [configured to create XMP side-car files for this added metadata](https://www.fstopapp.com/metadata/). It stores the favorite data in a custom `fstop:favorite` attribute.

This adds support for parsing this attribute and automatically marking a photo as favorite upon import.

This is my first contribution, I tried to get everything right but happy to follow up to get it merged.


internal/photoprism/index_mediafile.go
```diff
@@ -352,6 +352,9 @@ func (ind *Index) MediaFile(m *MediaFile, o IndexOptions, originalName string) (
 			details.SetSubject(metaData.Subject, entity.SrcXmp)
 			details.SetArtist(metaData.Artist, entity.SrcXmp)
 			details.SetCopyright(metaData.Copyright, entity.SrcXmp)
+
+			// Update externally marked as favorite.
+			photo.SetFavorite(metaData.Favorite)
```
lastzero: Won't this also automatically UNSELECT manually selected favorites?

internal/meta/data.go
```diff
@@ -20,6 +20,7 @@ type Data struct {
 	Title        string        `meta:"Title"`
 	Subject      string        `meta:"Subject,PersonInImage,ObjectName,HierarchicalSubject,CatalogSets"`
 	Keywords     Keywords      `meta:"Keywords"`
+	Favorite     bool          `meta:"FStopFavorite"`
```
lastzero: Does Exiftool support "FStopFavorite"? Otherwise the "meta" annotation isn't needed.

internal/photoprism/index_mediafile.go
akirk: You mean in the case when a photo has been made a favorite through PhotoPrism and is then re-indexed but without a side-car that marks it as favorite? It would be a similar problem for the reverse case, if a side-car file exists denoting it as favorite but it has been purposefully unselected from the UI.

Would a reasonably first step be to accept the latter for now by doing this?
```suggestion
			if metaData.Favorite {
				photo.SetFavorite(metaData.Favorite)
			}	
```

internal/meta/data.go
akirk: Actually it's `Favorite` there:
```
$ exiftool fstop-favorite.xmp
ExifTool Version Number         : 12.30
File Name                       : fstop-favorite.xmp
Directory                       : .
File Size                       : 336 bytes
File Modification Date/Time     : 2021:03:15 22:06:40+01:00
File Access Date/Time           : 2022:01:02 14:23:21+01:00
File Inode Change Date/Time     : 2022:01:02 12:52:08+01:00
File Permissions                : -rw-rw-rw-
File Type                       : XMP
File Type Extension             : xmp
MIME Type                       : application/rdf+xml
XMP Toolkit                     : XMP Core 4.4.0-Exiv2
Favorite                        : 1
```

So should I change it like this?
```suggestion
	Favorite     bool          `meta:"Favorite"`
```

lastzero:
@akirk My apologies for the long delay! Without conflicts and with your suggestions it looks like it can be merged :heart:

graciousgrey:
It should be noted that information from xmp files is always preferred over data from PhotoPrism. This means that if the favorite flag is set in the xmp file, it will also be set in PhotoPrism.

If you remove the favorite status in PhotoPrism, you must also remove the favorite flag from the xmp file, otherwise the photo will be marked as favorite again at the next complete rescan.

akirk:
@lastzero Thanks for the update. I've rebased the PR and added the suggestions.

@graciousgrey I agree but also I don't think that we can always be sure we're able to modify XMP files to sync this. I don't know the codebase well enough but is it possible to find out if an image was just added to the database and only then apply the favorite flag? In case of a full rescan the image would just be found and not added and thus the favorite flag could be ignored in order to avoid the re-favoriting that you described.

graciousgrey:
Long term we plan to support writing metadata changes from PhotoPrism back to Exif and XMP (optionally). Once this is implemented, the favorite flag would be removed from the xmp when the photo is marked as non-favorite in PhotoPrism. However, this feature will most likely not be available very soon.

>  In case of a full rescan the image would just be found and not added and thus the favorite flag could be ignored in order to avoid the re-favoriting that you described.

Some users intentionally edit xmp files, for example with other tools, of photos that have already been added to PhotoPrism and want that data to be read in during the next full rescan. Therefore, it would be problematic to ignore changes in xmp files if the photo has already been indexed before. 
For individual XMP fields this might make sense, but even there it is probably very dependent on the individual workflow.

akirk:
> Some users intentionally edit xmp files, for example with other tools, of photos that have already been added to PhotoPrism and want that data to be read in during the next full rescan. Therefore, it would be problematic to ignore changes in xmp files if the photo has already been indexed before.

Oh, this is not what I was trying to say. This was specifically about the favorite flag and setting from non-favorite in PhotoPrism to favorite. So the following behavior:
- New file, favorite in XMP → Added to PhotoPrism as favorite
- New file, not favorite in XMP → Added to PhotoPrism non-favorite
- Existing file, favorite in XMP, favorite in PhotoPrism → favorite in PhotoPrism (= no change)
- Existing file, non-favorite in XMP, favorite in PhotoPrism → non-favorite in PhotoPrism (= unfavorite)
- Existing file, non-favorite in XMP, non-favorite in PhotoPrism → non-favorite in PhotoPrism (= no change)
- Existing file, favorite in XMP, non-favorite in PhotoPrism → non-favorite in PhotoPrism (= no change !!)

graciousgrey:
That's what I meant by saying that it might make sense to have special logic for individual fields :)

However, it depends on the workflow, so there is no right or wrong way to do it.

In general I can see two use cases: 

A) The user sets the favorite status in xmp with another tool. He adds the photo to PhotoPrism and then only uses PhotoPrism to manage the favorite status of the photos.

This refers to:

> Existing file, favorite in XMP, non-favorite in PhotoPrism → non-favorite in PhotoPrism (= no change !!)

B) The user adds the photo to PhotoPrism. He then uses another tool to mark photos as favorites and wants this information to be displayed in PhotoPrism. 

This refers to

Existing file, favorite in XMP, non-favorite in PhotoPrism → Favorit in PhotoPrism 

instead of 

> Existing file, favorite in XMP, non-favorite in PhotoPrism → non-favorite in PhotoPrism (= no change !!)

Since this field seems to be an individual field created by the F-Stop app and is most likely not used by other apps, I assume that the use case in this case will only be A).  Since the app probably can't access the files once they are in the Originals folder, right?

akirk:
> That's what I meant by saying that it might make sense to have special logic for individual fields :)

Ok, right :)

> Since this field seems to be an individual field created by the F-Stop app and is most likely not used by other apps, I assume that the use case in this case will only be A). Since the app probably can't access the files once they are in the Originals folder, right?

I suppose it depends on how you sync your files. Something like Syncthing would sync updated XMP files from your phone.

A typical solution would be to introduce a setting but I could also see a possibility that if someone is capable of updating XMP files with a tool, maybe they could also set a second field that could indicate to PhotoPrism whether or not the changed setting in the XMP should overrule whatever is set in PhotoPrism.

Something like this:
```
<rdf:Description rdf:about=""
xmlns:photoprism="http://www.photoprism.app/xmp/"
photoprism:preferXmpMetadata="True"/>

```
vs the default:
```
<rdf:Description rdf:about=""
xmlns:photoprism="http://www.photoprism.app/xmp/"
photoprism:preferXmpMetadata="False"/>
```

This makes the flag per-file and allows "power users" to customize the behavior individually.

graciousgrey:
> I could also see a possibility that if someone is capable of updating XMP files with a tool, maybe they could also set a second field that could indicate to PhotoPrism whether or not the changed setting in the XMP should overrule whatever is set in PhotoPrism.

We appreciate your feedback!

If we add a setting, I think a setting in PhotoPrism that applies to all xmp files makes more sense. As far as I know, most people don't edit xmp files manually, but use tools (Digikam for example) that write to xmp. These tools don't write PhotoPrism-specific tags, and adding tags manually would make the workflow more complicated. Also, I assume that people's workflows in this regard are more or less the same for all their files. 
What do you think?

We'll think some more about possible settings and get back to you after our vacation.

lastzero:
@akirk I'm going to merge this PR and hope that it will make PhotoPrism more usable for you. We probably need to rework the implementation in a future release when we add full XMP support. Unfortunately, the current XML standard lib that comes with Go is not sophisticated enough for our needs, so we have to find an alternative first.

PR #1937, opened 2022-01-13
Videos: Exclude streaming from gzip compression

agramner:
Noticed that videos didn't start playing until fully downloaded when I used Firefox (OSX and Android). Started to investigate and found that the videos API path wasn't excluded from gzipping. When I turned off gzipping Firefox worked as expected.

This isn't a problem for Chrome and Safari because they don't seem to set `gzip` in the `Accept-Encoding` header for video resources but Firefox does.

The result is that, for Firefox, the server returns a gzipped stream and the browser seem to need to wait for all the data before start playing the video. And I guess gzipping compressed video cause unnecessary CPU usage as well.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -64,6 +64,7 @@ func Start(ctx context.Context, conf *config.Config) {
 				conf.BaseUri(config.ApiUri + "/zip"),
 				conf.BaseUri(config.ApiUri + "/albums"),
 				conf.BaseUri(config.ApiUri + "/labels"),
+				conf.BaseUri(config.ApiUri + "/videos"),
 			})))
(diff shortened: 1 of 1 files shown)


graciousgrey:
Thank you for your contribution!

- Docs need to be updated

Reviewing, testing and finally merging pull requests consumes significant resources on our side. Unless it's just a small fix, it may take several months.

Thanks for your patience :)

agramner:
Thanks for the great work!

Consider this pull request as a way of reporting my findings. You're welcome to close it and, if you wish, implement the changes in any other way 🙂. No need for me to sign anything for 1 line of code.

PR #1948, opened 2022-01-15
Contributing: add link to CLA

alexislefebvre:


lastzero:
Thanks a lot! I'm adding this directly as there seems to be a merge conflict.... should explain the background as well, so the question doesn't pop up or only when there is an actual problem, e.g. because somebody is working for a company than doesn't allow signing other CLAs (only their own...).

alexislefebvre:
You already added it with great explanation so this PR can be ignored. See https://github.com/photoprism/photoprism/commit/60924853fc38416c3579931edb389ad689f92cfe

lastzero:
Thank you! When CLAs came up, I was also very skeptical based on what I had heard "on the internet". For BSD and Apache-like licenses, a CLA may not be as important as for copyleft licenses.

Just learned that even the Linux Foundation now requires a Developer Certificate of Origin (DCO). Seems it all started with the [SCO–Linux disputes](https://en.wikipedia.org/wiki/SCO%E2%80%93Linux_disputes).

PR #1964, opened 2022-01-22
Docker: Fix tensorflow Dockerfile

srett:
There are no more *.diff files as of 4f05fde794c so the COPY
command makes building the container fail. The deleted diff
file was for TF 1.14.x which has long been dropped.


lastzero:
Thanks! Hopefully we don't need diff patches anymore....

PR #2007, opened 2022-02-04
Remove unused face detector property

kvalev:


lastzero:
Thanks! Looks like this variable is now consumed directly in `thresholds.go`:

```go
// QualityThreshold returns the scale adjusted quality score threshold.
func QualityThreshold(scale int) (score float32) {
	score = float32(ScoreThreshold)
```

PR #2152, opened 2022-03-14
Albums: Preload more items for auto-completion in "Add to album" dialog

ppprism:
This is a small hotfix to the "Add photos to album" dialog.

If there are more than 1000 albums in photoprism, a problem arises in the user interface:
- you cannot add photos to existing albums that are not included in the range
- worse, it looks like there is no such album yet, leading to duplicate albums with same name when pressing add button

This pull requests increases the max album fetch `count` to 10000 as quickfix and removes the magic number, declaring `MAX_COUNT` as explicit constant.

Future improvements of REST API may one of the following:
- include an "infinite count" flag, as the intention is always to provide *all* albums in autocomplete dialog
- make `count` query parameter optional

Background:
I have a bunch of auto-generated albums (> 1000) from external sources and found it very confusing, that only one half of albums have been displayed in the autocomplete.

---

Acceptance Criteria:

This is a small change (projects with < 1000 albums not affected in any way), hence following points are neglected:

### DIFF
--- a/frontend/src/dialog/photo/album.vue
+++ b/frontend/src/dialog/photo/album.vue
@@ -105,9 +105,13 @@ export default {
 
       this.loading = true;
 
+      // todo: either introduce infinite flag or
+      // make count parameter optional for REST API
+      const MAX_COUNT = 10000;
+
       const params = {
         q: q,
-        count: 1000,
+        count: MAX_COUNT,
         offset: 0,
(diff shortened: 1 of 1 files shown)


lastzero:
Thanks! Apparently we didn't expect private users to have more than 1000 albums when implementing this...

Issue #2258, opened 2022-04-16
UX: Implementation/improvement ideas for infinite scrolling


heikomat:
**Is your feature request related to a problem? Please describe.**

When scrolling through multiple month or years worth of pictures, the application gets slower and slower over time while consuming ever increasing amounts of ram, until it gets unusably slow (opening pictures takes >5 seconds)

I'm pretty sure that the culprit is the missing virtualization in the picture grid. The browser has to manage tens of thousands of picture nodes, likely adding up to hundrets of thousands of dom nodes, even though only a couple dozen are visible at a time

**Describe alternatives you've considered**

A workaround is to filter the pictures by month. Looking at the picutres of a single month, then scrolling back to the top, change the month and look at the next pictures.

**Additional context**

[this comment](https://github.com/photoprism/photoprism/issues/500#issuecomment-727917142) already talks about virtualization, but the corresponding issue was ultimately closed, as it was about timeline-view, and virtualization might have only been a part of that discussion.

---

**Suggestion for implementation**

I know that the relative positioning that is currently being used makes the UI easy to implement, but is not really compatible with conventional virtualization.

Not long ago i faced a similar problem where non-absolute-positioned elements had to be virtualized.
The solution was to use the [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver) so that every element could just get informed by the browser if it is visible or not, instead of implementing some logic to calculate that.
If an element is not visible, it just renders a single div of the same dimensions as the original as placeholder.
Not quite real virtualization, as it still requires the placeholders, but should be much, much faster when scrolling through large lists

Using an IntersectionObserver
- is easy to implement (because no virtualization-logic is required)
- is fast (because the browser has that info anyway)
- is [supported by all major browsers](https://caniuse.com/?search=IntersectionObserver)
- works with relative positioning and deeply nested and/or complex structures
- natively supports things like [rootMargin](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/rootMargin) that can be used to render a couple more lines than actually visible (reduces flickering while scrolling)

**Some hints**
- As is demonstrated [here](https://www.bennadel.com/blog/3954-intersectionobserver-api-performance-many-vs-shared-in-angular-11-0-5.htm), one can use a single IntersectionObserver to observe as many elements as required, increasing performance compared to using multiple intersection-observers
[...]

Issue #1623, opened 2021-10-13
UX: Memory consumption warning in Safari


skavans:
Hey folks,

Safari shows "This Webpage is using Significant Memory in Safari" message every time I use the PhotoPrism. And forcibly reloads the page after time.

It just a little bit depends on how far I'm scrolling the feed. Sometimes I can just simply navigate the sidebar items several times and the message appears.

I don't know if it's possible to beat this somehow though.

<img width="1016" alt="Снимок экрана 2021-10-13 в 22 57 13" src="https://user-images.githubusercontent.com/6472406/137204600-e1eb11aa-0f49-4e6a-ba12-2da07572041a.png">

lastzero:
Safari is right in that images and videos consume a lot of memory, in the browser and in the backend as well. We'll optimize this in future releases. It's also Apple's responsibility to reduce memory usage. HTML is a descriptive language, we don't tell Apple how to manage the memory. Won't come up with a solution by tomorrow. Is it possible to disable the warning or to use a different browser?

skavans:
> Is it possible to disable the warning or to use a different browser?

Sure, not a huge problem for me. Just wanted to let you know.

lastzero:
Thanks! 🌈

graciousgrey:
Released since no problems were reported!

Issue #1582, opened 2021-10-01
UX: Show hint/notice when infinite scroll is loading


robbymilo:
When infinite scrolling:
- There is no visual indication that more results may be loaded in the near future.
- There is no visual indication that all results have loaded.

Potential suggestions:

Show `Loading...` (maybe a spinning icon?) when the XHR request has been initiated:
![loading](https://user-images.githubusercontent.com/8106669/135686685-f3f144a6-1e61-405d-b5cc-a135aa182617.png)

Show how many results have been loaded, as well as total results:
![total](https://user-images.githubusercontent.com/8106669/135686765-5c8bf840-e837-40a7-b6a0-e4f8d2c9e049.png)

Show that you've reached the end of scrolling:
![the-end](https://user-images.githubusercontent.com/8106669/135687070-7665392a-6267-4a7b-88cb-9286e9a1c7c7.png)

lastzero:
There should be a top loading bar as well as notifications at the bottom.

robbymilo:
I see the top loading bar, but it only appears after the xhr request has finished loading, and after the images have already started downloading, so it doesn't give the user a hint that more images could be loaded soon.

I also see the "More than 50 results" notice, but that is only on initial load.

lastzero:
That's common as a response time of 100ms is perceived as instantaneous. Google it. Most loading bars work like this.

tribut:
I quite often miss the loading bar at the top and I've recently observed my significant other having the same problem. I'm staring at the bottom of a list. Looking at the top of my browser is just not very natural in that moment. I would appreciate a loading indicator at the bottom, where the new images are supposed to appear.

lastzero:
On top of the notifications that are also displayed at the bottom? If all rows are filled, there is no space otherwise. Showing a loading bar at the top is common, kind of a best practice. We can of course show it earlier if it helps.

tribut:
Are we talking about the same thing? I'm talking about what happens when I'm scrolling down a long list and then hit the "bottom" (which then triggers some sort of "load more" event). I don't think there is any notification at the bottom.

lastzero:
There can be notifications as it's the general notification area, for example when other users trigger actions such as creating albums or changing settings. Try for yourself, for example using our demo. It's also possible to initial result notification ("more than 50 pictures") is still visible when you scroll very fast.

tribut:
[...]

Issue #1438, opened 2021-07-19
Search: Performance Optimizations


graciousgrey:
**As a user I want to have the best performance possible, so that I enjoy scrolling through my library.**

This ticket is meant to collect possible performance optimizations:

- #1413 
- #1105 
- #1481

Dulanic:
Just adding some recommendations I am seeing for commonly used queries. Will edit and add more as I find them.

1. Change this where to a in, OR is horrible for performance. `(files.file_type = 'jpg' OR files.file_video = 1)`  should be for example...  `file_type IN ('jpg','mov','mp4','mts','wmv')`
2.  Get rid of offset etc... maybe a subquery would be better?  Or pull the list of Id's and store it in the program and pull just those 60 Id's you want instead of all and then pulling back just 60 from that? Or you could use a cursor, but I'm not sure how well SQLLite works with cursors. EX:  You could pull back the Id's in batches of 1000 store the Ids in memory, that would allow people to scroll up to 1000 before you needed to request more ids? I'd say use subqueries, but Maria DB doesn't like order by/limit with subqueries yet. 

You load 1000 photos to memory /w a simple query like so:
```
SELECT id, taken_at
FROM photos
order by taken_at
LIMIT 1000;
```
Then your query gets an additional where...  which would run 10x faster than the current queries. 
`AND photos.id IN (1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14)`

nautatva:
User needs to wait for a couple of seconds before the archive button tick archives the photo. What I think can be done is showing the change as done on front-end and roll it back to view incase of error. This change can be applied for all the operations (Face release is one other I see takes time).

lastzero:
One of the issues here is that many caches need to be invalidated as well as thumbnails / covers replaced / checked. Also every user should see the same and not one user the change as done and others not. It's a pretty complex problem we are happy to further optimize when we reach our funding goal and had time to work on the items already on our roadmap.

lastzero:
The new development preview is up to 5x faster than the previous version:

![Screenshot 2022-03-31 at 10 54 52](https://user-images.githubusercontent.com/301686/161017862-d967c841-603c-4b81-adc4-369196eccf22.png)

rdevries20:
Hi,

[...]

PR #2292, opened 2022-04-29
UI: Pseudo-virtualize mosaic, cards and list view

heikomat:
This implements pseudo-virtualization for the mosaic, cards and list-view using the [IntersectionObserver-API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API).

It also memoizes some photo-model-functions that are called thousands of times during rendering of the aforementioned views. These memoizations result in expensive calls like new Dates and array-filters only getting executed if actually necessary.

With these fixes, scrolling in the mentioned views is much faster and takes way less memory, because the browser renders near-empty placeholders instead of the full elements for elements that are not in view.
The memory-usage after scrolling through ~3600 images on any of these views is reduced from about 2.0-2.5 GB to about 650-850MB.

While this type of pseudo-virtualization is easy to implement without requiring much refactoring, it requires some rules to be followed:
1. The size of the placeholder-elements should be as close to the real element as possible. That's why i still render texts, because long texts could increase the size of the element
2. The placeholder-Elements **must** only consist of regular HTML-Elements, it **must not** contain vue-components. This makes sure that vue doesn't need to rerender components in order to check if the dom needs to beupdated, which in turn drastically increases performance. Rule of thumb: if a component appears in the vue-dev-tools, it should not be used for placeholder-elements

This PR adresses the issues mentioned in https://github.com/photoprism/photoprism/issues/2258

Acceptance Criteria:


heikomat:
> [CLA assistant check](https://cla-assistant.io/photoprism/photoprism?pullRequest=2292) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.io/photoprism/photoprism?pullRequest=2292) before we can accept your contribution.**1** out of **2** committers have signed the CLA.✅ heikomat❌ Heiko Mathes
> **Heiko Mathes** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/photoprism/photoprism?pullRequest=2292) it.

Most of the commits here were made on my desktop-pc, where git is not fully configured, hence it uses the non-existent email heiko@pop-os.localdomain.
I therefore can't sign the CLA for that account, as it can't exist.

Only the last few commits were made on my laptop, where git is fully configured. For these commits, the account was correctly associated, so i could sign the cla for that.

I hope that is not a problem. If it is, could steal my own commits using a correctly configured git and open a new PR, but i'd prefer not to do that.

lastzero:
We'll replace the CLA Assistant anyway since it doesn't work with our translation tool . For now, you can print https://photoprism.app/cla and send a signed scan/copy via email. Note the signature field is not visible on a screen, only when printing it becomes visible at the bottom (A4 paper format is best). Implemented this today as a temporary solution until we had time to set up a self hosted app for signing.

heikomat:
> We'll replace the CLA Assistant anyway since it doesn't work with our translation tool . For now, you can print https://photoprism.app/cla and send a signed scan/copy via email. Note the signature field is not visible on a screen, only when printing it becomes visible at the bottom (A4 paper format is best). Implemented this today as a temporary solution until we had time to set up a self hosted app for signing.

Just sent a mail with a signed cla to hello@photoprism.app. I wrote `heiko@pop-os.localdomain` as "github username", as that is the "mail" attributed to the commits

lastzero:
Thanks! We will get back to you as soon as possible. Need to review a few bug reports first. Looking forward to get this integrated.

heikomat:
When testing on a smartphone i observed a couple of things:
1. The changes have the intended effect. Scrolling down no longer causes the app to get slower and slower
3. When scrolling slowly to medium fast (on any of the 3 views), i bearly ever saw a placeholder, as the pictures were rendered fast enough
4. When scrolling fast on the cards and list view, the placeholders are visible, but only for a couple milliseconds, so no big deal
5. **BUT** when scrolling medium to fast in the mosaic view, placeholders are visible for about 1 seconds, maybe even 1.5 seconds. That is definitely noticable.

It seems that there are just to many view-elements to render at once on a single screen.

When looking at the rendered vue-elements i noticed that there are a lot of Buttons with Icons on every picture, that are rendered but invisible.

- When nothing is selected, 1 Button with an icon is visible on each picture.
- When something is selected, 3 Buttons with icons are visible on each picture.

However, no matter how many are visible, each photo in the mosiac view renders **6 Buttons with a total of 13 icons at all times**. If for example 18 pictures are on screen, and 27 are rendered, that means 126 Buttons with 351 Icons are rendered, even though only 27-81 of them are ever visible.

Then i thought, we might be able to cut the render time by 80-90% by just not rendering the buttons that aren't visible, but i hesitated after trying to find out why they are always rendered.

@lastzero in https://github.com/photoprism/photoprism/commit/8869e5b995e7169ccfc45eb91108e78dbc9431fe you changed the logic from rendering these buttons conditionally to rendering them at all times.
The commit message says that reduces lag when selecting images.
Could it be, that the coulprit of that lag was the fact that when selecting an image, the whole view had to be rerendered, which was slow because it wasn't virtualized?

If so, we might be able to go **back** to conditional rendering, which might actually speed up the mosaic view significantly

lastzero:
Rendering can be simplified and probably accelerated a lot by using CSS instead of if/else in templates. Just a matter of having the time to do it without breaking existing functionality for anyone, so lots of testing incl updating the automated tests we have.

heikomat:
> Rendering can be simplified and probably accelerated a lot by using CSS instead of if/else in templates. Just a matter of having the time to do it without breaking existing functionality for anyone, so lots of testing incl updating the automated tests we have.

I understand. Considering this, it might be best to tackle mosaic-element-renderperformance as a separate issue, either completely independent of this PR or as a followup, but not as a blocker.

I will definitely try a couple things out here and there anyway (on a new branch), just because i'm kinda curious now 😄

EDIT:
A quick tests shows, that indeed, most of the rendering time is spent rendering invisible buttons and icons. rendering them conditionally with templates speeds things up significantly, and makes the virtualized mosaic view way more usable when flicking through images

heikomat:
@lastzero Just finished a small PR (only ~~45 lines~~ (update: 67 lines) ignoring whitespace-changes, [here](https://github.com/heikomat/photoprism/pull/1)) that is based on this one. It just implements rendering the Buttons and Icons conditionally

The Result:
- Time spent executing JS is cut by ~~50%~~ (update: 75%)
  - pictures appear noticably faster
  - batch-loads finishing much quicker (because the cpu spends less time rendering and more time processing the requests)
- The original reason for not doing this seems to be obsolete with virtualization

I know i already claimed a lot of your time with all my talking and ideas and so on, but i'd really appreciate if you could read the PRs description. It includes flamegraphs and a video!

I really think that that improvement (if tested enough) can be the barely-invasive boost that this virtualization needs to go from "scrolling doesn't slow down anymore, but pics take a second to load" to "damn, i can scroll through thousands of pictures like it's nothing!"

Rolf-Smit:
This would be an amazing improvement. I was myself already looking into 'native' apps (Flutter), but the API is not ready yet. Because I do see a lot of performance issues when scrolling on especially the more budget friendly smartphones.

heikomat:
I'd love have the different photo-views (mosaic, list and cards) deduplicated, so that both of them benefit from the virtualization without having to implement it twice. These views exist once for the search, and once for shared albums.

This would also reduce the chance of there being differences for views that are probably supposed to be identical to the end-user, as they would then actually be the same component.

@lastzero said, that duplicating the views was the easiest and fastest way to implement the shared views, so that was the way to go, considering time is at a premium.

So i went ahead and deduplicated the views in [this PR](https://github.com/heikomat/photoprism/pull/2) (wich is based on this virtualization PR)
I would have just added the changes to this PR here, but i'd prefer to have the confirmation of @lastzero that deduplicating these views is compatible with the way he wants things to be. (And while it's fully implemented, i have yet to do a last check to see if i missed changing anything, like translations, and run the tests)

heikomat:
**UPDATE** to the followup-PR i mentioned in https://github.com/photoprism/photoprism/pull/2292#issuecomment-1120333528:

I managed to about half the render-time again by replacing `<v-card>` and `<v-img>` in these views with regular. `<div>`s resulting in just more of the same benefits! (while only adding 22 lines of changes)

With only the bare virtualization from this PR (aka without optimizing the render-performance of the photo-views), a screen full of images took about 1 second to render, of which `680ms` are used to execute JavaScript

With the optimizations from https://github.com/heikomat/photoprism/pull/1 the same number of pictures renders in ~1/3 of a seconds, of which only `120ms` are used to execute JavaScript

heikomat:
Updated https://github.com/heikomat/photoprism/pull/1 again. This time replacing `<v-btn>` with `<button>` and `<v-icon>` with `<i>`. saves another 30ms of JS in regular (non-selection mode) mosaic view. Probably more when more buttons/icons are displayed.

I think this is now at a point of diminishing returns, but that is ok because infinite scrolling is now plenty fast!
Just look how fast i can scroll and still have images visible!


https://user-images.githubusercontent.com/20770029/169658374-d0ff03db-bc99-434f-9570-0c5983c70d7d.mp4


Will now walk the dog. After that i'll add what i learned to the docs and do some more testing on more browsers (including smartphone) (i love free saturdays!)

@lastzero what would be the next steps to take here? What can i do to help make this ready for production?

heikomat:
[Added some docs](https://github.com/photoprism/photoprism-docs/pull/115). Have yet to proofread them though

heikomat:
For testing i just pushed a docker image based on https://github.com/heikomat/photoprism/pull/1: [`ghcr.io/heikomat/photoprism:preview`](https://github.com/heikomat/photoprism/pkgs/container/photoprism/24926641?tag=preview)

### USE AT YOUR OWN RISK!
This comes with 0 guarantees for anything and is by no means anything official from photoprism.

That being said, it should work as a drop-in-replacement for the `photoprism/photoprism:latest` or `photoprism/photoprism:preview` images

Rolf-Smit:
@heikomat I just want to say that I tried your Docker image and so far I have not found any issues.

The performance is so drastically improved. I use a relatively slow machine (Quad code, 8GB, Celeron J4125) and I have a large library of almost 100.000 pictures. Usually it takes a while for the next bunch of photos to load, scrolling then obviously stops. But even worse scrolling back up would sometimes even be extremely laggy.

When scrolling through the search now, loading is obviously still the same, but the overal experience is so much more usable. No more lagging. It even feels like the loading of batches itself has improved (but that is not necessary true, right?).

Anyway this testing was all done on mobile, this weekend I will probably also try it on a non-mobile device as well.

Rolf-Smit:
So last week I did some more testing on non-mobile devices, I found no big issues, but did observe a few things worth sharing although I'm pretty sure these are not related to the changes made here:

- For some reason I switched to `list` mode, but I'm unable to make PhotoPrism remember that I switched back to `mosaic` or `cards`, for some reason after refreshing the page it always goes back to `list` mode. This happens at least on the search view. I doubt it has anything to do with this PR though.
- The `mosaic` view works pretty nice on mobile, the amount of photo's that fit on mobile screens is pretty small so you have plenty of room for scrolling for each batch that is loaded. But on a computer monitor each batch only just fills the screen, this makes for a kinda weird scrolling experience at first. I don't think this is a big issue, but it got me thinking, maybe the `mosaic` cards should be slightly larger on bigger monitors to begin with? I now get 12 columns, the photo's are so small that I sometimes feel the need to open them to really check what I'm looking for. The `cards` view shows me 6 columns which is rather big, but fine with the extra info attached. Maybe `mosaic` should be more in the middle? 9 columns?
- I had a feeling that the view that is shown when you share a link to an album doesn't really seem to benefit from the changes in this PR, it felt slower then search for example. Maybe nothing has changed there, and this observation makes sense?

lastzero:
**Thanks for testing!**

We really appreciate it as it helps to gain confidence that these improvements are ready to be merged 👍 
 
> * For some reason I switched to `list` mode, but I'm unable to make PhotoPrism remember that I switched back to `mosaic` or `cards`, for some reason after refreshing the page it always goes back to `list` mode. This happens at least on the search view. I doubt it has anything to do with this PR though.

The last used view type is stored in [window.localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage):

- either you're using a view (such as an album sharing link) that doesn't support it
- or there are other tabs in the same browser open that are overriding it
- or you have a privacy enhancing browser plugin installed / use the privacy mode
- or it broke accidentally while implementing these changes

> * Maybe `mosaic` should be more in the middle? 9 columns?

I would save that for later, because it's better not to mix functional and non-functional changes, and generally not to change too many things at once. It might be worth asking more users, e.g. through a poll.

Also note that keeping the size the same, but loading and rendering more tiles at once so you can scroll more before a new batch needs to be retrieved, has not proven ideal so far:

- The number of parallel HTTP connections turned out to be a bottleneck that could block other, potentially more important requests (browsers generally have a fixed limit and then start blocking)
- This puts more load on the server by default, which is especially problematic for people who want to render thumbs entirely on demand or simply have weak hardware

**Edit: Commercial cloud photo hosting services get around this limitation by using a CDN with different subdomains so that requests to the API are not blocked by loading images. We don't have that luxury, unless we complicate the installation even more.**

heikomat:
> Also note that keeping the size the same, but loading and rendering more tiles at once so you can scroll more before a new batch needs to be retrieved, has not proven ideal so far:
> 
> * The number of parallel HTTP connections turned out to be a bottleneck that could block other, potentially more important requests (browsers generally have a fixed limit and then start blocking)

@lastzero Keep in mind that with the pseudo-virtualization, images that are too far outside the view are not loaded.
On the performance-branch that the self-build image is based on (https://github.com/heikomat/photoprism/pull/1) i increased the distance that is used to load the next batch from 2 x window-height to 6 x window-height, so you don't bump into the last-loaded-image-edge as often

With virtualization this loads the batches, but not the images. The images are only loaded when they come into view

heikomat:
> * I had a feeling that the view that is shown when you share a link to an album doesn't really seem to benefit from the changes in this PR, it felt slower then search for example. Maybe nothing has changed there, and this observation makes sense?

That makes sense, because the shared albums use components that are 99% identical to the ones from search, but are in fact not the same components.

Because they are so very similar i plan on making the shared album views use the actual same components as a followup to these virtualization and performance improvements.

I already got a PR for that, that implements just this. Will update that PR once the virtualization and performance fixes are merged. See https://github.com/heikomat/photoprism/pull/2

lastzero:
> With virtualization this loads the batches, but not the images. The images are only loaded when they come into view

That's why I said "so far", because your optimizations might change this and allow working with larger batches, at least optionally or on large screens.

In order to merge your PR as soon as possible, I would nevertheless recommend to perform any optional optimizations and modifications only after it has been merged.

Rolf-Smit:
Since I feel like these are important changes I try my best to help where I can and keep testing.

I found one more thing, again I'm not sure if this is related: the favorite? (⭐) icon is shown here both selected and unselected.
Screenshot_20220616-104747

@heikomat is your Docker image being updated with the more recent changes you made?

heikomat:
> Since I feel like these are important changes I try my best to help where I can and keep testing.
> 
> I found one more thing, again I'm not sure if this is related: the favorite? (⭐) icon is shown here both selected and unselected.
> Screenshot_20220616-104747
> 
> @heikomat is your Docker image being updated with the more recent changes you made?
> 

This has been fixed two days ago.
If you pull the image again, you should have an updated version where that is fixed. It also contains additional minor performance improvements and the latest upstream-changes

lastzero:
Related / Follow-Up Issues:

- #85
- #152
- #307
- #583
- #1582
- #1623

The ultimate outcome for a few of them, i.e. better UX when searching and scrolling, could be achieved with these improvements. Needs to be discussed.

dror3go:
I tested this on Firefox under Linux and under Android - the performance gains are really impressive and I didn't notice any issue.
Great work! :muscle:

rmmanseau:
Tested on Brave browser and the performance gains are really good. The page used to choke if I would scroll deep enough into the search tab but now it doesn't appear to be a problem at all.

One thing I noticed, If i scroll far enough eventually I get a yellow popup that says "Can't load more, limit reached". This limit seems arbitrary now that the performance has improved so much, it would be nice if I could just keep scrolling.

Thanks for all the effort put into this!!

heikomat:
> Tested on Brave browser and the performance gains are really good. The page used to choke if I would scroll deep enough into the search tab but now it doesn't appear to be a problem at all.
> 
> One thing I noticed, If i scroll far enough eventually I get a yellow popup that says "Can't load more, limit reached". This limit seems arbitrary now that the performance has improved so much, it would be nice if I could just keep scrolling.
> 
> Thanks for all the effort put into this!! 

The limit used to be at 3'333 results.
It was increased to 10'000 results about one or two weeks ago

lastzero:
Also note that the full-screen viewer has recently been optimized to initialize only with the required data, which is already in the right format. However, it will load all the results from the top and additional results further down, so you can click "Next" for a while or start a slideshow:

- #1438

That was good enough with the old search results views due to other bottlenecks that now seem to be mostly resolved. We will provide a new viewer, but not in the very short term:

- #1307
 
In the worst case, we'll have to implement another workaround, such as using an offset when loading viewer content.

rmmanseau:
good to hear the limit has been increased, but im curious why it wasn't removed / made optional? I would imagine that infinite scrolling is possible now after these changes. is that still not the case?

lastzero:
- As mentioned above, there are other considerations like the photo viewer and the fact that the same restriction currently applies to non-optimized views. It was a quick change I implemented for testing purposes to see if it even worked that way, so I didn't have time to refactor everything.
- For a timeline view, you still need a timeline, otherwise you don't know where you're going to end up with the scrolling.
- Meaning, from a user perspective, totally infinite scrolling might work if you don't mind being "blind" and using an API implementation that loads way too much data because it's not designed to load only the last results. However, this is not something I would promote or encourage users to do.
- In practice, it seems much more efficient to change the sort order to show the oldest results first than to scroll down for a very long time only to find that after 10,000 entries, performance is still not as good as expected.
- We should focus our testing efforts on "standard" use cases for now and make sure no bugs have been introduced and it works on all browsers before moving on to more extreme things, tempting as that may be.

PR #2371, opened 2022-05-28
Docker: Don't update permissions of originals on startup

BonzTM:
Separating Originals path from Storage and Imports path for CHOWN purposes.  This will allow users with specific ownership on Originals to retain those owners while allowing the rest of the pathing to be owned by the UID/GID in the container.

Photoprism states that it does NOT modify the Originals directory.  This is untrue.  A power user who controls the permissions of the Originals directory should be able to exclude it from being owned by container UID/GID.

Acceptance Criteria:

### DIFF
--- a/scripts/dist/entrypoint-init.sh
+++ b/scripts/dist/entrypoint-init.sh
@@ -17,15 +17,18 @@ case $DOCKER_ENV in
   prod)
     export PATH="/usr/local/sbin:/usr/sbin:/sbin:/bin:/usr/local/bin:/usr/bin:/scripts:/opt/photoprism/bin";
     INIT_SCRIPTS="/scripts"
-    CHOWN_DIRS=("/photoprism" "/opt/photoprism")
+    CHOWN_DIRS=("/photoprism/storage" "/photoprism/import" "/opt/photoprism")
     CHMOD_DIRS=("/opt/photoprism")
+    CHOWN_ORIGINALS=("/photoprism/originals")
     ;;
 
   develop)
     export PATH="/usr/local/sbin:/usr/sbin:/sbin:/bin:/usr/local/bin:/usr/bin:/scripts:/usr/local/go/bin:/go/bin:/opt/photoprism/bin";
     INIT_SCRIPTS="/go/src/github.com/photoprism/photoprism/scripts/dist"
-    CHOWN_DIRS=("/photoprism" "/opt/photoprism" "/go" "/tmp/photoprism")
+    CHOWN_DIRS=("/photoprism/storage" "/photoprism/import" "/opt/photoprism" "/go" "/tmp/photoprism")
     CHMOD_DIRS=("/opt/photoprism" "/tmp/photoprism")
+    CHOWN_ORIGINALS=("/photoprism/originals")
+
     ;;
 
   *)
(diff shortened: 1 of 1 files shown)


alexislefebvre:
You can mount any folder in read-only mode to restrict the access to files:

```yaml
   volumes:
      - "/mnt/photos/nature:/photoprism/originals/nature:ro"
```

BonzTM:
> You can mount any folder in read-only mode to restrict the access to files:
> 
> ```yaml
>    volumes:
>       - "/mnt/photos/nature:/photoprism/originals/nature:ro"
> ```

Very true.  You can do many different things.  

I only opened the PR because the current default behavior goes against the "we do not ever touch originals" wording.  I do understand from a non-paying-user support standpoint things are kept as simplistic as possible, but there should still be options.  We don't always know the tech stack that others are using, and maybe they may not have the ability to mark something read-only.  Regardless, being explicit is almost always a better thing than being implicit.

I have an OKD cluster w/ CephFS backing my `originals` directory.  I have had to mark this path readOnly, even though I may have other paths along this one that I do not wish to be marked as readOnly.

So my manifest looks like such:
```yaml
  spec:
    volumes:
      - name: originals
        cephfs:
          path: "/path/to/originals"
          readOnly: true
          monitors:
            - 10.1.2.30:6789
            - 10.1.2.31:6789
            - 10.1.2.32:6789
          user: okd-photos-user
          secretRef:
            name: password
```
            
My problem is solved, though I do not entirely like the implementation.  You may close this PR if you deem it harmful.

insomnia417:
This problem really bothered me a hundred times, either adding different user management permissions, or restricting the permission modification of my origins file by photoprism (sorry, my English is poor, from translation software)

lastzero:
Started a new preview build for testing! :+1:

BonzTM:
@lastzero thanks for the updates!  There is 1 issue that has surfaced later than your commits regarding 

```sh
switching to uid 1000620000:1000620000
/opt/photoprism/bin/photoprism start
setpriv: uid 1000620000 not found, --init-groups requires an user that can be found on the system
```

but I can open a separate issue for that and my issue and PR can be closed

lastzero:
UID 1000620000 is not supported, see list of valid ranges in the config examples.

BonzTM:
> UID 1000620000 is not supported, see list of valid ranges in the config examples.

Unfortunately this requires me to bypass certain security contexts in OpenShift/OKD, but I'll deal with it.

hiro5id:
Is this a configurable ?  For example, can I turn this behaviour back on if I want to ?

EDIT: Oh I see... I can set `PHOTOPRISM_DISABLE_CHOWN_ORIGINALS` environment variable to control this behaviour. 👍

BonzTM:
> Is this a configurable ? For example, can I turn this behaviour back on if I want to ?
> 
> EDIT: Oh I see... I can set `PHOTOPRISM_DISABLE_CHOWN_ORIGINALS` environment variable to control this behaviour. 👍

`PHOTOPRISM_DISABLE_CHOWN_ORIGINALS` was not implemented.  Instead, @lastzero changed the original argument from taking ownership of the originals directory in the following commit: https://github.com/photoprism/photoprism/commit/f06d768e78298763246ec853c019da4bc2960e50

hiro5id:
OOff..... so.... its not configurable then?

lastzero:
The originals folder itself is not touched anymore, but the behavior for the storage folder can be enabled / disabled as before.

PR #2414, opened 2022-06-12
Metadata: Parse IMG/VID-YYYYMMDD filename format used by WhatsApp

Bur0k:
With this PR parsing of `YYYYMMDD` in parts of the filenames is implemented.
This was made in order to work with the WhatsApp filename format and fixes/implements #1102

If needed, the regex could be alter so it would only match the WhatsApp filename format.
Depends if one would like this general case (See the tests for a behavior change in other cases)

Acceptance Criteria:


pkg/txt/datetime_filepath_test.go
```diff
@@ -61,13 +61,13 @@ func TestDateFromFilePath(t *testing.T) {
...
-		//assert.False(t, result.IsZero())
-		assert.True(t, result.IsZero())
+		assert.False(t, result.IsZero())
+		assert.Equal(t, "2013-05-18 00:00:00 +0000 UTC", result.String())
 	})
 	t.Run("20130518_142022_3D657EBD.jpg", func(t *testing.T) {
 		result := DateFromFilePath("20130518_142022_3D657EBD.jpg")
-		//assert.False(t, result.IsZero())
-		assert.True(t, result.IsZero())
+		assert.False(t, result.IsZero())
```
lastzero: That must not work as you'll otherwise end up with "known" dates derived from filenames automatically created during import, although the original filename did NOT contain any such date. This is why import filenames are explicitly excluded right now.

Bur0k: Ah, I see. Didn't knew about the filename that is generated by PhotoPrism itself.

So for WhatsApp images the filter would have to explicitly only allow IMG-YYYYMMDD and VID-YYYYMMDD. Is this a good idea or does it just get too specific

Bur0k:
I changed the regex to a WhatsApp specific one in order to exclude any other date filenames.
[Regexr Link `(?:IMG|VID)-(\d{4})(\d{2})(\d{2})-WA`](regexr.com/6nlnv)

Would this be a solution to this problem?

Another question: Should i rebase the 2 commits into 1?

Bur0k:
Did you come to a conclusion about allowing to infer the date from these WhatsApp specific filenames?

graciousgrey:
The regex looks good now. @lastzero needs to have a look at our metadata parser again before we can merge it, it's on our to do :)

srett:
Since this hasn't been merged yet; would it be worthwhile trying to generalize the regex a bit more?

Something like `(?:IMG|VID|WP|PANO)[_-](?P<year>\\d{4})(?P<month>\\d{2})(?P<day>\\d{2})[_-]` might already cover several missing variants.

I just ran `find -iname "*.jpg" | awk -F/ '{print $NF}' | sed -r 's/[0-9]/x/g' | sort -u`:
`IMG_xxxxxxxx_xxxxxx.jpg` - `YYYYMMDD_hhmmss`, normal photo from Android
`PANO_xxxxxxxx_xxxxxx.jpg` - `YYYYMMDD_hhmmss`, panorama photo from Android
`IMGxxxxxxxxxxxxxx.jpg` - that's `YYYYMMDDhhmmss`, these are from an Oppo A5, can also be `*_HDR.jpg`
`WP_xxxxxxxx_xx_xx_xx_*.jpg` - `YYYYMMDD_hh_mm_ss` - Windows Phone/Mobile
`xxxxxxxx_xxxxxx.jpg` - `YYYYMMDD_hhmmss`, sometimes with suffix like `*_HDR.jpg`, from Galaxy S3 (GT-I9300)
`mmexportxxxxxxxxxxxxx.jpg` - These are pictures received via MMS on an old Sony Ericsson. The name contains the unix timestamp in milliseconds. So probably not when the photo was taken, but when the MMS was received.

The IMG ones work the same for videos, using VID instead.

M1cha:
Additionally, pixel phones use `PXL_20221224_131143486.jpg` and `PXL_20221224_103150236.MP.jpg` (MP=Motion Photo)

tobiasfiechter:
It would be great if Images saved by [Threema](https://threema.ch/de/) would be correctly organized by parsing the date from the filename.
Screenshot_20230121-162328~2
Screenshot_20230124-102625~2

Bur0k:
I would like to add more regexes for the different filenames.

Would this be okay for the maintainers? Will this PR get pushed back then?

lastzero:
As these changes would be rolled out to everyone, we should be very careful with that and it will require extensive testing which we currently don't have time for. For maximum compatibility and flexibility, it may also be better to use a parser function as for Exif and XMP instead of using a long list of regular expressions.

Bur0k:
> As these changes would be rolled out to everyone, we should be very careful with that and it will require extensive testing which we currently don't have time for. 

So if I understood correctly this PR should stay specific for WhatsApp file names, right?

lastzero:
@Bur0k We apologize for the long delay. I'll merge this now and may refactor it later if any changes are needed. Thank you very much for your contribution! :heart:

Bur0k:
> I'll merge this now and may refactor it later if any changes are needed. Thank you very much for your contribution! heart

If there is any other things to do regarding these filenames, just let me know (or point me to an issue :D). Be it regarding further "filename formats" or doing this without regex.

kilianyp:
sorry if this is not the right place, but the date is only extracted to write the file in the right folder, but not to write this into the metadata of the image, is it?

I am asking because images from whatsapp are missing the day/month/year attributes, which makes them hard to find as they are not sorted like the others. AFAIK the whatsapp file name isn't necessarily when the image was taken, but it would make it for me much easier to find if this was set as an attribute

graciousgrey:
If no real taken at date is available PhotoPrism does set the date from the Whats App filename as taken at date. You can find it on the edit dialog  or the card view for example.

PhotoPrism does not yet write metadata back to your original files.

Issue #1307, opened 2021-05-20
UX: Build new photo viewer


graciousgrey:
At the moment we use photoswipe as photo viewer. As there are some limitations we plan to implement one on our own.

The new photo viewer should solve the following issues:

* Proper support for videos (see #1278 & #843)
* Provide a way to view all photos of a stack (#386)
* Customize which actions are provided (#508 & #660)

jucor:
Is there any worth to "just" build an interface on top of Photoswipe rather than rewriting the whole viewer?

lastzero:
Well, it's architecture is not really extensible. It would all be ugly workarounds. Already did that in large parts, but need to make a cut somewhere.

jucor:
Makes sense :)
Thanks for the answer!

RandomHacks-Git:
I was going to open an issue but I guess the solution would be implemented as part of this new photo viewer.
After briefly testing photoprism I'm currently using another solution for the sole reason that there is no easy way to skip to the next video (or go back to the previous) inside the lightbox, always having to exit the lightbox and open the next video is quite cumbersome and not very intuitive.

graciousgrey:
@RandomHacks-Git Yes, that is one of the main reasons why we need to build our own photo viewer :)

aravindhsampath:
I am here from the context of [People Recognition](https://github.com/photoprism/photoprism/issues/22) work. It would be wonderful if the new photo viewer considers to implement an UI wherein the photo shows tags/labels associated with it visually like how Facebook photos with people. It may be triggered upon mouse hover in the area, or a button on the top that says"labels" or "tags" may be clicked to show this info.

![](https://www.researchgate.net/profile/Joao-Paulo-Pesce/publication/254008917/figure/fig1/AS:298081326256128@1448079534668/Example-of-photo-tags-in-a-Facebook-photo-Three-out-of-the-four-people-are-tagged-When.png)

When manual tagging of people gets implemented in Photoprism, it would be very useful to see "who all appear in this photo". Perhaps, also adding the ability to manually tag people right there while viewing the photo.

IssueHuntBot:
An anonymous user has funded $50.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/119160553/issues/1307) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/119160553/issues/1307).
[...]

PR #2433, opened 2022-06-17
UX: Search view render performance improvements

heikomat:
This is a followup to https://github.com/photoprism/photoprism/pull/2292

After implementing the "Virtualization", the app no longer got slower and slower the further you scrolled.
I was however not satisfied with how long the placeholder-elements were visible when scrolling a little faster.

The main reason for the time it took to render the real elements instead of the placeholders came down to the fact that one element in the mosaic- or card-views just renders a lot of components, mostly buttons and icons.
However, most of these Buttons and icons aren't actually visible most of the time, but are rendered anyway.

By only rendering the Buttons and icons when required, i reduced rendering times by about 50%.
(**UPDATE**: I halfed the rendering times again by replacing `<v-img>` and `<v-card>` with `<div>`s), making scrolling through the pictures much more fun! (it also sped up loading the next batch, because the browser spends less time rendering stuff and more time processing requests)

(**UPDATE2**: improved performance again by 1. reducing layout-time by scoping css in mosaic view, 2. replacing `<v-data-table>` with a regular html-table and 3. moving some inline-styles to css-classes)

(These screenshots only include the conditional button-rendering, not the other performance improvements!)
| before (laptop) | after (laptop) |
| --- | --- |
| Screenshot 2022-05-08 at 03 05 59 | Screenshot 2022-05-08 at 03 00 45 |

| before (smartphone) | after (smartphone) |
| --- | --- |
| scroll-ohne-changes | scroll-mit-changes |

See how effortlessly one can scroll through 2000 pictures without waiting seconds for them to appear and without impacting selection-performance:

https://user-images.githubusercontent.com/20770029/170862806-d5e97c10-6f58-4f97-88c3-ad953047a647.mp4

---

There was of course a reason why all buttons and icons were rendered at all times. It was implemented ([here](https://github.com/photoprism/photoprism/commit/8869e5b995e7169ccfc45eb91108e78dbc9431fe)) to prevent lag when selecting pictures.
AFAIK the cause of the lag was, that selecting a picture updated the whole view, rerendering all loaded elements.
The solution was to always render the icons with `display: none` and when required apply a css-class to the photo that would make the corresponding icons appear.

Because of the virtualization however, the root cause no longer exists, because only a handful of elements are ever rendered at a time, so when a photo is selected, only the visible few elements need updating (instead of thousands of elements)

---

~~There is at least one Bug in these changes that causes the selection-icon to not show after clicking it. it only appeares after moving the mouse away from the photo and back over the icon. The icons "highlight" is visible however. This is probably something really small to fix.~~
UPDATE: I Fixed that bug

Acceptance Criteria:


lastzero:
Our [development preview](https://docs.photoprism.app/release-notes/#development-preview) and [demo.photoprism.app](https://demo.photoprism.app/) have been updated so everyone can help test these changes! :birthday: 

Note that the full-screen viewer has recently been optimized to initialize only with the required data, which is already in the right format. However, it will load all the results from the top and additional results further down, so you can click "Next" for a while or start a slideshow:

- #1438

That was good enough with the old search results views due to other bottlenecks that now seem to be mostly resolved. We will provide a new viewer, but not in the very short term:

- #1307
 
In the worst case, we'll have to implement another workaround, such as using an offset when loading viewer content.

gyto6:
I've tested from Firefox, chrome and iOS App. The scrolling is way smoother and I'm no more interrupted. Sounds good for now.
No ressources issues seen from the server side.

PR #2441, opened 2022-06-18
fix scrolling in cards-view lagging regularly on small screens

heikomat:
On small screens in card view, there might only be 2 or 3 elements in the rendered area (screen area + 50% height in both directions).

For performance reasons, only the visibility of every 5th item is **currently** observed.

To determine the first- and last element to render, at least one observed item has to be in the rendered area at all times.
If that is not the case:
- the first- and last visible element index to render become `NaN`
- Therefore all loaded Elements get rendered
- As soon as an observed element enters the visible area the first- and last visible element become correct numbers again
- Therefore everything outside the viewport gets replaced with placeholders again

This "Render everything, then remove everything" is obviously slow, resulting in very noticable lag when scrolling.

This PR fixes that by having every 2nd element observed in the cards view, instead of every 5th.
I don't think there is a screen whose rendered area (screen area + 50% height in both directions) would contain < 2 cards, as that would mean that even a single card wouldn't fit that screen.

Acceptance Criteria:

### DIFF
--- a/frontend/src/component/photo/cards.vue
+++ b/frontend/src/component/photo/cards.vue
@@ -317,13 +317,11 @@ export default {
       }
 
       /**
-       * observing only every 5th item reduces the amount of time
-       * spent computing intersection by 80%. me might render up to
-       * 8 items more than required, but the time saved computing
-       * intersections is far greater than the time lost rendering
-       * a couple more items
+       * observing only every 5th item doesn't work here, because on small
+       * screens there aren't >= 5 elements in the viewport at all times.
+       * observing every second element should work.
        */
-      for (let i = 0; i < this.$refs.items.length; i += 5) {
+      for (let i = 0; i < this.$refs.items.length; i += 2) {
         this.intersectionObserver.observe(this.$refs.items[i]);
       }
     },


lastzero:
The iPhone SE (2nd generation) is probably the smallest device that (still) has a relevant number of users. It's screen has about enough space to show ~1.5 cards. However, even worse, this is what I see when I open our demo with the Apple Device Simulator v12.4 under iOS 14.4 (note that images not loading could partly be a problem with KeyCDN and old CA certificates on the device - still seems wrong there is no space reserved for the thumbnail image and/or a placeholder):

heikomat:
> The iPhone SE (2nd generation) is probably the smallest device that (still) has a relevant number of users. It's screen has about enough space to show ~1.5 cards. However, even worse, this is what I see when I open our demo with the Apple Device Simulator v12.4 under iOS 14.4 (note that images not loading could partly be a problem with KeyCDN and old CA certificates on the device - still seems wrong there is no space reserved for the thumbnail image and/or a placeholder):

That is because safari 14 doesn't support the aspect-ratio css-property.
The images are there, but don't have any height

lastzero:
Would it then make sense to publish more detailed browser/device requirements? Currently we only state that the UI should work in "all modern browsers".

heikomat:
> Would it then make sense to publish more detailed browser/device requirements? Currently we only state that the UI should work in "all modern browsers".

We should, but not too detailed. we should define what "modern" means without specifing browser versions, becaus that would make the requirements outdated in about a year.

But i just looked at the list of devices that support iOS 15, and the 2nd gen iPhone SE is listed (even 1st gen).
So i'm curios: are you running iOS14? if so, can you update to iOS 15? if so, does that fix the images not being visible?

lastzero:
I rarely use the simulator and simply didn't have the time or motivation to update before testing.... Later I realized that it might not be the latest iOS, so I mentioned it for documentation purposes and because I don't want to waste your time forcing you to guess! 🤗

Edit: I may not be able to run iOS 15 because Apple is no longer providing software updates for my iMac, even though it is much faster than my MacBook Air, which continues to be supported (but is too slow for it).

Screenshot 2022-06-18 at 14 18 00

PR #2443, opened 2022-06-19
Fix Card border-radius and mosaic selected box-shadow

heikomat:
### DO NO MERGE YET.
- Performance has gotten worse in mosaic view, because the photos-array is accessed for placeholders now
- It is not yet tested on more browsers

---

- The selected mosaic-entries currently have no box-shadow, even though they should
- The Elements in the cards-view have no border-radius at the bottom, even though they should
- The Images in the elements in the cards view have a border-radius at the bottom, even though they shouldn't

This PR fixes thexe things.
It also implements two other cleanups
- make sure the element that has the `result`-class is a parent-element of the one that has the `image`-class in the mosaic view. cards- and list-view were already implemented that way, and to reduce special-cases, i adjusted mosaic-view to work the same.
- remove now unnecessary placeholder-element in mosaic-view by making the container that holds the whole element anyway look like the placeholder
- replace `box-shadow: 0 0 0 0 ....` with `box-shadow: none`, because a `0 0 0 0`-box-shadow is not visible

| before | after |
| --- | --- |
| card-before | card-after |
| mosaic-selected-before | mosaic-selected-after |

Acceptance Criteria:

- [ ] **Features and improvements are fully implemented** so that they can be released at any time without additional work

### DIFF
--- a/frontend/src/component/photo/cards.vue
+++ b/frontend/src/component/photo/cards.vue
@@ -28,7 +28,7 @@
           :data-index="index"
           class="flex xs12 sm6 md4 lg3 xlg2 xxxl1 d-flex"
       >
-        <div :class="photo.classes()" class="result card-background accent lighten-3">
+        <div class="result card-background accent lighten-3">
           <div v-if="index < firstVisibleElementIndex || index > lastVisibileElementIndex" class="placeholder">
             <div class="accent lighten-2 image"/>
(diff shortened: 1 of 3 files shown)


heikomat:
ok, forget about this PR.
The way i went about implementing this was absolutely not pragmatic and broke more than it fixed.

The things mentioned here should be fixed at some point, but they are very minor design improvements, and absolutely not relevant right now

PR #2448, opened 2022-06-19
UX: Support older browsers by adding aspect-ratio CSS fallback

heikomat:
some not sooo old browsers (Safari on iOS 14.5) do not yet support the aspect-ratio css-property, resulting in images having no height in the search-results.

This adds a fallback for browsers that don't support the aspect-ratio css-property

Acceptance Criteria:

### DIFF
--- a/frontend/src/css/results.css
+++ b/frontend/src/css/results.css
@@ -199,7 +199,7 @@ body.chrome #photoprism .search-results .result {
 }
 
 #photoprism .search-results .image-container {
-  aspect-ratio: 1;
+  /*aspect-ratio: 1;*/
   contain: layout paint style size;
 }
 
@@ -209,7 +209,7 @@ body.chrome #photoprism .search-results .result {
 {
   position: relative;
   user-select: none;
-  aspect-ratio: 1;
+  /*aspect-ratio: 1;*/
   background-position: center center;
   background-size: cover;
   background-repeat: no-repeat;
@@ -481,14 +481,13 @@ body.chrome #photoprism .search-results .result {
 }
 
 /* old browser support */
-@supports not (aspect-ratio: 1) {
+/*@supports not (aspect-ratio: 1) {*/
   #photoprism .search-results .image-container {
     height: 0;
     padding-bottom: 100%;
   }
 
   #photoprism .cards-view .result .image,
-  #photoprism .list-view .result .image,
   #photoprism .mosaic-view .result.image
   {
     height: 0;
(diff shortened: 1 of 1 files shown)


lastzero:
Code generally good to me. Should we also test it with specific browsers? Theresa could do this next week.

heikomat:
> Code generally good to me. Should we also test it with specific browsers? Theresa could do this next week.

Testing this specifically makes only sense in browsers that don't support `aspect-ratio`, as for all others nothing has changed.

I personally tested it with my ipad mini 2, running iOS 12.something.
With this fix it was fully usable, and i could only find a single visual difference: in the mosaic view (on the old ipad) the vertical space between elements is about 4px too large, probably because of the way old safari handles flex-elements and margins.

Interestingly enough, if i force a modern browser to use the padding-hack instead of the aspect-ratio property, this spacing in the mosaic view looks 100% correct.

So in short: we should find a list of browsers and devices we want to support, that do not support `aspect-ratio` (see [caniuse](https://caniuse.com/mdn-css_properties_aspect-ratio)). (for example your iPhone SE 2nd gen)
My expectation would be, that it is 100% usable and visually at least very close to what is shown in modern browsers.

I'd not expect it to look 100% pixel by pixel identical. If that is something you'd like to achieve, this fix needs a little more tweaking.

lastzero:
> I'd not expect it to look 100% pixel by pixel identical. If that is something you'd like to achieve, this fix needs a little more tweaking.

Confirmed. As mentioned in the chat, a best effort approach is enough, especially when it comes to minor layout issues. These can be resolved later if we run out of work and feel that this will improve the user experience.

heikomat:
> if we run out of work

😂

lastzero:
Merged because it shouldn't break anything and it's easier to test this way.

graciousgrey:
Looks good now on most old  browser versions I tested and that do not support the aspect-ratio :)

Performance is amazing!

In older Edge versions (Screenshot is from Windows 10 Edge 86) the card view looks weird. I am not sure if it is related to your changes or looked like this before. For me it is not a MUST to support this as in newer versions it looks good.

windows-10-edge-86

heikomat:
> Looks good now on most old browser versions I tested and that do not support the aspect-ratio :)
> 
> Performance is amazing!
> 
> In older Edge versions (Screenshot is from Windows 10 Edge 86) the card view looks weird. I am not sure if it is related to your changes or looked like this before. For me it is not a MUST to support this as in newer versions it looks good.

Thanks for testing! :)

It's probably not related to the aspect-ratio property, because that should only affect the size of the images in that view, not of the whole cards.

It might however be related to the performance improvements in general.
I've seen this bug before when the icon-font-size wasn't correct. The icons then don't show up correctly or at all, and the cards get tall.

I might check later if it is an easy fix (but no guarantees). My bet is on the icon-size being incorrect there.

heikomat:
@graciousgrey have a look at [this](https://github.com/photoprism/photoprism/pull/2456)

PR #2513, opened 2022-07-10
Exif: Extract creation time from DateTimeCreated, if available

carlonluca:
exiftool seems to also be able to extract the composite DateTimeCreated
tag. Adding also that tag may increase the chances of getting a
valid date for the photo.

Acceptance Criteria:

### DIFF
--- a/internal/meta/exif.go
+++ b/internal/meta/exif.go
@@ -25,7 +25,7 @@ import (
 var exifIfdMapping *exifcommon.IfdMapping
 var exifTagIndex = exif.NewTagIndex()
 var exifMutex = sync.Mutex{}
-var exifDateTimeTags = []string{"DateTimeOriginal", "DateTimeDigitized", "CreateDate", "DateTime"}
+var exifDateTimeTags = []string{"DateTimeOriginal", "DateTimeDigitized", "CreateDate", "DateTime", "DateTimeCreated"}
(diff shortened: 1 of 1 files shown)


lastzero:
Thank you! Could you also provide us with a test file? As far as I know, [this field is IPTC specific in Exiftool](https://exiftool.org/forum/index.php?topic=7121.0), which explains why I have never seen it before...

lastzero:
Started a new preview build for testing :+1:

carlonluca:
It should be possible to create one specifically for the test with exiftool. This is the metadata of an image from my collection that can be used:

```
exiftool ...  | grep -i date
File Modification Date/Time     : 2015:01:10 14:17:29+01:00
File Access Date/Time           : 2022:07:22 19:25:28+02:00
File Inode Change Date/Time     : 2021:12:15 23:16:13+01:00
Modify Date                     : 2007:02:19 00:00:00
Date/Time Original              : 0000:00:00 00:00:00
Create Date                     : 0000:00:00 00:00:00
Date Created                    : 2007:02:19
Profile Date Time               : 2003:07:01 00:00:00
Date/Time Created               : 2007:02:19 00:00:00+00:00
```

This is a mess, as DateTimeOriginal is there, but is null. DateTimeCreated is the only valid one. With my patch, everything is now properly sorted for me. Before the patch, it was placed on top.

lastzero:
Ended up a bit more work, check these commits and the preview 👍

Issue #1187, opened 2021-04-11
Places: Cluster View


red-avtovo:
Thank you very much for the project! I really enjoy using it.

I would like to notice, that it would be useful to group photos on a small scale which are very close to each other or let to open cluster view (similar as a moments view) and show the subset of photos from the place, clicked on. It is not very easy to navigate through the subset of the photos or share a group of them from a specific place
![image](https://user-images.githubusercontent.com/10682826/114313988-24ced100-9af9-11eb-984a-9fce0b3271dd.png)

shawnbarton:
I came here to make this exact suggestion. This is an important feature to me so I'm chiming in to show interest. Keep up the great work.

lastzero:
Thank you very much! It's true that we need to optimize this further. As you might have noticed, the maps already cluster automatically - but only up to a certain zoom level as you otherwise might not be able to see individual photos. It's more work to handle situations as this, e.g. by showing a "stack" that then opens in the photo viewer like stacks in regular search results.

magnuznilzzon:
I suspect it is pretty common to have the situation that @shawnbarton has, especially people with families will have a ton of pictures centered on their home. I know I do, that is how I found this thread.

el-tiuri:
Great idea! I like the way it’s implemented in iOS’ Photos app, and this would be pretty similar. The way it works on iOS is that it dynamically creates stacks of photos that are close together relative to the zoom level, and if you tap on a stack you get a grid of all the photos in that stack.

I especially like this because it enables you to easily see all the photos you took in a particular place (regardless of scale - could be a city, could be a continent) regardless of when you took it.

I think limiting the automatic clustering functionality to low zoom levels would be a bit limiting compared to the use it could have, and I don’t really see the benefit of it compared to enabling it on all zoom levels.

Let me know if you want me to share some screenshots of the iOS photos app, in case that’d be helpful.

ohthehugemanatee:
I have the same issue. We in fact already have the dynamic stacks. Just when you click/tap on a circle with a number of photos, that should take you immediately to a pre-filled search results page. That would be plenty.  Users can zoom in closer with the plus/minus keys and get more precise, smaller stacks.

fivestones:
I agree with @ohthehugemanatee. Right now the circles with the number inside just zooms in when clicked. Instead, clicking on these circles should open a search results page with all those photos shown. Zooming in on the map is easily done other ways and doesn't need to be done by clicking on the circles that have numbers inside.

[...]

PR #2596, opened 2022-08-06
Places: Add Cluster View

heikomat:
This is an attempt to implement the cluster-view proposed in https://github.com/photoprism/photoprism/issues/1187

Acceptance Criteria:


lastzero:
@heikomat We would love to test and merge this if you think it is ready to be included in our development preview? To do this, remaining conflicts need to be resolved - hopefully just the `package-lock.json` :crossed_fingers:

heikomat:
@lastzero github says there are no conflicts with the base branch.
give me a couple of minutes to merge develop into this branch

reubendowle:
This is a great improvement to the places screen. I think the way the clusters are created needs a little bit of work. I tested this out centered on my house, where there are many thousand photos. I can see it creates several independent clusters of photos centered around my house.

image

The red clusters are showing photos taken in my house, but showing as two separate clusters in the same house, and one large one centered in the neighbors house.

I think perhaps there needs to be a bigger minimum radius for the clustering?

heikomat:
@reubendowle could it be that the GPS coordinates in the pictures aren't fully accurate?

If they were shot with a Smartphone, its GPS info might have been inacurate the moment the picture was taken

reubendowle:
> @reubendowle could it be that the GPS coordinates in the pictures aren't fully accurate?
> 
> If they were shot with a Smartphone, its GPS info might have been inacurate the moment the picture was taken

I agree that inaccurate GPS information is definitely part of the issue.

But I also think it is a bit silly to have two separate groups with 99+ photos shown in the same building - it seems like a slightly larger minimum radius for groupings would be a simple way to improve the experience here.

That said, I don't think this minor thing should prevent this change being merged - it is much better than how it works at the moment.

lastzero:
@heikomat It seems that one of our changes has created another conflict. Once this is resolved and we have your go, I'll be happy to merge this to develop for further testing! :heart:

heikomat:
Will mergen develop tomorrow and resolve the conflict

lastzero:
Thank you very much! I'll merge this now and look forward to testing it on our demo :)

heikomat:
@lastzero the conflict was super minor.
You added a `:context="context"` to the `<p-photo-toolbar` in `photos.vue`.
In my branch i made the `<p-photo-toolbar` multiline because of the number of properties (it was single-line before). That's why git couldn't auto-merge it.

You "have my go" as far as i assume this to be an all around improvement, even when it's not perfect.
Have not done any further testing in the last while though.

lastzero:
@heikomat Would it be complicated to display the view next to (e.g. above or below) the map, so that the page is split in two sections, instead of using an overlay? That way you could see both at the same time.

heikomat:
@lastzero i'd assume that its at least not easy. Depends on the flexebility of the `<v-dialog`-component.
You'd also have to keep in mind that having them side-by-side would be nice on widescreen, but not possible on mobile

heikomat:
Oh, also the mosaic- and cards-view have the number of columns defined via css-selectors that base their calculations on the screen width. So this here works because the modal width is close to the screen width. Having the photos-view on half the screen would currently show way to many columns

lastzero:
I could also imagine displaying the overlay similar to this, so that users see a bit more context in the background and the full window height can be used:

overlay

When you open issue details in the GitHub project view, these are also displayed in an overlay on the right-hand side, using (almost) the full height:

github-example

Tile width should be adjustable via CSS if needed, unless there is additional JS code that needs to be changed?

graciousgrey:
> You "have my go" as far as i assume this to be an all around improvement, even when it's not perfect.
Have not done any further testing in the last while though.

Thank you, this is definitely a great improvement :heart: 
I will start with further testing today.

lastzero:
I've reduced the padding in the Mosaic view to make more use of the available space, especially on mobile screens:
- #3572

graciousgrey:
First of all - I love the clusters! It's much more fun to explore places now :)

Here are some first test results: 
1) Within the cluster window: When I type a search term and click enter the component closes instead of showing search results.
2) Within the cluster window: The filters on the extended search bar do not work.
3) When searching for e.g. flowers on places, I see clusters of only images with flowers. If I open such a cluster all images from this cluster are shown also the ones that do not contain flowers.

loulou91:
Just made a test. Promising but there are problems of performance with the display when zooming in.
I'm using "default" map, the first steps of zooming in are very slow even lagging till a change of detail level of the map. Then it's fluent...
Not sure to be clear but it should be easy to reproduce.

graciousgrey:
@loulou91 Do you experience this on your personal instance or our public demo? Can you confirm that you don't observe this issue with the latest stable release?

loulou91:
Hi Theresa, I'm working on last preview release! I just went back to latest release and all was fine!
image

lastzero:
My subjective feeling was also that it is a bit slower now than before. I suspect this could be due to the rendering of the previews on the map. Unfortunately, I probably won't have time to look into this until next week.

loulou91:
To be clearer, till this level of details it is very slow and it lags:
image
and from this level of details, no more problem...
image

reubendowle:
I have resolved the performance issue with this pull request https://github.com/photoprism/photoprism/pull/3577

heikomat:
> I have resolved the performance issue with this pull request #3577

Crazy that i haven't noticed the performance issue, espacially regarding that the fix is not that hard. Well done 👍

loulou91:
I made a pull of preview release this morning but the problem of performance is still there as I described it above.
Maybe the fix is still to be mergred in preview version?

graciousgrey:
@loulou91 We plan to merge the changes today :) Many thanks for your help with testing!

davidetoile:
Hey!
Thanks for this great feature. 
I have noticed that in some cases, when opening some clusters, grid of images is either empty either not corresponding to cluster count displayed…

Issue #2627, opened 2022-08-21
search-offset for back-navigation never reset


heikomat:
#### 1. What is not working as documented?

Albums-, Labels- and subject-views load their data lazy. When the end of the loaded data is about to be reached, the next batch gets loaded.
Additionaly, the current load-offset is remembered in localStorage (in `albums_offset`, `share_albums_offset`, `labels_offset` and `subjects_offset` respectively).
This load-offset is used on back-navigation, so that when the user for example exits an album via back-button, photoprism knows how many items it needs to load to recreate the scroll-position the user had when entering the album.

It seems that this remembered offset can only ever increase in value. i found no code that would decrease or reset its value.
This kinda kills the benefits of lazy-loading for future visits.

If i ever scrolled through for example 500 albums/labels/subjects, then on any future visit (with the same browser), photoprism will always load at least 500 elements, even when i start at the top of the search-results (as i'm supposed to) when navigating without using the back-button.

I noticed this when trying to implement the same behaviour for the regular results-view to fix https://github.com/photoprism/photoprism/issues/2597

#### 2. How can we reproduce it?

1. create a lot of albums or labels
2. scroll to the end of the list of elements, until no more elements are loaded
3. refresh the page while observing the network requests
4. observe how all albums/labels are loaded and rendered at once, even though you're at the top of the list

#### 3. What behavior do you expect?

i'd expect photoprism to not load much more elements than required when opening albums/labels

#### 4. What could be the cause of your problem?

The respective offsets are never reset.
Alternatively: the offsets are not ignored on non-back navigation

#### 5. Can you provide us with example files for testing, error logs, or screenshots?

I can if required, but i think the issue is really self explainatory.

#### 6. Which software versions do you use?

- Photoprism Build 220819-5d0a2ccf1 in a docker-container
- Firefox 103.0.2 on macOS 12.4

lastzero:
Leaving it like that was decided to save time. At a minimum we need to choose a strategy that meets the expectations of most/all users. Easiest might be to use a timeout. Not sure if there's a reliable way to know if the back button was used. Link targets may be on a different site in theory.

heikomat:
@lastzero I understand. Usually loading the 100 or something albums people have at once is really not a problem.

I came across this because having photoprism not load much more than required (aka correctly resetting the offset-value) would allow fixing the "Scroll-Pos after image download in the iOS-PWA"-issue without causing huge loads on regular search-view

lastzero:
[...]

Issue #2597, opened 2022-08-07
UX: Restore scroll position when navigating back


whoshoe:
#### 1. What is not working 

After clicking the download button on iOS and using the "X" to close the image view, the scrolling position is not remembered. Page goes back to the top.

#### 2. How can we reproduce it?

1. On iOS scroll down and open an image
2. Click on the download button, click "Done" and click "X" to go back to the list view.
3. Page goes back to the top

#### 3. What behavior do you expect?

Expect scrolling position to be remembered.

#### 4. Which software versions do you use?

(a) PhotoPrism Architecture & Build Number: AMD64, Build [220730-0e1222c83](https://docs.photoprism.app/release-notes/)

(b) Database Type & Version: MariaDB

(c) Operating System Types & Versions: iOS 15.6

(d) Browser Types & Versions: Safari on iPhone

lastzero:
@heikomat might be related to the virtual scrolling? AFAIK it was working before.

heikomat:
@lastzero possible. Would need to test that. Might have a look at it later, but I don't know if I find time today

heikomat:
@whoshoe two questions:

1. after clicking the x, is the scroll position instantly reset, or does it instead scroll to the top?

The "x" to close the image and the button to scroll to the top are at a similar location, maybe one click somehow triggers both.

2. Does it always happen when closing an image using the "x", or only after downloading it?

whoshoe:
> 1. after clicking the x, is the scroll position instantly reset, or does it instead scroll to the top?

I didn't see any scrolling actions being preformed. After clicking the x, the page just shows the top as if I was at the top.


> 2\. Does it always happen when closing an image using the "x", or only after downloading it?

Nope. Going into the image view and exiting using the "x" without clicking download DOES remember the position. Loss of scrolling position only happens after the download page being brought up.

lastzero:
Note I've fixed a similar issue in https://github.com/photoprism/photoprism/commit/5d7448d2df3d05580489b3e208ea09b8935bf69b by disabling history mode in full screen view when releasing the [search view render performance improvements](https://github.com/photoprism/photoprism/pull/2433) (probably also worked before):

https://github.com/photoprism/photoprism/blob/5d7448d2df3d05580489b3e208ea09b8935bf69b/frontend/src/common/viewer.js#L105

**See also:**
- #896

For this, we implemented the following event handler...

[...]

Issue #2571, opened 2022-08-01
UX: Close image viewer on browser back


agramner:
**Is your feature request related to a problem? Please describe.**

I mostly use PhotoPrism as PWA on mobile (Chrome Android) and one usage pattern I always do wrong (and annoys me) is:

1. I'm scrolling the gallery, looking at my pictures and find a picture I want to take a closer look at
2. I'm pressing the picture and the image viewer opens
3. After I'm done looking at the picture I press the native back button on my phone
4. I expect the image viewer to close but it doesn't.
5. I press the close button in the top right corner instead
6. I find the back action just navigated away from where I was in the gallery page and I need to start over scrolling the gallery (or do a new search)

I do this all the time and I guess it's because the (Samsung) gallery app on my phone closes the image viewer on back button.

**Describe the solution you'd like**

I think most users would expect the image viewer to close when pressing the back button

**Describe alternatives you've considered**

Maybe move the close button and/or change the icon to a back arrow

**Additional context**

Maybe a URL hash part (for example #imageviewer) could be added to the URL when the image viewer is opened

ark-:
Agreed, I swipe backwards on iPhone (equivalent to back on Android) and the same happens to me. Instead I have to swipe down.

Issue #1307 is looking to revamp the photo viewer. So perhaps this should be rolled into that work?

lastzero:
You're welcome to send a pull request while we are on vacation! I think we had history mode turned on but it interfered with the new (faster) scrolling, which is why I've turned it off recently. Handling the events manually instead of using the built-in handler should work. Just didn't have any time for that so far.

agramner:
Ok I see. May try to fix this if I have some time over.

lastzero:
Awesome, thank you! It's probably related to this issue:
- #2597

See the code I referenced in https://github.com/photoprism/photoprism/issues/2597#issuecomment-1211188162.

heikomat:
> Ok I see. May try to fix this if I have some time over.

@agramner I'm highly interested in fixing history mode, and am the one who implemented the virtualization that was the cause to disable it.

But like you, my time is tight, so I just haven't had the possibility to do so.
I'll probably have some time the coming Friday (19.08.).

If you find the time to implement fixes before I do, please feel free to ask anything you'd like to know about the virtualization. I'm happy to help you out where I can

agramner:
@heikomat Aah ok, nice work with the scroll performance! A really great improvement of a key feature. 

[...]

Issue #2477, opened 2022-06-28
UX: View type of search results and albums must be remembered


lastzero:
**When you switch between card, mosaic and list view, the currently selected view type must be remembered. This feature has already been implemented. The functionality needs to be restored as it seems to have been broken during a recent refactoring.**

In addition, the viewer's history mode must be deactivated, since the result views repeatedly jumped back to the top after closing the viewer while testing.

graciousgrey:
Released since no problems were reported!

heikomat:
@lastzero do you happen to remember the test-setup you had while observing the issues with history mode?

lastzero:
Regular chrome browser from what I remember. With the current release, the problem of course is not reproducible as history mode is disabled.

heikomat:
i understand, but even with history mode enabled (in the local dev setup) i can't seem to reproduce the "jump back to top" issue

lastzero:
When you use the back button of the browser? Not sure if only/also happens if you just close the viewer. Try esc key as well. Also it may depend on if you use back / next within the viewer.

heikomat:
a quick test of different combinations (back button, esc, with and without using back/next in the viewer) seems to indicate, that only the back-button of the browser (or the back-button on an android device) is impacted by history mode.

Everything else seems to act the same (at least acording to the tests i did just now).

When pressing the back button while the viewer is open:
- With history mode disabled, the browser navigates away from the result view in the background by going to the previous view **without** closing the photo-viewer (this is most likely not what one would expect)
- with history mode enabled, the photo-viewer gets closed

i'd love to fix history mode (a friend of mine even told me that it is annoying, that the back-button on android doesn't close the viewer but instead navigates away).
But i can only do that if i can reproduce the bugs that history-mode causes that you observed

lsueslack:
It would be so great if you could please fix the history mode.

lastzero:
Still on vacation, I don't have a computer to test. Just a glass of red wine and my phone. If you can wait until later next week, I'll send you a video.

heikomat:
Ah sorry, didn't know you were still on vacation.
I don't think trying to reproduce it myself much more would be very effective, as i don't know what the difference between your test and my test might be.

So i'll just wait until you come back from vacation and had time to reproduce it again.
That's totally fine with me. Enjoy your time off. Rome wasn't built in a day :P

lastzero:
[...]

PR #2648, opened 2022-08-26
UX: Restore scroll position when navigating back

heikomat:
This is a fix for https://github.com/photoprism/photoprism/issues/2597
This also re-enables history-mode, and i think that is actually required for this fix to work, but i haven't tested it yet.

These changes do 4 things
1. Detect back-navigation, and reset remembered load-offsets if the component was navigated to without using the back-button.
This is so that not every future visit of the search-results loads hundrets/thousands of images
    - see https://github.com/photoprism/photoprism/issues/2627
2. Re-enable history-mode.
History mode was enbaled in the past and got disabled because of issues when the scroll-performance was improved. However, these issues couldn't yet be reproduced again
    - see https://github.com/photoprism/photoprism/issues/2571
    - and https://github.com/photoprism/photoprism/issues/2477
3. Disable initial search in search-view when the viewer is open (happens when returning from the download-screen in the iOS PWA)
4. After initial search, scroll to the last opened photo by remembering its `uid` in localStorage when opening the viewer.

**On back-navigation in iOS PWA-mode, the results now load enough images and auto-scroll to the last opened image.**

One known issue is that this can cause too many images to be loaded once when navigating to the search-results using the back-button without having had the image viewer open 

**You can test it wit the [`ghcr.io/heikomat/photoprism:backnav`](https://github.com/heikomat/photoprism/pkgs/container/photoprism/39385259?tag=backnav) docker image**

Acceptance Criteria:


heikomat:
TODO: resetting the remembered offset happens in `search` for the photos-component, but in `created` for all other components.

It needs to happen in `search` for all components, because `search` is called by `created` AND the `$route`-watcher.
Moving it to search could prevent to many items being loaded after a filter gets applied

UPDATE: Fixed

Rolf-Smit:
I did some quick testing with this build and this is an enormous improvement in the user experience on mobile (Android).  The back button not closing the image viewer has been my number one annoyance, seeing it working as one would expect is awesome!

Nice job!

lastzero:
Seems this is ready for the preview so more users can test it?

heikomat:
yes, this is ready to be tested!

lastzero:
Shouldn't this be added to `share.js` as well so it works the same when sharing?

heikomat:
> Shouldn't this be added to `share.js` as well so it works the same when sharing?

Could very well be. I think you're talking about the popstate-detection in `app.js`?
Should (for now) only affect scenarios where a shared album is opened in PWA mode, but having the shared-scenario work differently than the regular scenario should definitely be avoided :)

lastzero:
> Could very well be. I think you're talking about the popstate-detection in `app.js`?

Yes, I moved the handler to `common/navigation.js` for now so that the files can be included by both. Does it make sense to use a global variable, or do you think it could/should be refactored as well?

Also, there was a question about the scrollbar in *Places*.... is this still relevant or have you found an answer yourself?

heikomat:
> > Could very well be. I think you're talking about the popstate-detection in `app.js`?
> 
> Yes, I moved the handler to `common/navigation.js` for now so that the files can be included by both. Does it make sense to use a global variable, or do you think it could/should be refactored as well?
> 
> Also, there was a question about the scrollbar in *Places*.... is this still relevant or have you found an answer yourself?

I think a global variable makes sense, but am open to suggestions. But! It Must be registered before the router, otherwise the event handler doesn't work.

The question regarding the scrollbars was related to the Cluster-View-Feature. Not this PR

lastzero:
Loading it at the very top :heavy_check_mark: Check the code!

Issue #1164, opened 2021-03-31
UX: Make information shown in Cards View configurable


yinkiu602:
The reasons that I raised this is because when I first try out to index my file, in Library\originals\{mounted folder name}, photo are named with their original file name. However, when I tried to share a link of the album of the indexed photo, it seems that the page will only show the TITLE, which is modified with things like Location\Year. 

So, I wish that:
A) Force the title of the indexed/uploaded photo to be the same as their original file name
or,
B) Ability to change the properties to display in the album. AFAIK the default now is TITLE, TAKEN, CAMERA, LOCATION. If it is possible to change the properties to display, i.e. :  FILENAME, TAKEN, CAMERA, LOCATION, then the code to change TITLE can remains unchanged.

By allowing the photos to display with their original file name, it will be easier for users to to mention the exact image with others, i.e. : "20210311_012804.jpeg", instead of seeing a whole page of unrecognizable title like: "Seashore / United Kingdom / 2021".

Thanks for anyone who helps!

graciousgrey:
Our card view shows the title PhotoPrism generates for each photo and not the original name. 
This is intended as those titles describe what's on the image whereas the original filenames often consist of numbers only. We don't plan to change the default behaviour.

What we could do at one point is adding an option to  configure which information is shown on the card view e.g. original name instead of filename. But first we have some other tasks with higher priority.

graciousgrey:
Please note that [our latest release](https://docs.photoprism.app/release-notes/) additionally displays the filename on the card view and allows you to delete generated titles.

lastzero:
We've just added these settings, so you can prevent titles and/or captions from being displayed in search results:

![Image](https://github.com/user-attachments/assets/c3bc24c8-b5c0-4653-8d04-3b45a6c13b07)

The changes will be [available for testing in our preview build](https://docs.photoprism.app/getting-started/updates/#development-preview) soon. Any help with that would be much appreciated!

lastzero:
Our preview build has just been updated with all the latest changes:
- https://docs.photoprism.app/release-notes/#development-preview

PR #2694, opened 2022-09-12
Search: Configurable display options for the Cards result view

telsch:
Add options to display Keywords and Notes in Card View.
May related to #1164

Acceptance Criteria:

       'Show Keywords', 'Show Notes', 'Card View Display options'

### DIFF
--- a/internal/search/photos.go
+++ b/internal/search/photos.go
@@ -59,7 +59,8 @@ func searchPhotos(f form.SearchPhotos, sess *entity.Session, resultCols string) 
 		Joins("JOIN photos ON files.photo_id = photos.id AND files.media_id IS NOT NULL").
 		Joins("LEFT JOIN cameras ON photos.camera_id = cameras.id").
 		Joins("LEFT JOIN lenses ON photos.lens_id = lenses.id").
-		Joins("LEFT JOIN places ON photos.place_id = places.id")
+		Joins("LEFT JOIN places ON photos.place_id = places.id").
+		Joins("LEFT JOIN details ON details .photo_id = photos.id")
 
 	// Accept the album UID as scope for backward compatibility.
 	if rnd.IsUID(f.Album, entity.AlbumUID) {
--- a/frontend/src/component/photo/cards.vue
+++ b/frontend/src/component/photo/cards.vue
@@ -68,6 +68,20 @@
                   <i/>
                   {{ photo.locationInfo() }}
                 </template>
+		<template
+		    v-if="displayOption.keywords"
+		    :title="$gettext('Keywords')"
+		>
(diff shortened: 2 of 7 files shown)


heikomat (changes requested):
Cards (amongst other things) are pseudo-virtualized.
That means that cards that are far enough outside the viewport only render partially to improve overall performance.

For this to work, the pseudo-virtualized cards need to have the same outer size as their regular in-view counterparts. If they don't, the position of the cards might start to jump around when scrolling.

Because dynamic texts inside the cards (like the keywords and notes) will affect the overall size of the card in an unpredictable way, these texts need to also be rendered in the virtualized cards (so they maintain their exact size).

You'll notice, that all texts inside the cards are rendered in two places in the `cards.vue`.
Once for the virtualized card, once for the regular in-view card.
Please make sure, that your newly added texts are also rendered in the virtualized cards.
They need to be added at about line 68, just below the `{{ photo.locationInfo() }}`-block

telsch:
Added the newly texts also in the rendered virtualized cards and fixed the album view.

lastzero:
@heikomat Does the frontend part look good to you now? Looking at the backend code, especially the SQL joins, I'm afraid this may have performance implications. We intentionally put all the large text fields in the details table so they wouldn't get in the way when searching. If we now always join with this table, that advantage would seem to be lost. I think I also spotted a space that might break the query (see comment).

internal/search/photos.go
```diff
@@ -59,7 +59,8 @@ func searchPhotos(f form.SearchPhotos, sess *entity.Session, resultCols string)
 		Joins("JOIN photos ON files.photo_id = photos.id AND files.media_id IS NOT NULL").
 		Joins("LEFT JOIN cameras ON photos.camera_id = cameras.id").
 		Joins("LEFT JOIN lenses ON photos.lens_id = lenses.id").
-		Joins("LEFT JOIN places ON photos.place_id = places.id")
+		Joins("LEFT JOIN places ON photos.place_id = places.id").
+		Joins("LEFT JOIN details ON details .photo_id = photos.id")
```
lastzero, line 63: Probably should be `Joins("LEFT JOIN details ON details.photo_id = photos.id")`, so without the extra space?

lastzero:
Since we can't easily merge it given the potential issues and questions outlined above, I'm closing this PR for now. However, we plan to add similar functionality once we have time to work on a concept and do the necessary testing to avoid bugs in a core functionality that would impact all of our users.

@telsch If you have time to get the changes to the point where we can merge them without further refactoring and testing, feel free to reopen the PR or create a new one. Thank you very much for your contribution!

PR #2721, opened 2022-09-19
JPEG: Automatically fix files with missing EOI marker

### LINKED ISSUE #2463: JPEG: Automatically check and repair broken/invalid images
### Go happily refuses to open JPEGs if they contain any glitches. As a user with such files, you still want to index and view them. It would therefore make sense to automate the repair of JPEGs, especially since the process is always the same.

**Step 1:** Use GraphicsMagick to check a JPEG for issues

GraphicsMagick is a great tool for this task. Depending on what operating system you are using, you may need to install it first: http://www.graphicsmagick.org/identify.html

Now run this command and check the report for any problems:

```
gm identify -verbose broken.jpg
```

*Note that ImageMagick may also have a command for this. This describes the specific process I used and that worked for me. There are variants and it can likely be optimized so that fewer tools and libraries need to be installed.*

**Step 2:** Use the ImageMagick `convert` command to create a valid copy of the image

To create a valid copy of the original JPEG:
 ```
convert broken.jpg fixed.jpg
```

Any glitches should then be fixed i.e. the repaired copy `fixed.jpg` can be opened without problems.

**Related Issues:**
- https://github.com/photoprism/photoprism/discussions/2462
- https://github.com/golang/go/issues/45902
- https://github.com/golang/go/issues/40130

### PR DESCRIPTION
**Pull request by AlD**
Fixes #2463

### DIFF
--- a/internal/thumb/open_jpeg_test.go
+++ b/internal/thumb/open_jpeg_test.go
@@ -17,8 +17,14 @@ func TestOpenJpeg(t *testing.T) {
 		}
 	})
 	t.Run("testdata/broken.jpg", func(t *testing.T) {
-		if _, err := OpenJpeg("testdata/broken.jpg", 0); err == nil {
-			t.Error("unexpected EOF while decoding error expected")
+		img, err := OpenJpeg("testdata/broken.jpg", 0)
+
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if img == nil {
+			t.Error("img must not be nil")
 		}
 	})
 	t.Run("testdata/fixed.jpg", func(t *testing.T) {
--- a/internal/thumb/open_jpeg.go
+++ b/internal/thumb/open_jpeg.go
@@ -1,58 +1,90 @@
 package thumb
 
 import (
+	"bytes"
 	"fmt"
 	"image"
+	"io"
 	"os"
 	"path/filepath"
+	"syscall"
 
 	"github.com/disintegration/imaging"
+	"github.com/mandykoh/prism/meta"
 	"github.com/mandykoh/prism/meta/autometa"
+	"golang.org/x/sys/unix"
 
 	"github.com/photoprism/photoprism/pkg/clean"
 	"github.com/photoprism/photoprism/pkg/colors"
 )
 
+var (
+	EOI = []byte{0xff, 0xd9}
+)
+
+func decode(reader io.Reader, logName string) (md *meta.Data, img image.Image, err error) {
+	// Read color metadata.
+	md, imgStream, err := autometa.Load(reader)
+
+	if err != nil {
+		log.Warnf("thumb: %s in %s (read color metadata)", err, logName)
+		img, err = imaging.Decode(reader)
+	} else {
+		img, err = imaging.Decode(imgStream)
+	}
+	return md, img, err
+}
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: internal/thumb/open_jpeg_test.go [hunk 1]
```diff
@@ -17,8 +17,14 @@ func TestOpenJpeg(t *testing.T) {
 		}
 	})
 	t.Run("testdata/broken.jpg", func(t *testing.T) {
-		if _, err := OpenJpeg("testdata/broken.jpg", 0); err == nil {
-			t.Error("unexpected EOF while decoding error expected")
+		img, err := OpenJpeg("testdata/broken.jpg", 0)
```
ValXp, line 20: I see two possible test cases missing:
- When an image is broken in such a way that recovery is attempted but the recovery itself fails (line 82). This could be due to stat failing or mmap failing (that could actually be two different test cases, but it might be tricky to test separately)
- When an image is broken in such a way that recovery is done but opening it later fails (line 85)

AlD: It's not like the tests currently have this level of coverage. Color profile handling, rotation etc. are also not tested.
That said, I'm happy to improve the test coverage here, but I'd need a sign-off from maintainer side on the general approach first.

lastzero: We do have metadata tests in the internal/meta package. The thumbs package is used by the photoprism package, which should also have resolution and orientation related tests.

Now, if we actively change/fix/update JPEG files, it makes sense to me that this requires additional testing beyond what currently exists?

AlD: > We do have metadata tests in the internal/meta package. The thumbs package is used by the photoprism package, which should also have resolution and orientation related tests.

I see, thanks for pointing that out.

> Now, if we actively change/fix/update JPEG files, it makes sense to me that this requires additional testing beyond what currently exists?

Absolutely. I was just waiting to get indication that this approach is something you'd consider doing at all, but I'll take your comment as that now :)
Will try to add the tests as soon as I find some time.

lastzero:
My apologies for not getting back to you sooner! This looks good at first sight. One potential issue could be higher memory usage when a file must be "repaired", if I understand it correctly?

maxime1992:
Thanks a lot @AlD for the PR! This issue has been crippling my workflow with photoprism where I run a check and fix command on Linux first on all photos before uploading... That'll be tremendously helpful :heart: :pray: !

lastzero:
@maxime1992 Could you please test it again with our latest preview build? Thank you very much! 🙏 

If you still have problems with some JPEG files, feel free to share them with us so we can fix the remaining issues for you:

- https://www.photoprism.app/contact#file-samples

Note that we will be on vacation for the first two weeks of July, so responding to emails may take longer than usual.

maxime1992:
@lastzero I just tried the preview with the [image I shared](https://github.com/photoprism/photoprism/issues/3363#issue-1693771256) in my original issue and it worked straight away! Amazing :)

maxime1992:
@lastzero I see you added the label "please-test". Not sure you've seen my message above but I did test it and can confirm it works :heavy_check_mark:

lastzero:
I've actually removed the label :) Thanks for testing!

Issue #2228, opened 2022-04-06
Videos: Improve hardware transcoding on Synology NAS devices


sunbeam60:
On Synology NAS devices (and [possibly other setups](https://github.com/linuxserver/docker-plex/issues/207), /dev/dri/renderD128 has permissions 'u=rw,g=,o=' and is owned by user "videodriver".

This configuration prevents hardware transcoding using h264_qsv as the VAAPI device cannot be unless the Synology user runs the docker package as root:

> /usr/bin/ffmpeg -qsv_device /dev/dri/renderD128 -i /photoprism/originals/2022/02/27/IMG_8370.MOV -c:a aac -vf format=rgb32 -c:v h264_qsv -vsync vfr -r 30 -b:v 25M -maxrate 25M -f mp4 -y /photoprism/storage/sidecar/2022/02/27/IMG_8370.MOV.avc"
> [SNIP]
> Failed to create a VAAPI device

If the file permissions for /dev/dri/renderD128 inside the container is set to 'u=rw,g=rw,o=rw', PhotoPrism's invocation of ffmpeg with h264_qsv succeeds.

This is a common issue for docker containers using hardware transcoding and it seems the [common pattern it to iterate across the dev/dri/render devices](https://github.com/linuxserver/docker-jellyfin/pull/112/files), ensuring groups have rw rights and adding whatever group the docker's user belongs to as group owners.

lastzero:
Do you know what ID the user videodriver has?

lastzero:
As described in the [release notes](https://docs.photoprism.app/release-notes/), the upcoming preview build includes default user accounts for common user ID ranges (50-99, 500-549, 1000-1099) that belong to the groups `photoprism` (1000), `video` (44), and `render` (109):
- If this does not meet your requirements, we can add other users/groups if you provide us with the specific IDs
- I'd rather not change the permissions for the host device inside a guest container if we can find another solution

sunbeam60:
> Do you know what ID the user videodriver has?

uid=937(videodriver) gid=937(videodriver)
... on my Synology. Whether that's persistent across Synology devices, I shan't say.

Apologies for late response - been away for Easter. Sounds like that's not part of the default range you've proposed.

lastzero:
No, I'm bad at guessing IDs correctly ;)

Now also spending time with my family...

lastzero:
If you find a way to open a terminal to execute commands in the container, you could try this manually first to save us some time / reduce feedback loops.

[...]

PR #2730, opened 2022-09-22
Docker: Switch from gosu to setpriv in entrypoint.sh

aaronkollasch:
## Description

This PR switches the su binary in entrypoint.sh from [gosu](https://github.com/tianon/gosu) to [setpriv](https://manpages.debian.org/bullseye/util-linux/setpriv.1.en.html) (included in Debian). This enables the following improvements:
- Supplementary groups are not lost if `PHOTOPRISM_UID` and `PHOTOPRISM_GID` are set. Supplementary groups are required for hardware transcoding on many devices. (#2228)
- Inherited capabilities from Docker root are no longer passed to `photoprism` if `PHOTOPRISM_UID` is set. (This has no effect in typical use, as CapPrm and CapEff are cleared when changing to unprivileged user in Docker.)
- No longer depends on gosu; removes unnecessary binary from images

`setpriv` has been available in the `util-linux` package since Debian 10.0 [Buster](https://packages.debian.org/buster/amd64/util-linux/filelist) and Ubuntu 20.04 [Focal](https://packages.ubuntu.com/focal/amd64/util-linux/filelist), so it is present in all supported Docker images. It is also listed as an alternative on the gosu [README](https://github.com/tianon/gosu/blob/cc6a15501b77eaa6dfb0a36b783ab4cff79f3405/README.md#setpriv).

### Examples
Before (gosu):
```console
root@220901-bullseye:/photoprism$ gosu "${PHOTOPRISM_UID}" id
uid=1032(user-1032) gid=1032(group-1030) groups=1032(group-1030),33(www-data),44(video),105(davfs2),109(renderd),115(render),937(videodriver),1000(photoprism)

root@220901-bullseye:/photoprism$ gosu "${PHOTOPRISM_UID}:${PHOTOPRISM_GID}" id
uid=1032(user-1032) gid=100(users) groups=100(users)
# setting the GID clears supplementary groups
```

After (setpriv):
```console
root@220901-bullseye:/photoprism$ /usr/bin/setpriv --reuid "${PHOTOPRISM_UID}" --regid "$(/usr/bin/id -g ${PHOTOPRISM_UID})" --init-groups id
uid=1032(user-1032) gid=1032(group-1030) groups=1032(group-1030),33(www-data),44(video),105(davfs2),109(renderd),115(render),937(videodriver),1000(photoprism)

root@220901-bullseye:/photoprism$ /usr/bin/setpriv --reuid "${PHOTOPRISM_UID}" --regid "${PHOTOPRISM_GID}" --init-groups id
uid=1032(user-1032) gid=100(users) groups=100(users),33(www-data),44(video),105(davfs2),109(renderd),115(render),937(videodriver),1000(photoprism),1032(group-1030)
# supplementary groups are preserved
```

## Acceptance Criteria

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,7 +98,6 @@ install:
 	rm -rf --preserve-root $(DESTDIR)/include
 	(cd $(DESTDIR) && mkdir -p bin sbin lib assets config config/examples)
 	./scripts/build.sh prod "$(DESTDIR)/bin/$(BINARY_NAME)"
-	GOBIN="$(DESTDIR)/sbin" go install github.com/tianon/gosu@latest
 	rsync -r -l --safe-links --exclude-from=assets/.buildignore --chmod=a+r,u+rw ./assets/ $(DESTDIR)/assets
 	wget -O $(DESTDIR)/assets/static/img/wallpaper/welcome.jpg https://cdn.photoprism.app/wallpaper/welcome.jpg
 	wget -O $(DESTDIR)/assets/static/img/preview.jpg https://cdn.photoprism.app/img/preview.jpg
--- a/docker/photoprism/armv7/Dockerfile
+++ b/docker/photoprism/armv7/Dockerfile
@@ -101,7 +101,6 @@ RUN echo 'APT::Acquire::Retries "3";' > /etc/apt/apt.conf.d/80retries && \
     echo 'APT::Install-Suggests "false";' > /etc/apt/apt.conf.d/80suggests && \
     echo 'APT::Get::Assume-Yes "true";' > /etc/apt/apt.conf.d/80forceyes && \
     echo 'APT::Get::Fix-Missing "true";' > /etc/apt/apt.conf.d/80fixmissing && \
-    mv /opt/photoprism/sbin/gosu /usr/local/sbin/gosu && \
     apt-get update && apt-get -qq upgrade && apt-get -qq install --no-install-recommends \
       libc6 ca-certificates sudo bash tzdata \
       gpg zip unzip wget curl rsync make nano \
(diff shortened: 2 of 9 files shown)


lastzero:
Excellent, I didn't know about `setpriv`! Although it is a small change (perfect), it might take a while before I can integrate it. We are currently working on multi-user support, which requires all our attention.

Since you seem to have a lot of experience with Unix privileges, feel free to check out the `entrypoint-init.sh` script as well. With this we install additional dependencies even if the container was not started as root - maybe there is a (better and safer) alternative:

https://github.com/photoprism/photoprism/blob/455028ad289146061e1f76221fd8f655fcb76d3f/scripts/dist/entrypoint.sh#L48-L53

aaronkollasch:
Awesome! In the meantime, I can just manually edit the entrypoint script and fallback to unaccelerated transcoding if it gets overwritten, no big deal.

I admit the sudo setenv nopasswd entry caught my eye when looking at the Docker container. I haven't found any issues `entrypoint-init.sh`, and the run-once nature of the make section reduces the potential attack surface, but it's definitely an important script to watch from a security standpoint. 
I can't think of a better alternative for installing additional dependencies + not starting as root that doesn't involve `docker-compose build` with a Dockerfile based on the release image. Also you'd still need root for the recursive `chown`, if you wanted to keep that.

Side note: while trying to get acceleration to work, I found that the Bookworm image doesn't have `intel-opencl-icd` available - is that just because it's a Debian testing release?

lastzero:
> Side note: while trying to get acceleration to work, I found that the Bookworm image doesn't have `intel-opencl-icd` available - is that just because it's a Debian testing release?

Yes, it seems so - but we are switching to Ubuntu Jammy anyway, since Debian Testing does not receive security updates. The main reason for using Bookworm was the newer versions of FFmpeg, RawTherapee and Darktable which fixed bugs. These are now also available on Ubuntu.

I'll see if I can merge your PR today, since my authentication enhancements have been pushed. They'll take some time to get released anyway - more than 10,000 lines of code.

Issue #2466, opened 2022-06-26
Login Security: Display a static logo page 1st for the base domain before rewriting the URL to /auth/login


n0hot:
**Description:**
I'm using a free Cloudflare firewall/gateway as a reverse proxy for my public facing Photoprism server that lives in my home. The Photoprism app is publicly accessible. Cloudflare works as a reverse proxy in that I do not have any forwarding rules or any ports open in my home firewall. The cloudflared daemon running on my server establishes the tunnel **from** my home **to** Cloudflare.  

**Solution to Secure the Login Page:**
I would like Photoprism to display an initial **logo** page with a "Login" button (no login fields). Then **only** after clicking the login button, would the URL be rewritten to the /auth/login page. 

The one click to login would provide me the ability to only **ALLOW** the login page to be displayed from IP addresses that I add to the Cloudflare firewall. I can **BLOCK** all others from getting to the login page with a regex using /auth/login/. Shared links would bypass the /auth/login and go directly to shared albums as usual. 

Thanks for considering this!  ...Scott

**More info:**  I have my Photoprism server setup with my own domain. So putting (e.g.) name.net into the URL bar routes thru the Cloudflare gateway to my server as https://name.net  Photoprism immediately rewrites the URL to https://name.net/auth/login **from the backend**. I cannot stop this login page from being displayed to the public. If I refresh the page, it's blocked by my rules at Cloudflare but by then it's too late. I can block all kinds of attacks on a login page with Cloudflare but it would be much better to not display it. A logo page would enable all login attacks to be blocked. Only my IP addresses would be allowed to reach the login page.

n0hot:
This doesn't seem like a very difficult thing to change. Anyone else worried about SQL injection attacks on the login page?

graciousgrey:
Thank you very much for your kind feedback. We will consider this at a later time, as we are extremely busy at the moment.

n0hot:
👍🏻

n0hot:
**Update:**  After a bit of analysis, I did come up with a solution to create a firewall rule that blocks the POST of the /auth/login page. That effectively mitigates any attempts to login from IPs other than then ones I allow.

lastzero:
Note that the frontend as such is not critical. For security reasons, you should instead block the REST API endpoint `POST /api/v1/session`. That's where sessions are ultimately created, not in the UI. It only communicates with the API.

n0hot:
Thank you.

Issue #840, opened 2021-01-06
UI: Prefix routes with /library/... so that 404 can be returned for non-existing assets


benmccann:
During development I was requesting a JS file that did not exist. However, it took me a very long time to figure this out because the browser was returning a 200. I thought I had Babel configured wrong or something because it kept telling me I had a malformed JS file. In reality it was returning HTML inside my JS

lastzero:
Routing is performed on the client. There is no simple way the server may know what routes exist in JS - and we can't afford to add much more complexity right now.

We may of course implement a 404 page in JS saying "No found, click here to return to home" instead of redirecting to home straight away. Not sure how much value this provides to our users though.

benmccann:
Ah, that makes sense. While we can't do it for a generic route, we might be to implement a 404 just for non-existent `.js` and `.css` files

benmccann:
It looks like this would be handled here:

https://github.com/photoprism/photoprism/blob/a747cca962fd8ea77ad81e6728b2d16e3ae8f60b/internal/server/routes.go#L151

lastzero:
Nobody complained so far, so I'll move ahead and close this as done!

PR #2782, opened 2022-10-14
Fix scrollpos after returning from download in iOS pwa mode

heikomat:
- This PR improves upon the bugfix from https://github.com/photoprism/photoprism/pull/2648
  - This improved fix now instantly puts the user at the correct scrollPos without any unnecessary reloads
- It also fixes all the related bugs found by @graciousgrey [here](https://github.com/photoprism/photoprism/issues/2597#issuecomment-1276390493)
  - It does so by implementing a general solution to the underlying ios-pwa-scrollpos-restore-after-download-bug that all the mentioned bugs share
  
As it turns out, iOS in PWA mode uses something called bfcache (backward-forward-cache) to cache the state of the whole webpage before navigating to the fullscreen-download-page, and restores the whole page from that cache after returning to the website.

The regular scrollpos-restore-mechanism found in `app.js` just doesn't support that bfcache-scenario. To fix this, i made the regular scrollpos-restore-mechanism compatible with bfcache scnearios by
- detecting them using `pageshow` and `pagehide` events
- remembering the current scrollpos when the page is about to be cached
- providing the old scrollpos to the restore-mechanism upon page-restore-from-cache.

To make it fully work i also skipped resetting `lastFilter` in the `$route`-watcher of album-photos and search-view. That way the search-method detects that the filter is unchanged and skips the search, thus **not** resetting the scrollpos to `0,0`. i'm not 100% sure why the `lastFilter` was always being reset on every `$route` change before and hope that that doesn't break anything.

I expect the potential for new bugs caused by these fixes to otherweise be very small, because other than ios-pwa-fullscreen-downloads, there is barely any scenario where `pageshow` or `pagehide` are triggered, especially with `event.persisted === true`.

Acceptance Criteria:


lastzero:
The reason `$route()` currently resets `this.lastFilter` and other component variables is that VueJS automatically reuses an existing component instance when the next route uses the same type. So, for example, if you navigate from `/archive` to `/browse` or `/review`. 

Otherwise, you might see pictures that require a review when you navigate back or forward to `/browse`. I'm not 100% sure without testing if `lastFilter` is an exception here that doesn't necessarily need to be reset, but in general it is required.

Note that with my next commit, I will change the frontend routes to be prefixed with `/library` in order to implement the following requests:
-  #840 
-  #2466.

heikomat:
@lastzero then maybe a solution could be to only reset lastFilter when the routeName changes. Do you think that could work?

lastzero:
Might work!? Current implementation was meant as a workaround until we find a better solution. You could force the router to always reload everything, but that's hurting performance.

heikomat:
Will Test it later today, by first checking if the issues you mention happen for me when not resetting lastFilter at all, and then testing if these issues dissaperar when only resetting on routName-change.

If that works for both cases (the issues you mentioned and the returning from fullscreen-Download) then we have a winner

lastzero:
I think it is used to figure out if the cached results need to be replaced (otherwise they can just be displayed or an offset can be used to retrieve more results from the server). Unfortunately I have no idea how this would work in combination with Apple's caching.

lastzero:
Make sure you also check the raw request data and not just what is displayed in the UI, as I recently added server-side logic to prevent content in review from being displayed by accident and tampering. This is part of our security enhancements for multi-user support.

heikomat:
@lastzero your hint with why the lastFilter was reset was completely right, and it looks like my idea with the `routeName` works:
| scenario | switching between `/browse` and `/search` | returning from cached browser-state |
| --- | --- | --- |
| **current develop:** `lastFilter` is reset on **every** `$route`-change | ✅ | ❌ a search is triggerd, resetting the scroll-pos |
| **local test:** `lastFilter` is **never** reset | ❌ search results don't change, because no search is triggered | ✅ |
| **this PR until 5 minutes ago:** `lastFilter` is only reset if `this.filter` changes | ❌ search results don't change, because no search is triggered | ✅ |
| **this PR now:** `lastFilter` is only reset if `routeName` changes | ✅ | ✅ |

I still need to check wether the raw-request data is fine, but i have a guess that they probably are.
The only thing that changed is **when** the `prevent Search-Api-call because filter is unchanged`-early-exit triggers. if the search is not prevented, then it is executed with the same parameters as before this PR.

As for the cached-and-restored state of the website: If i understood it correctly, from the websites point of view, it is 100% as if the user had never left. The whole in-memory-state of the website gets cached and restored as if nothing happened at all. That is also why components don't rerender upon return

lastzero:
Excellent work, as always! We should test it again, but it sounds plausible that it can work this way, since the configuration of the page component depends mainly on the route, e.g. which static search filters to use.

Dynamic filters and other settings, on the other hand, should be restored as they were by the URL query parameters.

There are two special cases to keep in mind:

1. the navigation contains links to "pages" that are actually just the regular search, but with additional search filters to serve as an example, e.g. [Monochrome](https://demo.photoprism.app/browse?q=mono%3Atrue%20quality%3A3%20photo%3Atrue)
2. settings like the view type are also stored in localStorage to remember the last used value

Let me know when you want this to be merged! 👍

heikomat:
> Excellent work, as always! We should test it again, but it sounds plausible that it can work this way, since the configuration of the page component depends mainly on the route, e.g. which static search filters to use.
> 
> Dynamic filters and other settings, on the other hand, should be restored as they were by the URL query parameters.
> 
> There are two special cases to keep in mind:
> 
>     1. the navigation contains links to "pages" that are actually just the regular search, but with additional search filters to serve as an example, e.g. [Monochrome](https://demo.photoprism.app/browse?q=mono%3Atrue%20quality%3A3%20photo%3Atrue)
> 
>     2. settings like the view type are also stored in localStorage to remember the last used value
> 
> 
> Let me know when you want this to be merged! 👍

just tested these special cases, and it looks like they work just fine. But that makes sense!
`lastFilter` is now only reset on route change. and clicking on `monochrome` doesn't change the route, but it DOES change the filter, so a new search is triggered (because `this.filter` is now !== `this.lastFilter`).

Thinking about it in a different way: if the route is unchanged, and the filter is unchanged, then the results should probably also not change!
- If the route changes, the filter is reset -> new search
- if the filter changes it is !== lastFilter -> new seach

Do you know of any cases where a new search needs to be triggered upon `$route`-change even though the routeName and filter are unchanged?

heikomat:
@lastzero  From my point of view this can be merged, but maybe we should wait for feedback from @graciousgrey. She was the one who found the original bugs this fixes, and if i recall correctly, she wanted to test these changes anyway.

lastzero:
Sorry for the conflicts! :rofl:

heikomat:
Can't look at them right now. I'll quickly resolve the conflicts once I'm at my laptop

lastzero:
Renamed the pages folder to page for consistency.

heikomat:
strange that github showed conflicts there. i was able to merge cleanly, because git detected that the files were just renamed

lastzero:
Thanks! I'll merge this now so we can test it on our demo and the temporary `:unstable` image we are currently using internally until a new development preview is available (it should be more stable than we can guarantee at the moment).

graciousgrey:
Just tested all cases again on different devices/browsers/pwa. Works perfect :heart:

Issue #1439, opened 2021-07-20
Video: Cannot close player in Firefox when plugin is installed


itsthejoker:
System info:

* Ubuntu 21.04
* 32GB RAM
* Ryzen 7 2700X
* Firefox 90.0

When launching a video on the demo site (https://demo.photoprism.org/browse?view=mosaic&order=newest&public=true&quality=3) I cannot exit the video player on Firefox. Clicking around the video generates an error in the console but does nothing visually. I can close the video as expected when using Chromium.

Error:

```
TypeError: e.pause is not a function    app.js:2:2368793
    stop https://demo-cdn.photoprism.org/static/build/app.js?1f68e8e1:2
    onClose https://demo-cdn.photoprism.org/static/build/app.js?1f68e8e1:2
    click https://demo-cdn.photoprism.org/static/build/app.js?1f68e8e1:2
    Ue https://demo-cdn.photoprism.org/static/build/app.js?1f68e8e1:2
    n https://demo-cdn.photoprism.org/static/build/app.js?1f68e8e1:2
    _wrapper https://demo-cdn.photoprism.org/static/build/app.js?1f68e8e1:2

    Ge https://demo-cdn.photoprism.org/static/build/app.js?1f68e8e1:2
    $e https://demo-cdn.photoprism.org/static/build/app.js?1f68e8e1:2
    je https://demo-cdn.photoprism.org/static/build/app.js?1f68e8e1:2
    Ue https://demo-cdn.photoprism.org/static/build/app.js?1f68e8e1:2
    n https://demo-cdn.photoprism.org/static/build/app.js?1f68e8e1:2
    _wrapper https://demo-cdn.photoprism.org/static/build/app.js?1f68e8e1:2
```

graciousgrey:
Just tested it here (Linuxmint, Firefox 90.0) and it works fine. 

Is anyone else observing this error?

rickysarraf:
No problem here either. Running PhotoPrism `develop` branch + Firefox 90.0

lastzero:
I've noticed a similar error in Firefox last year, triggered by mouse over playback on live photos. We fixed it, but maybe there's still an edge case left?

lastzero:
@itsthejoker I'm not sure how to trigger this error. Works for me on our demo using the latest Firefox on macOS and Linux. If you can reproduce it, can you please describe what exactly you're doing, or share a screen recording?

itsthejoker:
[...]

PR #2949, opened 2022-11-28
Videos: Improve player compatibility with browser plugins

ecker00:
When using the extension https://github.com/codebicycle/videospeed it's no longer possible to close the video player. As it grabs the wrong element. Making a more robust approach to always select the correct element.

### DIFF
--- a/frontend/src/component/video/player.vue
+++ b/frontend/src/component/video/player.vue
@@ -82,7 +82,7 @@ export default {
   },
   methods: {
     videoEl() {
-      return this.$el.childNodes[0];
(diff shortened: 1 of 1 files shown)


ecker00:
Fixes #1439

lastzero:
Did you test with Chrome as well? Probably should work, just to be sure.

ecker00:
Yes. I'm using the plugin in both Firefox and Chrome, same behavior and fix. 👍

lastzero:
Thank you very much!

PR #2991, opened 2022-12-12
Auth: Allow public WebDAV access without password

chain710:
SingleUser use case:

Home lan users only, don't want password. But authmode=public also disable webdav

Changes:
- No login required when auth_mode=singleUser(actually login with empty name&password, so create session api respond with built-in admin user)
- Account tab hidden
- Webdav enabled

Acceptance Criteria:

### DIFF
--- a/frontend/src/common/config.js
+++ b/frontend/src/common/config.js
@@ -570,12 +570,20 @@ export default class Config {
     return [];
   }
 
+  isAuthRequired() {
+    return this.values && this.values.authMode === "password";
+  }
+
   isPublic() {
     return this.values && this.values.public;
   }
 
   isDemo() {
     return this.values && this.values.demo;
+  }
+
+  isSingleUser() {
+    return this.values && this.values.authMode === "singleUser";
   }
 
   isSponsor() {
--- a/frontend/src/component/navigation.vue
+++ b/frontend/src/component/navigation.vue
@@ -530,7 +530,7 @@
           </v-list-tile-content>
         </v-list-tile>
 
-        <v-list-tile v-show="auth && !isPublic && $config.feature('settings')" class="p-profile" @click.stop="onAccount">
+        <v-list-tile v-show="auth && isAuthRequired && $config.feature('settings')" class="p-profile" @click.stop="onAccount">
           <v-list-tile-avatar size="36">
             <img :src="userAvatarURL" :alt="accountInfo" :title="accountInfo">
           </v-list-tile-avatar>
@@ -549,7 +549,7 @@
           </v-list-tile-action>
         </v-list-tile>
 
-        <v-list-tile v-show="isMini && auth && !isPublic" class="nav-logout" @click.stop.prevent="onLogout">
+        <v-list-tile v-show="isMini && auth && isAuthRequired" class="nav-logout" @click.stop.prevent="onLogout">
           <v-list-tile-action :title="$gettext('Logout')">
             <v-icon>power_settings_new</v-icon>
           </v-list-tile-action>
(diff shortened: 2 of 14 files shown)


lastzero:
👉 https://www.zdnet.com/article/hacker-ransoms-23k-mongodb-databases-and-threatens-to-contact-gdpr-authorities/

Note we have no control over where our software is installed, so we see a big danger offering this to our inexperienced user base. See above for an example of what could happen.

chain710:
> 👉 https://www.zdnet.com/article/hacker-ransoms-23k-mongodb-databases-and-threatens-to-contact-gdpr-authorities/
> 
> Note we have no control over where our software is installed, so we see a big danger offering this to our inexperienced user base. See above for an example of what could happen.

I agree. And I also think this option is useful to experienced user 
Maybe add a global unsafe option? like PHOTOPRISM_ENABLE_UNSAFE_FEATURES=singleUser,feat1,feat2
Or enable it only if bind localhost address?

lastzero:
Would using client side SSL certificates be an option? That way, there is no danger and you need no password.

chain710:
Client side SSL certificates means every client that accesses the website must have the certificate installed. Right?
If so, the operation may be cumbersome. Besides it still need server side patch, like auth_mode=client_ssl_cert

chain710:
By default, the singleUser auth mode is disabled. Users must read the documentation before using it. Perhaps we can emphasize the risks of this option in documentation

alxjsn:
I've been using PhotoPrism as a single user within my own LAN with not password from the start and the change that forced WebDAV off was unexpected behavior for me. I think secure defaults is helpful to many people, but the option to turn on just WebDAV without enabling a whole auth system would be nice.

chain710:
> I've been using PhotoPrism as a single user within my own LAN with not password from the start and the change that forced WebDAV off was unexpected behavior for me. I think secure defaults is helpful to many people, but the option to turn on just WebDAV without enabling a whole auth system would be nice.

That's exactly the use case this PR try to solve

Are you using `auth_mode=public`? Because maybe that's the only way to use photoprism with no password

chain710:
Update: 

@lastzero  How about `Auth Proxy`, it is widely adpoted in many projects, is it appropriate solution for this use case?

> Auth Proxy allows to authenticate a user using an HTTP header provided by an external authentication service

For example:
1. Authentication provided by authelia/keycloak
2. Nginx pass `X-Auth-User` header to photoprism which comes from auth provider
3. photoprism lookup user in database and return session

reference:
- https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/auth-proxy/
- https://github.com/gogs/gogs/issues/165
- https://github.com/miniflux/v2/pull/570

lastzero:
@chain710 After weighing all the pros and cons, we cannot merge this. Note that any software you have installed can perform a simple port scan, even on a private network, to find the unprotected file server and upload malicious files to it. If this is exploited, for example due to a misconfiguration or misunderstanding of how the software works, we might have to release an emergency patch and get an official CVE assigned for a major vulnerability.

In comparison, entering your password once is only a minor inconvenience for you. If you want to avoid an expired session when using the web interface, you can increase `PHOTOPRISM_SESSION_MAXAGE` and `PHOTOPRISM_SESSION_TIMEOUT`:

- https://docs.photoprism.app/getting-started/config-options/#authentication

In practice, it also makes little difference whether you send an `X-Auth-User` header or `Authorization` for the currently implemented WebDAV Basic Auth (if you use HTTPS so your password is not exposed):

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization

Of course, we will add more authentication options over time when we get a chance.

**Thank you for your support and understanding!**

PR #2925, opened 2022-11-20
allow to use make with docker-compose

alexislefebvre:
The `Makefile` doesn't work with the current LTS version of Ubuntu:

```
$ lsb_release --all
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.1 LTS
Release:	22.04
Codename:	jammy

$ docker --version
Docker version 20.10.12, build 20.10.12-0ubuntu4

$ make docker-build
docker compose pull --ignore-pull-failures
unknown flag: --ignore-pull-failures
See 'docker --help'.

Usage:  docker [OPTIONS] COMMAND

A self-sufficient runtime for containers

[…]

Run 'docker COMMAND --help' for more information on a command.

To get more help with docker, check out our guides at https://docs.docker.com/go/guides/

make: *** [Makefile:297 : docker-build] Erreur 125
```

So this PR add the ability to use `docker-compose` instead of `docker compose`.

Acceptance Criteria:

- [x] **Features and enhancements are fully implemented** so that they can be released at any time without additional work
- [x] **Automated unit and/or acceptance tests have been added** to ensure the changes work as expected and to reduce repetitive manual work
- [x] **User interface changes are fully responsive** and have been tested on all major browsers and various devices
- [x] Database-related changes are compatible with SQLite and MariaDB
- [x] Translations have been / will be updated (specify if needed)
- [x] Documentation has been updated: https://github.com/photoprism/photoprism-docs/pull/136
- [x] Contributor License Agreement (CLA) has been signed

PR #3044, opened 2022-12-31
Update Makefile

soupdiver:
Fix docker-compose command

`docker compose` -> `docker-compose`

### DIFF
--- a/docker/demo/Makefile
+++ b/docker/demo/Makefile
@@ -1,15 +1,15 @@
 .PHONY: all restart terminal logs;
 all: restart
 restart:
-	docker compose pull demo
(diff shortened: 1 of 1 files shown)


lastzero:
See:
- #2925

soupdiver:
> See:
> 
> * [allow to use make with docker-compose #2925](https://github.com/photoprism/photoprism/pull/2925)

I see.
Is the auto detection only wanted in the root `Makefile` and not under `docker/...`?

lastzero:
Since "docker-compose" is deprecated and may no longer work, either the "docker compose" command should be used or there should be reliable auto-detection.

PR #3146, opened 2023-01-26
added 5 languages to country names (AD to PK)

dvpfig:
Added translations to country names in german, spanish, italian, portuguese, swedish. From AD:Andorra until PK:Pakistan.

Acceptance Criteria:

- [ ] **Features and enhancements are fully implemented** so that they can be released at any time without additional work

### DIFF
--- a/pkg/txt/resources/countries.txt
+++ b/pkg/txt/resources/countries.txt
@@ -4,18 +4,26 @@ FR:Frankreich
 FR:Francia
 FR:França
 FR:Frankrike
-SK:Bratislava
+FR:Chamonix
+FR:Réunion
+FR:Strassburg
+FR:Strasbourg
+FR:Straßburg
+FR:Paris
 SK:Slovakia
+SK:Slovaquie
 SK:Slovacchia
 SK:Eslovaquia
 SK:Eslováquia
 SK:Slovakien
+SK:Bratislava
 AT:Austria
-AT:Hallstatt
-AT:Wörthersee
 AT:Österreich
+AT:Autriche
 AT:Áustria
 AT:Österrike
+AT:Hallstatt
+AT:Wörthersee
 AT:Vienna
 AT:Wien
 AT:Salzburg
@@ -23,14 +31,15 @@ AT:Sölden
 AT:Klagenfurt
 AT:Innsbruck
 HU:Hungary
+HU:Hongrie
 HU:Ungheria
 HU:Hungría
 HU:Hungria
 HU:Ungern
 GE:Georgia Republic
-GE:Georgia
-GE:Geórgia
-GE:Georgien
+GE:République-de-Géorgie
+GE:República-da-Geórgia
+GE:Georgien Republik
 VA:Vatican City
 VA:Vatican
 VA:Vatikan
@@ -38,6 +47,7 @@ VA:Vaticano
 VA:Vatikanen
 PL:Poland
 PL:Polen
+PL:Pologne
 PL:Polonia
 PL:Polónia
 PL:Polônia
@@ -45,6 +55,7 @@ PL:Warsaw
 PL:Warschau
 KH:Cambodia
 KH:Kambodscha
+KH:Cambodge
 KH:Cambogia
 KH:Camboya
 KH:Camboja
@@ -53,21 +64,28 @@ LA:Laos
 LA:Vientiane
 LA:Lao Republic
 VN:Vietnam
+VN:Viêt Nam
 VN:Vietnã
 DK:Denmark
 DK:Dänemark
+DK:Danemark
 DK:Danimarca
 DK:Dinamarca
 DK:Danmark
 MM:Myanmar
+MM:Birmanie
+MM:Birmania
+MM:Birmânia
 CZ:Czechia
 CZ:Czech
 CZ:Tschechien
+CZ:Tchèque
 CZ:Repubblica Ceca
 CZ:República Checa
 CZ:Tjeckien
 CZ:Karlovy Vary
 CH:Switzerland
+CH:Suisse
 CH:Svizzera
 CH:Suiza
 CH:Suíça
@@ -85,12 +103,14 @@ CH:Lucerne
 CH:Luzern
 BE:Belgium
 BE:Belgien
+BE:Belgique
 BE:Belgio
 BE:Bélgica
 BE:Belgien
 SE:Sweden
 SE:Sverige
 SE:Schweden
+SE:Suède
 SE:Svezia
 SE:Suecia
 SE:Suécia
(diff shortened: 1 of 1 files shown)


lastzero:
Mexico vs New Mexico might become a problem.

dvpfig:
Correct that detecting the keyword "Mexico" in pictures of "New Mexico" creates ambiguity. 
For the current implementation "New Mexico" is not a keyword but two, and the only option is to have the state of New Mexico assigned if there is a keyword "NewMexico" (no space). 

This actually leads me to a question with wider impact that I had in mind earlier and again when attempting to make this country list more comprehensive: 

**Keywords are only single-words.** (at least I haven't seen so far in my database any multi-word keyword)
We see addresses such as "street professor john" producing the keywords: "street", "professor", "john". I have examples of this in other languages than english but I'm guessing it's the same in english. 
So the state of "New Mexico" would produce keywords "new" and "mexico", taking either of those keywords separately leaves too much of ambiguity to take a decision on which place the photo was taken. 

**Keywords context with multiple words.**
To solve these ambiguities, which will be many. Which of the two ways shall we expect PhotoPrism may take in the future?
1. take into account multiple single-words keywords to take a decision? (harder to implement, but a maybe better approach)
2. have keywords that consist of multiple words? (simpler to implement, but maybe not a proper definition of a keyword)

lastzero:
Much of that functionality was developed years ago. Unfortunately, I'm on my phone right now and out of office until Monday, so I can't look it up to present you with references to the code and inline or other developer docs...

The matching was never primarily about keywords, but about file and folder names in the first place. Now it's true that keywords can be extracted from file and folder names, so it may look like this list focuses on keywords and thus you wonder why keywords are single words, but this list of "well known location names" includes strings with spaces. In addition, it may even be that they are compared from longest to shortest, but not 100% about that anymore. What is clear is that it must only include well known names that are not ambiguous (so with a very high confidence, not necessarily 110%). Otherwise that can cause more work to be user than it might help.

dvpfig:
Thanks for the reply. 
I've checked the code, and I can confirm that my initial thinking about Keywords was incorrect. 

The code searches the text strings from photo title, file name, path. And not from keywords (of course, keywords are also created based on those sources). 
Then I also found that it only checks country names for 1 word or max 2 consecutive words. 

Below are the code that deals with this:

**1. photo_estimate.go** - checks if the country name (string) is present in the title or in the file name or in the path or in the orginal file name.  
https://github.com/photoprism/photoprism/blob/develop/internal/entity/photo_estimate.go
```go
[...]
	// Reset country.
	unknown := UnknownCountry.ID
	countryCode := unknown

	// Try to guess country from photo title.
	if code := txt.CountryCode(m.PhotoTitle); code != unknown && m.TitleSrc != SrcAuto {
		countryCode = code
	}

	// Try to guess country from filename and path.
	if countryCode == unknown {
		if code := txt.CountryCode(m.PhotoName); code != unknown && !fs.IsGenerated(m.PhotoName) {
			countryCode = code
		} else if code := txt.CountryCode(m.PhotoPath); code != unknown {
			countryCode = code
		}
	}

	// Try to guess country from original filename.
	if countryCode == unknown && m.OriginalName != "" && !fs.IsGenerated(m.OriginalName) {
		if code := txt.CountryCode(m.OriginalName); code != UnknownCountry.ID {
			countryCode = code
		}
	}

	// Set new country?
	if countryCode != unknown {
		m.PhotoCountry = countryCode
		m.PlaceSrc = SrcEstimate
		m.EstimatedAt = TimePointer()
		log.Debugf("photo: estimated country for %s is %s", m, clean.Log(m.CountryName()))
	}
[...]
```

**2. places.go** - function countrycode, checks if the string input matches any country name string with 1 word or with 2 consecutive words (but not longer), if so it returns the country code. 

https://github.com/photoprism/photoprism/blob/develop/pkg/txt/places.go
```go
[...]
var CountryWordsRegexp = regexp.MustCompile("[\\p{L}]{2,}")

// CountryCode tries to find a matching country code for a given string e.g. from a file oder directory name.
func CountryCode(s string) (code string) {
	code = UnknownCountryCode

	if s == "" || s == UnknownCountryCode {
		return code
	}

	words := CountryWordsRegexp.FindAllString(s, -1)

	for i, w := range words {
		if i < len(words)-1 {
			search := strings.ToLower(w + " " + words[i+1])

			if match, ok := Countries[search]; ok {
				return match
			}
		}

		search := strings.ToLower(w)

		if match, ok := Countries[search]; ok {
			code = match
		}
	}

	return code
}
[...]

```

dvpfig:
The limit of 2 words is for performance reasons, correct? 
If so, I'll remove the country names that have longer names. 

Based on this learning of the code to check for strings, I don't understand your concern about "New Mexico" vs "Mexico": 
- "New Mexico" will be correctly assigned as US, the two consecutive words match only one string of a country code (US). 
- "Mexico" on the other hand, matches the country code MX string (single word), and does not match the "New Mexico" string.

lastzero:
I'm on my phone, so I couldn't easily check if New Mexico also exists and/or run tests to see if it might be confused or not. However, it was important to me to raise awareness and explain potential issues as you are new to this.

dvpfig:
OK, Thanks for pointing it out the "New Mexico" vs "Mexico" ambiguity.

"New Mexico" is still not in the list, it wasn't there before my pull request and I started first by adding only variations of country names. 
But I do intend to continue with adding city names and some well-known regions/states within some countries, so I'll make sure to include all states of US.

I'll review the list for such ambiguities before proposing to merge, it's still a draft work-in-progress. 

Btw, this is quite manual work to be doing this modifications in the file, so let me know if you plan for a re-write of this functionality in the next couple of months and see this extension of functionality as not so useful. (I don't see it in the roadmap)

lastzero:
We mainly need to be careful not to match files wrongly as there is no easy way to undo this. So less is more if unsure (or no time to test). No major rewrite planned as of now. Thank you!

dvpfig:
I widen considerably the scope of keyword matching to countries in this file. It's a good tool. 

However, without a simpler way to correct or confirm the location estimate (based on these words taken from the folder path/name), the estimate is prone to too many errors. 

Even with a much narrower scope, the estimate is still prone to make very wrong estimates. 
A simple example is "Paris", such keyword can easily mean one of three things: 
* "Paris" as capital of France - that would be correct match of the tool. 
* "Paris, Texas" as a city in USA
* "Paris" used as name of a person, as in "Paris Hilton"
* "Restaurant Paris" used as a name of a restaurant somewhere. 

There are way too many examples of using a name of a place to mean something not related to the main use of the place. 

So, I think it would be really cool to extend this "Location Estimate", but wait until PhotoPrism has (at least) the feature of "Batch Edit" for country names in folders. i.e. to make it easy to correct wrong assumptions.

lastzero:
We are closing this PR as stalled. Thank you for your contribution!

Issue #3216, opened 2023-02-21
Query breaks in strict compilers, or Mariadb in only_full_group_by mode


dankasak:
#### 1. What is not working as documented?

Databases that are stricter than old MySQL / Mariadb throw an error on the query generated from:
 https://github.com/dankasak/photoprism/blob/develop/internal/config/client_config.go#L613

#### 2. How can we reproduce it?

Install into a MySQL compatible database in only_full_group_by mode and look at error log

Install into a MySQL compatible database in only_full_group_by mode and look at error log. Eg in TiDB:

[2023/02/15 21:42:01.175 +11:00] [WARN] [session.go:2188] ["compile SQL failed"] [conn=3747630407693107665] [error="[planner:1055]Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'photoprism.l.label_uid' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by"] [SQL="SELECT l.label_uid, l.custom_slug, l.label_name FROM `categories` JOIN labels l ON categories.category_id = l.id WHERE (l.deleted_at IS NULL) GROUP BY l.custom_slug ORDER BY l.custom_slug LIMIT 1000 OFFSET 0"]

#### 3. What behavior do you expect?

No SQL error

#### 4. What could be the cause of your problem?

SELECT l.label_uid, l.custom_slug, l.label_name FROM `categories` JOIN labels l ON categories.category_id = l.id WHERE (l.deleted_at IS NULL) GROUP BY l.custom_slug ORDER BY l.custom_slug LIMIT 1000 OFFSET 0

 ... should be:

SELECT l.label_uid, l.custom_slug, l.label_name FROM `categories` JOIN labels l ON categories.category_id = l.id WHERE (l.deleted_at IS NULL) GROUP BY l.custom_slug, l.label_uid, l.label_name ORDER BY l.custom_slug LIMIT 1000 OFFSET 0

#### 5. Can you provide us with example files for testing, error logs, or screenshots?

NA

#### 6. Which software versions do you use?

(a) PhotoPrism Architecture: AMD64, Build Number: docker pull from today

(b) Database Type & Version: TiDB

(c) Operating System Types & Versions: Linux

(d) Browser Types & Versions: Firefox

(e) Ad Blockers, Browser Plugins, and/or Firewall Software? NA

#### 7. On what kind of device is PhotoPrism installed?

[...]

PR #3217, opened 2023-02-21
Fix only_full_group_by compilation error in SQL - #3216

dankasak:
Acceptance Criteria:

- [X] Automated unit and/or acceptance tests are mandatory to ensure the changes work as expected and to reduce repetitive manual work
- [X] Frontend components must be responsive to work and look properly on phones, tablets, and desktop computers; you must have tested them on all major browsers and different devices
- [X] Documentation and translation updates should be provided if needed
- [X] In case you submit database-related changes, they must be tested and compatible with SQLite 3 and MariaDB 10.5.12+

### DIFF
--- a/internal/config/client_config.go
+++ b/internal/config/client_config.go
@@ -610,7 +610,7 @@ func (c *Config) ClientUser(withSettings bool) ClientConfig {
 		Select("l.label_uid, l.custom_slug, l.label_name").
 		Joins("JOIN labels l ON categories.category_id = l.id").
 		Where("l.deleted_at IS NULL").
-		Group("l.custom_slug").
+		Group("l.custom_slug, l.label_uid, l.label_name").
 		Order("l.custom_slug").
 		Limit(1000).Offset(0).
 		Scan(&cfg.Categories)


lastzero:
Thank you! Would this be the only change required for TiDB? I assume not?

Unfortunately we have very little time to test such changes, also performance wise. For now, it seems safest to add TiDB as a separate dialect and use a switch if needed.

Note that we have supported TiDB in the past. There just wasn't enough time to maintain it, and there hasn't been much demand for it either. So if we were to reintroduce TiDB support, this would require ongoing help with maintenance and testing...

dankasak:
Yes I understand the laborious task of testing changes. In this case, I guarantee you that there are no functional or performance impacts ( the "group by" on the 3 columns is implicit in the query, but should be explicitly listed in the "group by" clause ), but I get that you'd prefer to not have to take someone's word for it. And yes, there are some other changes required for seamless TiDB support. I've opened a discussion at: https://github.com/photoprism/photoprism/discussions/3218 and detailed my intention to complete and maintain TiDB support. I understand that you'd prefer to have these changes in a separate dialect ... in that case I could use some pointers as to what would be required. Perhaps we can continue in the above discussion?

lastzero:
I have merged this and built a new development preview for testing. Note that we can't add many more changes for the next release. However, this change seemed low risk to me, also based on your confirmation. Thank you!

PR #3222, opened 2023-02-21
CLI: Add `photoprism find` command to search the index for specific files

seeschloss:
This PR adds a CLI command called "search" that takes a search query as its only argument, in the same format as the search box in the UI, and lists all filenames found matching this search query.

For photos that have multiple files, all files are listed.

This is a rather small PR that does not touch any existing code.

As I said on the community chat I couldn't really make tests work for CLI commands, but this might not be a big problem. I looked around a bit but didn't find CLI-specific documentation that needed to be updated to add this new command.

Acceptance Criteria:

### DIFF
--- a/internal/commands/commands.go
+++ b/internal/commands/commands.go
@@ -65,6 +65,7 @@ var PhotoPrism = []cli.Command{
 	ShowCommand,
 	VersionCommand,
 	ShowConfigCommand,
+	SearchCommand,
 	ConnectCommand,
 }
 
--- a/internal/commands/search.go
+++ b/internal/commands/search.go
@@ -1,0 +1,56 @@
+package commands
+
+import (
+	"context"
+	"time"
+	"strings"
+
+	"fmt"
+
+	"github.com/urfave/cli"
+	"github.com/photoprism/photoprism/internal/form"
+	"github.com/photoprism/photoprism/internal/search"
+	"github.com/photoprism/photoprism/internal/entity"
+)
+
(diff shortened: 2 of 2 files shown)


alexislefebvre:
Could you please describe a use case? How would you use it? I wonder why use the CLI instead of the GUI where you can see the pictures.

lastzero:
From how I understand it, this is to use the output in scripts for automation.

seeschloss:
Yes, my main use case was deleting all photos from my archive containing one person:

`photoprism search person:Foo | xargs rm`

And now I also use it for periodically backing up specific albums.

lastzero:
Thank you very much! I will merge this now and may refactor it later.

lastzero:
@seeschloss My changes rename the "search" command to "find" and add support for additional output formats like Markdown and CSV. It is now also possible to limit the number of results (10k by default). In case no format is specified, only the filenames are returned as before. Will this work for you?

lastzero:
An updated preview build will be available for testing soon. The release notes have already been updated:

https://docs.photoprism.app/release-notes/#development-preview

seeschloss:
> @seeschloss My changes rename the "search" command to "find" and add support for additional output formats like Markdown and CSV. It is now also possible to limit the number of results (10k by default). In case no format is specified, only the filenames are returned as before. Will this work for you?

Sure, it's fine :) I really just implemented the bare minimum I needed, thanks for the improvements!

PR #3233, opened 2023-02-25
prevent rerendering of faces page on every typed letter

### LINKED ISSUE #3151: People: Entering names becomes slow with many faces tagged
#### 1. What is not working as documented?

I scanned my photo library and ended up with about 400 faces to tag.
I noted that the face tagging UI gets very sluggish sometimes. i.e. typing a letter doesn't show anything in the page for a few seconds. This seems to happen only when editing a name already added.

#### 2. How can we reproduce it?

- Import lots of faces
- Go to People > New
- Tag one person and press enter
- Try to edit that same person name
- The edits take a long time to reflect on the page

#### 3. What behavior do you expect?

For inputs to reflect immediatelly on the page while typing

#### 4. What could be the cause of your problem?

Maybe the rendering framework not being able to keep up with the input changes and the amount of faces to display

#### 5. Can you provide us with example files for testing, error logs, or screenshots?

https://user-images.githubusercontent.com/1005498/215434280-6fccd09c-1df5-48fb-b9e4-78eddea1df0c.mp4

It is not super obvious in the recording, but I type backspace several times and don't see a change for several seconds. Then the change appears suddenly.

#### 6. Which software versions do you use?

(a) PhotoPrism Architecture & Build Number: 
Intel
Build 221118-e58fee0fb

(b) Database Type & Version: MariaDB

(c) Operating System Types & Versions: Linux PopOS 22.04

(d) Browser Types & Versions: Firefox 108

#### 7. On what kind of device is PhotoPrism installed?

(a) Device / Processor Type: Intel Nuc i5

(b) Physical Memory & Swap Space in GB: 16GB

(c) Storage Type: HDD 240GB.

#### 8. Do you use a Reverse Proxy, Firewall, VPN, or CDN?

No

heikomat:
This Fixes https://github.com/photoprism/photoprism/issues/3151.

Setting and then editing a name of a face was slow, because the whole faces-page with all its components rerendered on every typed letter. This was because the face-model-name was bound directly to the input, therefore updating the model on every letter change.

With these changes, the model is only updated when the user actually commits the change (either by leaving the input or pressing enter).

The faces page still renders slow because it uses a LOT of vue-components (instead of native dom-elements), but speeding that up would require a more invasive change. With these less invasive changes, the rerender might still be slow, but because it only happens if something actually changes, the user-experience is improved.

Acceptance Criteria:

### DIFF
--- a/frontend/src/model/face.js
+++ b/frontend/src/model/face.js
@@ -117,15 +117,15 @@ export class Face extends RestModel {
     return Api.put(this.getEntityResource(), { Hidden: this.Hidden });
   }
 
-  setName() {
-    if (!this.Name || this.Name.trim() === "") {
+  setName(newName = this.Name) {
+    if (!newName || newName.trim() === "") {
       // Can't save an empty name.
       return Promise.resolve(this);
     }
 
     this.SubjSrc = src.Manual;
 
-    const payload = { SubjSrc: this.SubjSrc, Name: this.Name };
+    const payload = { SubjSrc: this.SubjSrc, Name: newName };
 
     return Api.put(Marker.getCollectionResource() + "/" + this.MarkerUID, payload).then((resp) => {
       if (resp && resp.data && resp.data.Name) {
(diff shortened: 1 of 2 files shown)


lastzero:
Thank you very much! I remember to have fixed the same issue in other places before.

PR #3244, opened 2023-03-03
Fix duration compare in photos location estimate

rconn01:
There is currently a bug in how durations are compared when determining if locations should be estimated again or not. As a result a photos location is not estimated every 7 days as the code intends. The code converts to number of hours as an int, but compares against a duration which ends up being that many nano seconds. This is best shown with a sample

The following [code](https://go.dev/play/p/PnvRypIMRbH)

```golang
package main

import (
	"fmt"
	"time"
)

func main() {
	p := fmt.Println
	var MetadataEstimateInterval = 24 * 7 * time.Hour

	now := time.Date(2023, 02, 28, 00, 30, 00, 00, time.UTC)
	tenDay := time.Date(2023, 02, 18, 00, 30, 00, 00, time.UTC)
	oneDay := time.Date(2023, 02, 27, 00, 30, 00, 00, time.UTC)

	p(now.Sub(oneDay) / time.Hour)
	p(now.Sub(tenDay) / time.Hour)

	if now.Sub(oneDay)/time.Hour > MetadataEstimateInterval {
		p("Division: One Day is > than 7")
	} else {
		p("Division: One Day is < than 7")
	}

	if now.Sub(tenDay)/time.Hour > MetadataEstimateInterval {
		p("Division: ten Days is > than 7")
	} else {
		p("Division: ten Days is < than 7")
	}

	p("----------------------")

	p(now.Sub(oneDay))
	p(now.Sub(tenDay))

	if now.Sub(oneDay) > MetadataEstimateInterval {
		p("Duration: One Day is > than 7")
	} else {
		p("Duration: One Day is < than 7")
	}

	if now.Sub(tenDay) > MetadataEstimateInterval {
		p("Duration: ten Day is > than 7")
	} else {
		p("Duration: ten Days is < than 7")
	}

}
```

**Produces:** 
```
24ns
240ns
Division: One Day is < than 7
Division: ten Days is < than 7
----------------------
24h0m0s
240h0m0s
Duration: One Day is < than 7
Duration: ten Day is > than 7
```

as you can see, dividing by `time.Hour` does not produce the expect result. Since `Time.Sub` returns a duration, this can simply be compared against the `MetadataEstimateInterval` duration

_Note:_ A result of this change is users libraries will start to be estimated more frequently than today, but based on the comments of the code it appears the intent was always to estimate every 7 days. If users don't want this behavior they can already disable the estimate feature 

**Acceptance Criteria:**

### DIFF
--- a/internal/entity/photo_estimate.go
+++ b/internal/entity/photo_estimate.go
@@ -64,7 +64,7 @@ func (m *Photo) EstimateLocation(force bool) {
 		return
 	} else if force || m.EstimatedAt == nil {
 		// Proceed.
-	} else if hours := TimeStamp().Sub(*m.EstimatedAt) / time.Hour; hours < MetadataEstimateInterval {
+	} else if hours := TimeStamp().Sub(*m.EstimatedAt); hours < MetadataEstimateInterval {
 		// Ignore if location has been estimated recently (in the last 7 days by default).
 		return
 	}
--- a/internal/entity/photo_estimate_test.go
+++ b/internal/entity/photo_estimate_test.go
@@ -75,6 +75,21 @@ func TestPhoto_EstimateLocation(t *testing.T) {
 		assert.Equal(t, UnknownCountry.CountryName, m2.CountryName())
 		assert.Equal(t, SrcAuto, m2.PlaceSrc)
 	})
+	t.Run("NotRecentlyEstimated", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


lastzero:
Seems you're right, MetadataEstimateInterval is defined as `24 * 7 * time.Hour`! Thank you :)

PR #3276, opened 2023-03-13
Added map for audio stream, too

ghostwheel:
It seems that if -map is used for video (`-map 0:v 0`) then a map is also needed for audio, otherwise no audio is produced.

Acceptance Criteria:


lastzero:
Any idea why it's not "0:a:0" when video is "0:v:0"?

ghostwheel:
Yes, actually `0:a:0` makes more sense than `0:a`. I had tried it, but it probably failed for some other reason. Best is actually `0:a:0?`, which won't fail if there is no audio.

PR #3283, opened 2023-03-14
Changed audio conversion to only use one stream

ghostwheel:
`0:a` means take all audio streams of first file. 
`0:a:0 `means first stream of first file. 
`0:a:0?` means first audio stream of first file if available.
This is a fix following #3276 

Acceptance Criteria:


lastzero:
Thanks! Hopefully this is correct now as building a new version takes hours every time :)

lastzero:
Do you have any sample files with multiple video streams for us? That would be really helpful.

ghostwheel:
No, sorry. I just had problems with no audio being produced. I guess that now that I know what `-map` means, I can try to join streams from two files into one (`-map 0:v:0 -map 1:v:0` I think.) I'll try that.

lastzero:
Just make sure this doesn't break everyone else's videos, because most people don't have multiple streams of everything... let us know when your changes are fully tested and ready to use :)

ghostwheel:
Me neither. I don't have videos with several different streams. I think for me the added parameter `-map 0:v:0` broke things, because none of my videos had sound then when played through photoprism. So I'd be happy with not having that parameter. But if it is there, then `-map 0:a:0?` is needed to also get audio.

lastzero:
See:
- #3284

PR #3360, opened 2023-05-03
ignore .git through .dockerignore

firefly2442:
Probably not worth the time for the CLA so if someone just wants to  add the line in and close this that is likely easiest.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -16,6 +16,7 @@
 .idea
 .DS_Store
 .env
+.git
 *.db
 *.db-journal
 Dockerfile


lastzero:
Wouldn't this break the build scripts as they try to detect the version based on the git metadata?

firefly2442:
If you're wanting to run git commands from within Docker I suspect it would.  This basically just prevents the `.git` folder from being loaded into the Docker context.  If you're unsure probably best to just close it.  For really large git repositories though this can definitely speed up Docker builds.  Cheers.

kkovaletp:
I don't know what exactly is going on during your build procedure, but in general running git commands inside docker is bad practice. If you need to pass some info from git to the build procedure, it is much better to do something like `docker compose -f my-docker-compose.yaml build --build-arg GIT_HASH=$$(git rev-parse --short HEAD) --pull my_image` - in this case, the variable `GIT_HASH` will be available during the docker build and can be accessed in Dockerfile like:
```
ARG GIT_HASH
yout_build_command -key ${GIT_HASH}
ENV GIT_HASH=${GIT_HASH} # In this way you declare the system environment variable so that your running build app (like NPM) can access it like any other environment variables
```

lastzero:
It's a unified build procedure that also works without Docker. So having it depend on passing a variable with Docker would be bad practice. While we continuously improve our processes, this hasn't been a pain point so far and thus doesn't have priority compared to the long list of tasks on our roadmap. These are features our users are waiting for.

kaiyou:
Also, the unified build process makes it fairly easy to use it on any kind of CI, including templated CI that knows nothing about the project but to run a Docker build on the proper dockerfile.

I would agree this being shady practice if `.git` metadata were retained in the output image or any intermediate layer, which is not the case here since all production images are nicely multistaged. The only harm I can think of is that changing some Git metadata without altering code will indeed prevent from using the Docker build cache, I can live with it.

lastzero:
Thank you very much for the suggestion! I'll close the PR since it looks like it might cause problems with our current build process.

PR #3365, opened 2023-05-03
Auth: Correct feature for account settings

kvalev:
Use the `account` feature to determine whether the account settings and login button should be shown (and not the `settings` feature, which has a different purpose).


lastzero:
Thank you very much! I'm not quite sure why we checked for settings instead of account. Maybe because account was added at a later time, after the navigation was already implemented.

PR #3436, opened 2023-05-29
Performance: Full virtualization of the mosaic view

heikomat:
This updates the pseudo-virtualization of the mosaic view with a real virtualization.

This increases scroll-performance yet again (less blinking, faster loading), while decreasing memory usage.
The effect should be especially noticeable when scrolling through a lot of images on a small screen.

The virtualization is implemented by measuring the size of the first Element using a ResizeObserver and assuming, that all other elements have the same size and no spacing inbetween.

This has not yet been tested on iOS devices yet.

~~There is a known bug on Firefox, that the number of columns is off by 1 on certain resolutions. I highly suspect that that is because of the non-integer widths of the columns on most screen-sizes. Because the position-calculations are simple JS, it's probably the size reported by the ResizeObserver. Will fix that later~~
[Fixed it](https://github.com/photoprism/photoprism/pull/3436/commits/e6e5bb9da30a8bc2029472754b4a7acdbfbe2b9e)

~~There is a known bug where often when opening a photo, and using the left/right arrows to switch between images, the wrong images are shown. This is strange, because the viewer opening logic isn't touched in this PR. Will try to find the cause and fix it soon.~~
[Fixed it](https://github.com/photoprism/photoprism/pull/3436/commits/72d0c7e94198d629d63b486f24336cbbfc46f44c)

**Performance comparison**
Scrolling through 1400 images on a smartphone-sized screen

**before**
performance-old

**after**
performance-new

Acceptance Criteria:


lastzero:
Looks amazing. We'll test it when are back in our office. Let us know when you think this is ready to be merged!

heikomat:
@lastzero will do.
[Fixed](https://github.com/photoprism/photoprism/pull/3436/commits/e6e5bb9da30a8bc2029472754b4a7acdbfbe2b9e) the Firefox bug.
~~but: using the left/right arrows in the photo viewer shows the wrong pictures? (which is interesting, because the viewer opening logic isn't touched in this PR)~~ **update:** [fixed](https://github.com/photoprism/photoprism/pull/3436/commits/72d0c7e94198d629d63b486f24336cbbfc46f44c)

Now the only things missing are
- ~~fixing the viewer-prev/next-bug~~ **update:** [fixed](https://github.com/photoprism/photoprism/pull/3436/commits/72d0c7e94198d629d63b486f24336cbbfc46f44c)
- more testing on more devices
- maybe some more benchmarking in different scenarios

heikomat:
update: viewer-prev/next-bug is fixed

lastzero:
My understanding is that this should first be tested on more devices, in particular iOS, before it's safe to be merged into develop?

heikomat:
> My understanding is that this should first be tested on more devices, in particular iOS, before it's safe to be merged into develop?

That is correct

heikomat:
I found a small bug with these changes where when you scroll to the bottom of the currently loaded pictures super fast, the infinite-scroller seems to not always recognize that the end of the list was reached. If this happens, the next batches are only loaded once you scroll the tiniest but upwards.

This does not happen when not scrolling super fast, presumably because the infinite scroller isn't missing any events in that case

lastzero:
When you are ready, I'd be happy to merge this as well.

heikomat:
Not yet. Let me fix this last bug mentioned [here](https://github.com/photoprism/photoprism/pull/3436#issuecomment-1627520821) first, and let me do some performance tests on different devices first.

Wjxfi:
Any updates?

heikomat:
> Any updates?

When i left the PR it was very close to be final. It was basically 99% done. Usable, but with one rare bug remaining.

Sadly, i have to admit that i moved on and i'm currently not contributing to photoprism.

lastzero:
@heikomat Thanks for letting us know! We'll try to find a way to finish this without your support.

Issue #3466, opened 2023-06-12
Video: Limit FFmpeg Transcoding Resolution


lukasdotcom:
**Describe what problem this solves and why this would be valuable to many users**

A lot of the time I am recording video in 4k, but on all of the clients I never actually need to watch the video at 4k and the sidecar file is then unnecessarily big and uses an unnecessary amount of bandwidth when played.

**Describe the solution you'd like**

For this purpose the environment variable "PHOTOPRISM_FFMPEG_SIZE" should be added to set the maximum resolution FFmpeg converts the video to, for example:

- `PHOTOPRISM_FFMPEG_SIZE=720`
- `PHOTOPRISM_FFMPEG_SIZE=3840`

lastzero:
@Lukasdotcom @theresagresch I wonder if it makes sense to only allow standard video resolutions to be specified? So HD, Full HD, Ultra HD, 2K and 4K. For example, few users will know that 4K actually means 3840 × 2160, and NOT a width of 4000 or 4096 pixels, as one might assume.

lukasdotcom:
We could either add cases to the original environmental variable to detect sizes like HD, HD+, UHD, 2K, and 4K and that they will cause certain predetermined resolutions to be used. Or you can only put in the more human readable names in that environmental variable and we create a new one that works like the original one called FFmpeg Resolution Advanced.

lukasdotcom:
Ill make a new pull request for either option depending on what you think is the better option.

lukasdotcom:
I just made the pull request based on the option where the FFMPEG_RESOLUTION variable can either use a standard one or an advanced user can specify a resolution down to the pixel #3549.

lastzero:
@Lukasdotcom Thanks a lot! I appreciate your PR but was prepared to refactor the config option myself today :)

lastzero:
@Lukasdotcom After checking all possible options, I decided to (re)use the default sizes defined in the thumbs package and adjust the config value to the next appropriate size. Supported sizes can be displayed by running the `photoprism show video-sizes` command in a terminal.

I also renamed `PHOTOPRISM_FFMPEG_RESOLUTION` to `PHOTOPRISM_FFMPEG_SIZE` so that the name of the option matches the image quality options e.g. `PHOTOPRISM_THUMB_SIZE` and `PHOTOPRISM_JPEG_SIZE`.

**It would be great if you could quickly review my changes and let us know if this works for you so we can release it this week. Thank you!**

[...]

PR #3498, opened 2023-06-24
Config: Add video transcoding resolution option

lukasdotcom:
This is a starting point for work on #3466. This update will add a configuration variable of PHOTOPRISM_FFMPEG_RESOLUTION that will let you set a maximum resolution for videos. This allows you to save disk space and more importantly bandwidth when watching videos in the browser.

Acceptance Criteria:


lastzero:
@Lukasdotcom Thank you very much for your contribution! This looks good to me at first sight. Of course, it should still be very well tested, as video transcoding is a core feature that users expect to work at all times. Let us know when you have questions or think it's ready to be merged so users can test it with our development preview build.

internal/ffmpeg/convert.go
```diff
@@ -25,6 +25,7 @@ func AvcConvertCommand(fileName, avcName string, opt Options) (result *exec.Cmd,
 			"-i", fileName,
 			"-movflags", "faststart",
 			"-pix_fmt", "yuv420p",
+			// "-vf", "scale='-2:trunc("+opt.Resolution+"/2)*2'",
```
lukasdotcom, line 28: Should I have the resolution limit also apply to animated images or only to videos?

lastzero: That's a great question we should actually ask our users...

lukasdotcom:
This should now work well enough for testing.
I just finished testing this with software encoding and apple's hardware encoding.

lastzero:
If you give me a thumbs up as confirmation, I'll be happy to merge your changes for further testing!

lukasdotcom:
👍

lastzero:
@Lukasdotcom Would it make sense to specify the resolution as the longest side, depending on the aspect ration, rather than the height only? For images, it's typically done that way, but for videos, I'm not 100% sure. Phones for example often show videos in portrait mode.

Feel free to make suggestions on this or provide a follow-up PR. We could also rename RESOLUTION to HEIGHT, for example.

lastzero:
With FFmpeg it seems to be possible to keep the aspect ratio and specify only the longest side:
```
ffmpeg -i input.jpg -vf scale=320:-1 output_320.png
```

Docs: http://trac.ffmpeg.org/wiki/Scaling#KeepingtheAspectRatio

lukasdotcom:
I think I'll make another PR that uses the longer side to specify the resolution.

lastzero:
Also be aware that we are still using FFmpeg 5 (included with Ubuntu 23.04). FFmpeg 6 sypports slightly different parameters, from what I know.

lukasdotcom:
I just noticed when I looked at what google photos does in the web video player, that it uses the height to specify the resolution 720p, 1080p, etc. I can still make the PR for it to use the larger side for the resolution or I can just follow what google photos does.

lastzero:
Google might behave differently on a phone though (where they use their native app, which we don't have). So we have to expect the videos to be played directly on a phone, in which case you might otherwise have bad quality. At least that is my understanding.

lastzero:
There's a transcoding test failing in the photoprism package, so that should be fixed as well (make sure to use our official development image for testing or, at a minimum, the same FFmpeg version):

failed

lukasdotcom:
Sorry about not seeing that earlier. I added it really quickly to the PR that makes the resolution use the larger side of the video. #3510

PR #3510, opened 2023-06-28
Config: Downscale videos based on largest dimension

lukasdotcom:
Has resolution mean the larger side and Fixes the failing tests mentioned in #3498.


internal/ffmpeg/convert.go
```diff
@@ -14,7 +14,7 @@ func AvcConvertCommand(fileName, avcName string, opt Options) (result *exec.Cmd,
 	} else if avcName == "" {
 		return nil, false, fmt.Errorf("empty output filename")
 	}
-
+	scale := "scale='if(gte(iw,ih), min(" + opt.Resolution + ", iw), -2):if(gte(iw,ih), -2, min(" + opt.Resolution + ", ih))'"
```
lastzero, line 17: Oh wow, I thought the scale parameter required would look simpler as it's a pretty common use case.
Will this also avoid upscaling? See http://trac.ffmpeg.org/wiki/Scaling#AvoidingUpscaling.

lukasdotcom: I think this would already avoid upscaling due to it using the minimum of either the maximum resolution or the original resolution. That is what the min is for.
```
min(opt.Resolution, iw)
```

lastzero: Thanks! I'll merge this then and see what happens :)

Issue #628, opened 2020-11-21
Details view: always show thumbnail in top menu bar


alexislefebvre:
When editing a picture, we don't see the current image:

> ![image](https://user-images.githubusercontent.com/2071331/99863163-30971680-2b9d-11eb-8d2a-e09162e0340a.png)

It would be nice to show a small thumbnail of the image (in top bar ?) so that I can keep track of the current image.

lastzero:
There is very little space for it in this view, especially on mobile. For now, you have to go back to details quickly if you don't remember the picture or used the prev / next buttons.

alanbernstein:
An alternative approach might be a popout side panel for details, rather than a separate view.

> There is very little space for it in this view, especially on mobile.

Fair enough; I had to laugh at this because am trying out photoprism on a desktop, and this view is 90% whitespace.
![Screenshot from 2022-06-08 01-18-36](https://user-images.githubusercontent.com/16123214/172782520-1798ccc5-3995-4b53-bddd-16b0610288af.png)

Sohalt:
Maybe at least the desktop view could always show the picture (using CSS media queries to hide it, if there's not enough space)

PR #3532, opened 2023-07-09
UX: Add preview image to the labels tab in the photo edit dialog

reubendowle:
This pull requests adds an image preview when editing labels.

image

This is to address this issue: https://github.com/photoprism/photoprism/issues/628

The issue points out that there is limited space for thumbnails on mobile devices, so it will be hidden on mobile devices.

Acceptance Criteria:

### DIFF
--- a/frontend/src/dialog/photo/edit/labels.vue
+++ b/frontend/src/dialog/photo/edit/labels.vue
@@ -1,93 +1,115 @@
 <template>
   <div class="p-tab p-tab-photo-labels">
-    <v-data-table
-        v-model="selected"
-        :headers="listColumns"
-        :items="model.Labels"
-        hide-actions
-        class="elevation-0 p-files p-files-list p-results"
-        disable-initial-sort
-        item-key="ID"
-        :no-data-text="$gettext('No labels found')"
-    >
-      <template #items="props" class="p-file">
-        <td>
-          <v-edit-dialog
-              :return-value.sync="props.item.Label.Name"
-              lazy
-              class="p-inline-edit"
-              @save="renameLabel(props.item.Label)"
+    <v-layout row wrap align-top fill-height>
+      <v-flex
+        v-if="!$isMobile"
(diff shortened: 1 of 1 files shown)


lastzero:
Thanks! We'll provide more feedback when we are back in our office.

lastzero:
@reubendowle I'll merge this, but need to refactor it as `$isMobile` only indicates if it's a mobile browser e.g. with touch event support, so doesn't depend on the size of the browser window, which is the actual concern:
https://github.com/photoprism/photoprism/blob/84a4c3c57e8e430dbedf803e0f18071549fc9b1b/frontend/src/app.js#L64-L66

graciousgrey:
Ideally we can also add a thumbnail to the ⚙️ tab.

lastzero:
@graciousgrey I'll check if that's possible. It might be good to add a component for this so we don't have to copy and paste the code into each tab....

lastzero:
Updated our release notes and the preview build, so you can test this:

reubendowle:
> @reubendowle I'll merge this, but need to refactor it as $isMobile only indicates if it's a mobile browser e.g. with touch event support, so doesn't depend on the size of the browser window, which is the actual concern:

@lastzero personally this looks fine for me at most sizes of the browser window on PC. I guess when the width gets too narrow triggers a scrollbar:

image

image

But this is the same behavior as the details tab. So maybe it can just be left like that, and maybe remove the isMobile condition completely?

lastzero:
I've already merged and refactored it yesterday. It's now deployed on our demo and available with our preview build.

loulou91:
Useful feature, but (for me) the most useful would be to have it on "people" tab...

lastzero:
@loulou91 Instead of rendering the image on each tab and thus reducing the space available for the actual content, I was thinking about a statically positioned preview e.g. in the lower left corner. However, we just merged a lot of PR and are still busy making them releasable, so this will have to wait a bit.

Issue #271, opened 2020-03-29
UX: Add batch edit dialog to edit multiple pictures at once


jannewulf:
First of all: I just discovered your project and really like it so far. I think it has great potential.

I wanted to edit the location of a group of photos, I think that is a common thing to do for DSLR or older photos. Right now I would have to click through all the photos that I selected and set the location for each. 

I think the more intuitive thing to happen if you select multiple photos and click the edit button is a batch editing mode. So you would set a information for all the selected images and the rest of the information would stay as it is.

This kind of relates to the discussion in #212, but is not restricted to label editing. So I opened a new Issue.

------------------------------------------------------------------------------------------------------------------------------------------------------------

**Acceptance Criteria:**
- [x] There MUST be a separate batch edit dialog
- [x] User MUST be able to change **at least** location, date and labels for a batch of photos
- [x] For fields that can have multiple values (e.g. keywords, labels) it MUST be possible to add a new value without overwriting all existing values
- [x] For fields that can have multiple values (e.g. keywords, labels) it MUST be possible to remove single values
- [x] Write unit tests
- [ ] Write acceptance test
- [ ] Update documentation

lastzero:
You're right, we'll implement batch editing in a future release. Also changing locations should be more convenient. At the moment, you can only change lat and long manually.

jannewulf:
I have not much experience with web development, but I would give it a try to implement this functionality. But there are some things that need to be discussed first.

I think we should discuss which fields should go into the edit dialog, in order to keep it clean. 
1. The most important are probably Date/Time and Location. The labels from object detection are also kinda important as I get lots of wrong classifications until now. 
   - Should go into batch edit dialog
2. There are also lots of information that are rarely modified, like description.
   - Not sure about this. I never used a description field, so I don't know if batch editing makes sense for this.
   - I saw that, in Google Photos, it is possible to edit the description in the extended info view (screenshot attached). 
3. And last but not least, there are fields that are not going to be edited in a software like this (Google Photos etc. replacement). These include Camera, Lens, Focal Length, ISO, Exposure, File names, ...
   - Should be moved to a read only extended info view.

I don't really get the Keywords field. Where do these come from? Not or not only from the object detection, I guess, as they are not the same as the labels. I was not able to find more info on that.

Screenshot of the Google Photos Extended Info view with the option to edit the description ("Beschreibung hinzufügen") there.
[...]

PR #3537, opened 2023-07-10
Batch edits

reubendowle:
This is an attempt to add the ability to modify all selected photos in one go. It allows:
* Adding a new label to all selected photos
* Setting the latitude & longitude on all selected photos

This is my attempt to address https://github.com/photoprism/photoprism/issues/271

This significantly improves my ability to import my large (50k photos) library into photoprism.

This is my first attempt at any project with either go or javascript, so there may be issues!

image

image

image

Acceptance Criteria:

### DIFF
--- a/frontend/src/component/photo/clipboard.vue
+++ b/frontend/src/component/photo/clipboard.vue
@@ -134,6 +134,17 @@
           <v-icon>delete</v-icon>
         </v-btn>
         <v-btn
+            v-if="canEdit" fab dark
+            small
+            :title="$gettext('Add Label')"
+            color="edit"
+            :disabled="selection.length === 0 || busy"
+            class="action-add-label"
+            @click.stop="dialog.addlabel = true"
+        >
+          <v-icon>new_label</v-icon>
+        </v-btn>
+        <v-btn
             fab dark small
             color="accent"
             class="action-clear"
@@ -151,6 +162,8 @@
                           @confirm="addToAlbum"></p-photo-album-dialog>
     <p-share-upload-dialog :show="dialog.share" :items="{photos: selection}" :model="album" @cancel="dialog.share = false"
                            @confirm="onShared"></p-share-upload-dialog>
+    <p-label-add-dialog :show="dialog.addlabel" @cancel="dialog.addlabel = false"
+                          @confirm="addLabel"></p-label-add-dialog>
   </div>
 </template>
 <script>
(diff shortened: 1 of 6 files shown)


dror3go:
Looks nice! I'm glad this issue gets attention.

I haven't looked at the implementation itself (no time, sorry) - how are you sending the actual edits to the server? Note that a single HTTP request per file chosen might be a problem.

I personally would expect a batch edit to use the existing pencil button to open the same editing form, with some fields being disabled or not visible, and not to add a new button to that context menu for evefy single editing option. Ofcourse, it needs to somehow clarify that this will affect X amount of photos.
What do you think?

reubendowle:
> Looks nice! I'm glad this issue gets attention.

When I imported my huge photo library, the first question from my wife was "so how do I add a label to lots of photos". So I thought I would have a go at coding up a solution for it.

> I haven't looked at the implementation itself (no time, sorry) - how are you sending the actual edits to the server? Note that a single HTTP request per file chosen might be a problem.

I have added new API endpoints to do the batch editing which take the list of photos to change, along with the value to set. The current implementation is perhaps not as flexible as some people have been requesting, since it currently only allows adding new labels and setting the gps location.

> I personally would expect a batch edit to use the existing pencil button to open the same editing form, with some fields being disabled or not visible, and not to add a new button to that context menu for evefy single editing option. Ofcourse, it needs to somehow clarify that this will affect X amount of photos. What do you think?

I think using the existing window is not great, as it would just create confusion on what photos are being edited. Also, there are lots of fields which are not common (eg ISO), and how would you indicate to user that only some fields and not others would be applied to all photos.

Maybe a compromise might be a single new batch edit window, which had the various fields available for changing. They would start out blank, and blank fields would not be modified?

lastzero:
@reubendowle **Thank you very much for your contribution!** I have just merged your other PR. Unfortunately, this one can't be merged easily, mainly for two reasons:

(1) All functionality needs to be fully responsive, meaning it needs to work on both small and large screens. However, the context/clipboard menu becomes too large this way, especially since there are many more fields that users will want to edit (so we need to find another approach)

(2) Similarly, adding a new dialog for each field does not scale well, meaning that the maintenance and testing overhead would increase significantly with each new batch edit dialog.

Because of this, I see no other option but to come up with a different UX approach. We may even need to test different solutions to gain enough confidence. Ideally, we can offer a single responsive dialog that works on mobile phones and desktop computers.

lastzero:
Since we unfortunately can't merge it at the moment for the reasons mentioned above, I'm closing this PR. Once we have a better idea of how Batch Edit can be implemented (given the constraints), we may come back to your changes to see if they can be integrated.

PR #3555, opened 2023-07-20
Develop: Add Delve to install-go-tools.sh script

gokcedilek:
- Add the Delve debugger as a dependency


lastzero:
Additional Go tools for development can be added to this script:
- https://github.com/photoprism/photoprism/blob/develop/scripts/dist/install-go-tools.sh

It is included in the Dockerfile from which we build our [develop base image](https://hub.docker.com/r/photoprism/develop/tags):

https://github.com/photoprism/photoprism/blob/73fa7bbe86b9d24f6a408ddeac7ca169033c105d/docker/develop/lunar/Dockerfile#L73-L77

We can then trigger a new build for you, so all you have to do is pull the latest image and use it for your development environment. To do this, simply run the following command in the main project directory (from the host):

```
make docker-build
```

Issue #3080, opened 2023-01-09
Albums: Update UpdatedAt when photos are added/removed to/from an album


graciousgrey:
**As a user I want to see albums with newly added photos on top when filtering for "Recently Updated"**

At the moment, the UpdatedAt field of albums is only updated when the user makes changes in the album's edit dialog (https://docs.photoprism.app/user-guide/organize/albums/#edit-album-details). It would be great if the UpdatedAt field would be additionally updated when photos are added to or removed from an album.

- [x] The UpdatedAt field of an album (type album) MUST be updated when new photos are added
- [x] The UpdatedAt field of an album (type album) MUST be updated when photos are removed from the album
- [x] Write Unit Tests
- [ ] Update Documentation

PR #3566, opened 2023-07-21
Albums: Refresh updated timestamp when photos are added

gokcedilek:
This PR is to address issue #3080

- The `updatedAt` field is updated when we save an entry into the database. This means, when we add a new photo to an album, in addition to saving the `PhotoAlbum` object, we also need to save the associated `Album` object with the updated value for `updatedAt`.


lastzero:
My changes add a simple `UpdateAlbum()` function that can be used to update the timestamp (or other values) with as little overhead as possible, e.g. it's not needed to fetch the album record before updating the column and the column should only be updated after all entries have been added if it's a batch operation to avoid unnecessary update queries.

It is still needed to add unit tests and to check if there are other situations where the content of an album is changed by the user and therefore updating the timestamp would be appropriate. For automatically created Moments, Folders etc. it should be checked if these are updated only if there are changes or routinely so the value in updated_at is meaningless.

Also note:
- `AddPhotoToUserAlbums()` is only used for adding files to albums when they are newly imported. When the user adds photos to a album through the UI otherwise, `Album.AddPhotos()` is used instead.
- The `Updates()` function provided by the ORM should automatically refresh the `update_at` column when other values are updated as it calls the hooks, see https://v1.gorm.io/docs/update.html#Update-Columns-w-o-Hooks
- In contrast, `UpdateColumns()` will only update the specified value(s) and does not call any hooks.

internal/entity/album.go
```diff
@@ -639,6 +650,15 @@ func (m *Album) UpdateFolder(albumPath, albumFilter string) error {
 	} else if err = UnscopedDb().Exec("UPDATE albums SET album_path = NULL WHERE album_type = ? AND album_path = ? AND id <> ?", AlbumFolder, albumPath, m.ID).Error; err != nil {
 		return err
 	}
+	
+	// update folder
+	if folder := FindFolder("/", albumPath); folder != nil {
```
lastzero: The `Updates()` function used above should automatically refresh the `update_at` column, so updating it again should not be necessary.

internal/entity/album.go
```diff
@@ -639,6 +650,15 @@ func (m *Album) UpdateFolder(albumPath, albumFilter string) error {
...
+		folder.Updates(
+			map[string]interface{}{
```
lastzero: In the entity package, you can use `Values{}` as a shortcut for `map[string]interface{}`.

internal/entity/album.go
```diff
@@ -125,6 +125,17 @@ func AddPhotoToUserAlbums(photoUid string, albums []string, userUid string) (err
 			if err = entry.Save(); err != nil {
 				log.Errorf("album: %s (add photo %s to albums)", err.Error(), photoUid)
 			}
+
+			// save the album with updated value for updated_at
+			m := Album{}
```
lastzero: See the code I've just pushed for how to update just the `updated_at` column without loading the record first and then indirectly updating it through the hooks that `Save()` calls.

frontend/src/page/albums.vue
```diff
@@ -346,7 +348,7 @@ export default {
           {value: 'newest', text: this.$gettext('Newest First')},
           {value: 'oldest', text: this.$gettext('Oldest First')},
           {value: 'added', text: this.$gettext('Recently Added')},
-          {value: 'edited', text: this.$gettext('Recently Edited')},
+          ...(typeName === 'album' ? [{value: 'edited', text: this.$gettext('Recently Edited')}] : [])
```
lastzero: We use the computed `context` property elsewhere for this. Would it be possible to use it here as well?
```js
  computed: {
    context: function () {
      if (!this.staticFilter) {
        return "album";
      }

      if (this.staticFilter.type) {
        return this.staticFilter.type;
      }

      return "";
    }
  },
```

gokcedilek: yes, updated! I was also thinking instead of `:items="context === 'album' ? options.sorting : options.sorting.filter(item => item.value !== 'edited')"` I could do something like `:items="context === 'album' ? options.albumSorting : options.sorting"` and define a second property `albumSorting` which contains all the album-specific sorting properties, since we might want to add/remove more in the future specifically for albums (for example I think @graciousgrey mentioned we might hide one of the sort options "Newest first" or "Recently Added" for manually managed albums as they are the same). but if we only hide recently edited for now in this PR I think this implementation is also just fine, let me know if you have a preference 👍

internal/entity/album_test.go
```diff
@@ -503,6 +570,44 @@ func TestAlbum_RemovePhotos(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		first_photo_updatedAt := strings.Split(entries[0].UpdatedAt.String(), ".")[0]
+		second_photo_updatedAt := strings.Split(entries[1].UpdatedAt.String(), ".")[0]
+		album_updatedAt := strings.Split(a.UpdatedAt.String(), ".")[0]
+
+		assert.Truef(
+			t, first_photo_updatedAt <= album_updatedAt,
+			"Expected the UpdatedAt field of an album to be updated when"+
```
graciousgrey, line 602: I found only one tiny thing that could be improved: Shouldn't it say "photos are removed" instead of "...new photos are added" in this test?

gokcedilek: thank you for the catch, fixed! 😄

graciousgrey:
From my side we can merge this :)

lastzero:
Looks like this can be merged now. Thank you! :hugs:

lastzero:
@gokcedilek Next time, run `make fmt` in the main project directory to automatically reformat the code if needed :+1:

PR #3568, opened 2023-07-23
UX: Allow pasting lat / lng in a single operation

reubendowle:
This speeds up the workflow when pasting location found in external mapping application

Acceptance Criteria:

### DIFF
--- a/frontend/src/dialog/photo/edit/details.vue
+++ b/frontend/src/dialog/photo/edit/details.vue
@@ -158,6 +158,7 @@
                 placeholder=""
                 color="secondary-dark"
                 class="input-latitude"
+                @paste="pastePosition"
               ></v-text-field>
             </v-flex>
 
@@ -174,6 +175,7 @@
                 placeholder=""
                 color="secondary-dark"
                 class="input-longitude"
+                @paste="pastePosition"
               ></v-text-field>
             </v-flex>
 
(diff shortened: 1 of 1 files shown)


frontend/src/dialog/photo/edit/details.vue
```diff
@@ -469,6 +471,27 @@ export default {
 
       this.time = taken.toFormat("HH:mm:ss");
     },
+    pastePosition(event) {
+      /* If text is pasted in the form of "XXX.XXXX, YYY.YYYY" (for example if
+       * pasting a position from external mapping app) then auto-fill both lat
+       * and lng fields
+       */
+      let paste = (event.clipboardData || window.clipboardData).getData("text");
+      let s = paste.split(/[ ,]+/);
```
lastzero, line 480: Looks like it might throw an exception if paste isn't a string for some reason?

lastzero:
Thank you very much! I didn't know a `@paste` event exists and can be used with VueJS. Will this work with all browsers? Does it require further testing e.g. with browsers you haven't tested yet?

lastzero:
As I'm not sure how much time I have for this next week, I'll move ahead to merge your PR and add checks in case the browser clipboard cannot be accessed for some reason. Thank you for your contribution! :heart:

reubendowle:
> Thank you very much! I didn't know a @paste event exists and can be used with VueJS. Will this work with all browsers? Does it require further testing e.g. with browsers you haven't tested yet?

It seems you can capture any DOM event in Vue. This should work with basically any browser. They all support the paste event (https://developer.mozilla.org/en-US/docs/Web/API/Element/paste_event#browser_compatibility)

frontend/src/dialog/photo/edit/details.vue
reubendowle: The specification of this seems to indicate that you will always get an empty string result even if the data type does not support strings (https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/getData#return_value).

I tested this by pasting some non-text data (eg image data), and the code just ends up with s.length == 0, and thus ignoring it.

PR #3571, opened 2023-07-23
UX: Refactor position pasting to allow for any DMS format

reubendowle:
A refactor of previous pull request https://github.com/photoprism/photoprism/pull/3568 to use a library to allow position to be pasted in any of the common DMS position formats.

Acceptance Criteria:

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -47,6 +47,7 @@
         "eventsource-polyfill": "^0.9.6",
         "file-loader": "^6.2.0",
         "file-saver": "^2.0.5",
+        "geodesy": "^2.4.0",
         "hls.js": "^1.4.10",
         "i": "^0.3.7",
         "karma": "^6.4.2",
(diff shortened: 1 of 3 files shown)


lastzero:
Thank you for reviewing my changes! I've just improved the parser to split the values using the regex you suggested in your earlier PR. Based on my tests, it now works with all floating point formats, including `52.4402889,13.5956639,15.52z` when you copy the coordinates from the Google Maps URL.

Would you be ok with not including the `geodesy` lib to limit the code size, at least until someone really needs it?

reubendowle:
> Would you be ok with not including the geodesy lib to limit the code size, at least until someone really needs it?

I don't personally need the other formats, I just did that for completeness. So its fine by me to leave it out.

The Dms.parse() function is actually a relatively short function. If we wanted this functionality we could perhaps just copy that function without the rest of the library.

lastzero:
Awesome! Thanks for your quick feedback. I'll close this for now then. If anyone needs the library, we'll be happy to add it, though most people will probably wait for us to have a dialog where they can simply select the location on a map.

PR #3577, opened 2023-07-26
UX: Don't process all features for clusters exceeding 100 photos.

reubendowle:
Fixes performance issues with large numbers of photos. For me I was seeing render take over 6s sometimes. Profiling shows it was spending a lot of time in getClusterLeaves(). This was iterating over thousands of photos for some clusters, which is very slow.

image

We only need to know about up to 100 photos per cluster, so now the result size is limited to that. This completely resolves the performance issues for me

This could possibly be further optimized, since we only need to store data for the first 4 to be able to render the thumbnail.

Acceptance Criteria:

### DIFF
--- a/frontend/src/page/places.vue
+++ b/frontend/src/page/places.vue
@@ -634,8 +634,7 @@ export default {
         }
       });
 
-      this.map.on('render', this.updateMarkers);
-
+      this.map.on('idle', this.updateMarkers);
 
       this.search();
     },
--- a/frontend/src/page/places.vue
+++ b/frontend/src/page/places.vue
@@ -305,7 +305,7 @@ export default {
       });
     },
     selectClusterById: function(clusterId) {
-      this.getClusterFeatures(clusterId, (clusterFeatures) => {
+      this.getClusterFeatures(clusterId, -1, (clusterFeatures) => {
         let latMin,latMax,lngMin,lngMax;
         for (const feature of clusterFeatures) {
           const [lng,lat] = feature.geometry.coordinates;
@@ -490,16 +490,16 @@ export default {
 
       this.map.on("load", () => this.onMapLoad());
     },
-    getClusterFeatures(clusterId, callback) {
-      this.map.getSource('photos').getClusterLeaves(clusterId, -1, undefined, (error, clusterFeatures) => {
+    getClusterFeatures(clusterId, limit, callback) {
+      this.map.getSource('photos').getClusterLeaves(clusterId, limit, undefined, (error, clusterFeatures) => {
         callback(clusterFeatures);
-      });;
+      });
     },
     getMultipleClusterFeatures(clusterIds, callback) {
       const result = {};
       let handledClusterLeaveResultCount = 0;
       for (const clusterId of clusterIds) {
-        this.getClusterFeatures(clusterId, (clusterFeatures) => {
+        this.getClusterFeatures(clusterId, 100, (clusterFeatures) => {
           result[clusterId] = clusterFeatures;
           handledClusterLeaveResultCount += 1;
 
--- a/frontend/src/page/places.vue
+++ b/frontend/src/page/places.vue
@@ -548,9 +548,10 @@ export default {
 
               const clusterFeatures = clusterFeaturesById[props.cluster_id];
               const previewImageCount = clusterFeatures.length > 3 ? 4 : 2;
-              const images = clusterFeatures
-                .slice(0, previewImageCount)
-                .map((feature) => {
+              const images = Array(previewImageCount)
+                .fill()
+                .map((a,i) => {
+                  const feature = clusterFeatures[Math.floor(clusterFeatures.length * i / previewImageCount)];
                   const imageHash = feature.properties.Hash;
                   const image = document.createElement('div');
                   image.style.backgroundImage = `url(${this.$config.contentUri}/t/${imageHash}/${token}/tile_${50})`;


graciousgrey:
Thank you :)

heikomat:
Very cool

heikomat (changes requested):
`getClusterFeatures` is used to determine min/max lat/lng of a cluster.
If not all features of the cluster are considered, then the min/max lat/lng used to determine what photos to show when opening the cluster might be wrong.

But: You could consider using only like 4 features for the first render and only request all features of a cluster when it is actually clicked.

Update: see here: https://github.com/photoprism/photoprism/blob/36bac7ab481d47a0d10a922596534ddbe1b742c0/frontend/src/page/places.vue#L307-L328

But this function is only used on click of a cluster:
https://github.com/photoprism/photoprism/blob/36bac7ab481d47a0d10a922596534ddbe1b742c0/frontend/src/page/places.vue#L567-L569

So instead of having a fixed `-1` or `100`, `getClusterFeatures` could get a new parameter so that it can fetch all on a cluster-click, and only a couple on regular render

reubendowle:
> `getClusterFeatures` is used to determine min/max lat/lng of a cluster. If not all features of the cluster are considered, then the min/max lat/lng used to determine what photos to show when opening the cluster might be wrong.

Ah, I did not notice. I will fix this up.

> But: You could consider using only like 4 features for the first render and only request all features of a cluster when it is actually clicked.

But this is probably the best compromise. But note that this will still make clicking on the cluster quite slow. It was taking as much as 10s to get on some of the large clusters I have. I have a ryzen 3700x which is reasonable perf, but on some mobile devices this would probably be unacceptably slow. 

I will have another look and see if there are some other ways of optimizing. Possibly when just collecting the min/max position we can do less work when iterating the features.

reubendowle:
@heikomat I have implemented this as suggested and now use a function parameter to limit the number of features only during render.

I have also made another change to only update the markers on the final render. Even with this fix, it is still a relatively expensive operation. Usually takes ~12ms on my PC, but when clicking on the marker the marker update is also triggered and seemed to take 70ms there. I think on slow mobile devices only doing this once at the end this makes the scrolling and zooming a bit less processor intensive.

Lastly I made a small adjustment to the photos included in the marker. Previously it was the first four, but now I have spread them out amongst the photos to give a more varied picture of what is under the marker.

heikomat (approved):
Looks very good to me! 👍 

The only thing i'm not exactly sure about is when 'render' and 'idle' are triggered.
could it be that one of them is triggered more often than the other?
- does 'idle' always trigger after render, or are there cases where render is triggered and idle isn't?
- do idle and/or render trigger after for example window resize?

If you can confidently say that 'idle' is a good event to use then these changes are fine imo

reubendowle:
> The only thing i'm not exactly sure about is when 'render' and 'idle' are triggered.
> could it be that one of them is triggered more often than the other?

That is the idea. The render event is fired 10s of times when you change zoom, and constantly as you pan (ie each time the map re-renders itself). But there is not really a good reason to constantly update the map markers - the existing ones are moved around and zoomed, so it still looks nice and smooth.


> *   does 'idle' always trigger after render, or are there cases where render is triggered and idle isn't?
> *   do idle and/or render trigger after for example window resize?

The maplibre GL JS docs have this to say about "idle" event:
> Fired after the last frame rendered before the map enters an "idle" state: - No camera transitions are in progress - All currently requested tiles have loaded - All fade/transition animations have completed

So I think this will always fire at the same time that the render event would do so, but it only fires exactly once.

> If you can confidently say that 'idle' is a good event to use then these changes are fine imo

I have limited experience with this map component, so I can't be certain of this. It seems to work fine in the testing I have done.

heikomat:
In that case i agree, idle seems to be way more suitable for our usecase here 👍 well done

graciousgrey:
@reubendowle @heikomat Thank you very much for working on / reviewing this :heart: 

From your point of view, is this ready to be merged after @lastzero takes one final look?

lastzero:
Thank you! I'll merge this now for testing 👍

lastzero:
A new development preview build should be available soon :)

lastzero:
It seems that the markers are no longer updating properly with these changes. The following screenshots were taken on our public demo and should show exactly the same (number of) pictures. Note that in Europe only one picture is shown until you zoom in (and even then the position is often wrong, e.g. a cluster taken in Berlin is shown in Erfurt until you zoom in and out several times):


Screenshot 2023-08-15 at 07 26 05

From what I know, the picture shown in Europe in the screenshot above was actually taken in Canada. When you zoom in, it's gone and the actual pictures taken in Europe are displayed:

Screenshot 2023-08-15 at 07 26 46

lastzero:
@reubendowle @heikomat I've also added a comment with my findings to the original cluster view issue #1187 so others can see that this is an open problem that needs to be solved. Any ideas on how to fix what I described above are welcome. I'll otherwise work on it later this week so we can get this released.

reubendowle:
@lastzero When I test it out myself on the public demo, what I see is that it always renders correctly eventually. I think the change from render to idle here might be causing the issue:

https://github.com/photoprism/photoprism/blob/16c78767b26c61caabb09a90c54dec0bcedd343c/frontend/src/page/places.vue#L638

What I find when I test it is that some of the map tiles can take several seconds to finish loading in, so the final render rendering before idle takes too long. Note that the original 'render' event is not ideal since as per the definition:

> Fired whenever the map is drawn to the screen, as the result of - a change to the map's position, zoom, pitch, or bearing - a change to the map's style - a change to a GeoJSON source - the loading of a vector tile, GeoJSON file, glyph, or sprite

It fires not just on changes to the map position/zoom etc, but also when each tile is loaded in. This means the clusters get recalculated for many unnecessary render events which would not result in changes the clusters.

Unfortunately there is no event which triggers for all the changes we want to catch (position, zoom, pitch, bearing changes), so we would probably have to implement half a dozen or more events to properly optimize this, or alternatively we could add code to the render event to check if these properties have changed since the last time the clusters have changed.

I will have a look at this when I have some free time in a couple of days, but I suspect the easiest thing would just be to revert the change on that line.

lastzero:
> It fires not just on changes to the map position/zoom etc, but also when each tile is loaded in. This means the clusters get recalculated for many unnecessary render events which would not result in changes the clusters.

That makes sense. On the other hand, it seemed to work quite well before? Was that because the callback was less complex and therefore simply executed faster?

I think I built this based on an example provided by the library authors, so I assumed that was correct. Perhaps you can check to see if there are updated examples or if the docs give any hints?

Reverting your change would be possible, but then we would still have to deal with the performance issues we had before - so no real improvement.

Thank you very much for looking into this!

reubendowle:
> Reverting your change would be possible, but then we would still have to deal with the performance issues we had before - so no real improvement.

Actually almost all of the improvement was from only using a maximum of 100 photos to create the cluster. Previously it was unbounded, which got VERY slow when there were 10s of thousands.

This render vs idle change is actually really minor and is probably just me over optimising where it is not needed.

lastzero:
@reubendowle @heikomat I've replaced `100` with `-1` locally in this line, hoping this is all that's needed:

https://github.com/photoprism/photoprism/blob/be0fdc1774266bd4ec09e01ab93496fb07a4cbed/frontend/src/page/places.vue#L502

It unfortunately did not fix the issue for me...

Also, when I searched for `scan:true`, the map zoomed to the correct position, but no thumbnails were displayed at all. They were normal photos, just flagged as scans. So I don't think that only this particular filter is causing problems.

reubendowle:
> @reubendowle @heikomat I've replaced 100 with -1 locally in this line, hoping this is all that's needed:
> It unfortunately did not fix the issue for me...

I don't think that line is not the source of the issue - it is caused by the updateMarkers function not being called at the correct times. I think this pull request should fix it:

https://github.com/photoprism/photoprism/pull/3636

With this, the update of the markers is actually done according to the examples for maplibre-gl: https://maplibre.org/maplibre-gl-js/docs/examples/cluster-html/

lastzero:
Thanks, I'll give it a try! If it works locally, we will provide a new preview build and update our public demo with it.

cincodenada:
Just wanted to drop a note here to say thank you!! I finally got around to upgrading my instance of Photoprism and this, along with #2596, has made the map usable again!

When I zoom in on my house, I have literally 11,000 photos in one cluster at the highest zoom level, and many many other clusters with hundreds or thousands of photos around my neighborhood and other frequently-visited places. That volume was absolutely destroying performance of the map with the old code. Smooth as butter now, thank you!

Issue #439, opened 2020-08-03
Live Photos: Add support for Samsung Motion Photos


ark-:
Initial support for Samsung Motion Photos is now available for testing on our public demo at https://demo.photoprism.app/ and/or with the development preview build:

- https://docs.photoprism.app/release-notes/#development-preview
- https://docs.photoprism.app/getting-started/updates/#development-preview

Please do not upload any private pictures to [our demo](https://demo.photoprism.app/), as everyone can see them there. Thank you very much! 🙏

---

Support for Samsung Motion Photos could be implemented similar to Apple Live Photos:
- https://github.com/photoprism/photoprism/issues/177

The videos are stored in the exif data under the following tags:

```
EmbeddedVideoType: MotionPhoto_Data
EmbeddedVideoFile: <Binary File>
```

Where `EmbeddedVideoFile` is an MP4 file with its own exif data and is of type `video/mp4`. Version MP4 v2 [ISO 14496-14], compressor avc1.

I'm sure `EmbeddedVideoType` could be something other than `MotionPhoto_Data` for other Android phone manufacturers as I imagine "MotionPhoto" is a trademark name. Unfortunately I do not know the details of the other devices but some research could make this improvement cover other manufacturer's phones too.

I found discussion and exiftool commands to extract this information here: https://forum.xda-developers.com/android/software/samsung-motion-photo-extractor-t3339997

Example commands:

```
exiftool -EmbeddedVideoFile -b -w _MotionPhoto.mp4 *.jpg

exiftool -trailer:all= -if "$EmbeddedVideoType eq 'MotionPhoto_Data'" *.jpg
```

lastzero:
Samsung embeds video files in JPEGs, is that correct?

Extracting makes sense if this is a common file type and Samsung still uses it... seems to be a hack on the first sight, so I wouldn't be surprised if they start(ed) using a different solution like Apple e.g. storing multiple files.

As a workaround, you can probably write a script to "manually" extract videos before indexing (or maybe you already did so)?

lastzero:
Anyhow, it would be good to get example files for testing! If we don't have time now, then maybe later.

ark-:
Hmmm, the script would be useful for a one time import but if I'm auto uploading from my phone that's not ideal. BUT it's a workable solution. I may do this to see how it looks in the UI.

Let me ask around some people I know with newer Samsung phones and get sample photos from them. Also if others realise they could use this feature maybe some thumbs up on this suggestion would help show a need.

[...]

Issue #3559, opened 2023-07-20
Live Photos: Sidecar video file dimensions are indexed incorrectly


graciousgrey:
#### 1. What is not working as documented?

On the files tab of the edit dialog a wrong video size is displayed.

#### 2. How can we reproduce it?

1. Index a live photo, e.g. IMG_123.heic ("ImageSize": "4032x3024") + IMG_123.mov ("ImageSize": "1920x1440"). 
2. Go to the files tab of the edit dialog

For the .heic file, the correct size is displayed. For the .mov file, the size of the .heic file is displayed.

#### 3. What behavior do you expect?

Display the correct size for all files.

lastzero:
With these changes, the correct file properties should be displayed after reindexing all affected files.

Issue #3429, opened 2023-05-24
Files: Unstack button is missing for stacked media files


schellbell:
I have "stack sequences" enabled in my library. Generally speaking, PhotoPrism handles stacks properly, and where it stacks photos it shouldn't I can unstack them easily enough. However, I am occasionally photos that, for whatever reason, cannot be unstacked. 

For example, I have a live photo shot on iPhone and the associated jpg files for the thumbnail. PhotoPrism originally stacked 4 files together: "IMG_1376_HEVC.MOV", "IMG_1376_HEVC.MOV.jpg", "IMG_1377_HEVC.MOV", and "IMG_1377_HEVC.MOV.jpg"

Obviously, the naming of these files is sequential, and the photos were taken in a relatively short period of time. I wanted to separate the two sets of files (the .MOV paired with its .jpg). I am able to unstack the .MOV file, but there is no button to unstack the jpg file resulting in the screenshot below

![image](https://github.com/photoprism/photoprism/assets/26511627/63568c6f-7f5f-4d29-919c-e425c58d7a0c)

If I wanted to, I could unstack IMG_1377_HEVC.MOV from IMG_1377_HEVC.MOV.jpg, but there is no option to unstack 1376. 

![image](https://github.com/photoprism/photoprism/assets/26511627/9fc5996a-0904-4e0b-b006-d840767f8efa)

This is just one example, but I have encountered others, always when it is a *.jpg that is associated with another file (movie, raw format photo, etc).

I'm running PhotoPrism Build [230514-0b780defb] and MariaDB 6.10.

graciousgrey:
Thank you very much for reporting!

Could you send us some sample files to samples@photoprism.app ?

lastzero:
We will be happy to provide you with a fix as soon as we can reproduce the issue.

schellbell:
Apologies for the delay - I've sent the sample files now. Thanks for your help!

graciousgrey:
I will go ahead and close this for now as it was not reproducible with the sample files and no on else experienced this in the meantime.

Issue #2788, opened 2022-10-18
Metadata: Flag Samsung/Google Motion Photos as Live Photos


MEDumont:
#### 1. What is not working as documented?

When importing Google Motion Photos of the format PXL_{Timestamp}.MP and PXL_{Timestamp}.MP.jpg, if the video is less than a second long, the stack is not marked as a Live photo, even though it is a live photo. This results in many manual edits to fix them.

#### 2. How can we reproduce it?

1. Setup Photoprism with an originals folder containing many Google Motion Photos
2. Navigate to the Library Tab
3. Check Complete Re-Scan
4. Click Start and wait to finish
5. Navigate to Search tab and find any Google Motion Photo with a video length of less than 1 second.
6. It will be categorized as a video, not a live photo.

I have the default settings for thumbnail quality. I am using an NFS share for my originals folder. Photoprism is running a docker container that has that NFS share mounted directly to it.

#### 3. What behavior do you expect?

I expect my Google Motion Photos to be categorized as Live if they are under 3 seconds in length.

#### 4. What could be the cause of your problem?

I believe the cause of the problem could be a hard-coded rule regarding the length of video files under 1 second in the indexer.

#### 5. Can you provide us with example files for testing, error logs, or screenshots?
![issue](https://user-images.githubusercontent.com/7817161/196487168-fb789ee9-1111-49c4-a80d-cfe5e8137866.png)

#### 6. Which software versions do you use?

(a) PhotoPrism Architecture & Build Number: Build [220901-f493607b0]

(b) Database Type & Version: MariaDB 10.8

(c) Operating System Types & Versions: Ubuntu 22.04

(d) Browser Types & Versions: Vivaldi 5.5

(e) Ad Blockers, Browser Plugins, and/or Firewall Software? PiHole DNS, Adblock Plus

#### 7. On what kind of device is PhotoPrism installed?

Docker container running in an Ubuntu 22.04 Virtual Machine

#### 8. Do you use a Reverse Proxy, Firewall, VPN, or CDN?

I use Nginx as a reverse proxy, but I am not having issues with that.

kvalev:
Google motion photos are not supported, see #1739

MEDumont:
I'm aware they are not supported natively, but Google Takeout splits the video and photo into two separate files and they are imported as live because they stack and have the same filename (which is awesome!). The only issue is: aren't videos that are stacked with photos under 3 seconds supposed to be categorized as live, not just standalone video?

[...]

PR #3588, opened 2023-07-29
Live Photos: Add Support for Samsung Motion Photos

gokcedilek:
Issue #439 

- Convert Samsung motion photos into video files
  - Extract the video using `exiftool`
  - Identify the mime type of the video file 
  - Save the extracted video into the `sidecar` folder with the same path, the same way preview images that are extracted from videos are saved
- Also a JSON sidecar of the video file is generated afterwards


internal/photoprism/convert_sidecar_json.go
```diff
@@ -58,7 +58,39 @@ func (c *Convert) ToJson(f *MediaFile, force bool) (jsonName string, err error)
...
-		return "", fmt.Errorf("exiftool: failed creating %s", filepath.Base(jsonName))
+		return "", fmt.Errorf(
+			"exiftool: failed creating %s", filepath.Base(jsonName),
+		)
+	}
+
+	// attempt to extract the video file if it exists
+	var extractCmd *exec.Cmd
+	path := filepath.Join(
+		Config().SidecarPath(), f.RootRelPath(), "%f"+fs.ExtMP4,
```
lastzero: Is there a way to tell if the embedded file is actually an MP4 video? This might change in the future when new formats are introduced.

internal/photoprism/convert_sidecar_json.go
```diff
@@ -58,7 +58,39 @@ func (c *Convert) ToJson(f *MediaFile, force bool) (jsonName string, err error)
...
+	)
+
+	if f.IsJpeg() {
```
lastzero: Is this performed for all JPEGs? If so, it could slow down indexing and should only be executed if the file is likely to contain an embedded video, e.g. based on the previously extracted metadata.

lastzero:
@gokcedilek Based on my current understanding, it may be best to extract and add embedded files in the `MediaFile.RelatedFiles()` method, right next to this block, which adds edited versions of a file based on Apple's proprietary naming scheme:

https://github.com/photoprism/photoprism/blob/71c30e4218a7f430ec2149f57ced86ae44207f82/internal/photoprism/mediafile_related.go#L54-L56

This way, the additional video file would be treated like any other related file, even if it does not exist yet.

Would that be feasible and does it make sense for you? :thinking:

internal/photoprism/mediafile.go
```diff
@@ -17,6 +22,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/h2non/filetype"
```
lastzero: Note that we've switched to the `github.com/gabriel-vasile/mimetype` library for mime type detection and that the `fs` package includes useful helper functions to work with file types and the filesystem:
https://github.com/photoprism/photoprism/blob/develop/pkg/fs/mime.go

internal/photoprism/mediafile.go
```diff
@@ -314,6 +320,113 @@ func (m *MediaFile) EditedName() string {
 	return ""
 }
 
+func (m *MediaFile) VideoFileName(jsonName string) (string, error) {
```
lastzero: Maybe you should rename `VideoFileName` to `EmbeddedVideoName` or `ExtractEmbeddedVideo` so that it is clear that it does not return the filename of a regular video sidecar file?

internal/photoprism/mediafile.go
```diff
@@ -314,6 +320,113 @@ func (m *MediaFile) EditedName() string {
...
+			"mediafile: %s not found", clean.Log(m.RootRelName()),
+		)
+	} else if m.Empty() {
+		return "", fmt.Errorf(
+			"mediafile: %s is empty", clean.Log(m.RootRelName()),
+		)
+	}
+
+	// given the filename, read the json sidecar
+	jsonBytes, err := os.ReadFile(jsonName)
```
lastzero: Maybe we can add this logic to the existing ExifTool JSON parser in the `meta` package?

https://github.com/photoprism/photoprism/blob/91cc358fc59741b6d6faad6992514ef5c99bc41e/internal/meta/json.go#L22-L38

See also https://github.com/photoprism/photoprism/blob/develop/internal/meta/json_exiftool.go.

The field name of the embedded video file (if any) could be written to a new field you add to https://github.com/photoprism/photoprism/blob/develop/internal/meta/data.go, e.g. `meta.Data.EmbeddedVideo`.

All you need to do then is:
```go
if metaData := m.MetaData(); metaData.Error == nil {
  // your code
}
```

go.mod
```diff
@@ -59,10 +59,7 @@ require github.com/olekukonko/tablewriter v0.0.5
 
 require github.com/google/uuid v1.3.0
 
-require (
-	github.com/chzyer/readline v1.5.1 // indirect
-	github.com/zitadel/oidc v1.13.4
```
lastzero: Note that we also maintain some private extension repos, including an OIDC implementation that is not yet ready for release. Unfortunately, there seems to be no other way than adding their deps to the main `go.mod` file. So if you update `go.mod` and `go.sum` without the additional code being in the project directory, those deps will be removed even though they may be needed. Ideally, we will find another solution to this in the future.

lastzero:
I've searched GitHub for issues related to (additional) JPEG sidecar files, especially in conjunction with Live Photos:

- https://github.com/photoprism/photoprism/issues/926
- https://github.com/photoprism/photoprism/issues/1028
- https://github.com/photoprism/photoprism/issues/1895
- https://github.com/photoprism/photoprism/issues/2788
- https://github.com/photoprism/photoprism/issues/3429
- https://github.com/photoprism/photoprism/issues/3559

Reading through these may give you a better idea of what the best strategy might be and how it affects indexing/viewing (for better or worse). Rather than opening a new issue, it might be better to add a comment to one of these issues if you see fit. There may also be other related issues that I just haven't found yet.

**To sum up my insights:**

- Not having the additional JPEG for live photos would definitely avoid situations where this low-res JPEG would be used as the primary JPEG instead of a high-res JPEG created from the HEIC (if it doesn't already exist).
- Also, the indexer doesn't currently seem to use the video JPEG for extracting metadata (but uses the primary instead), so it's *probably* of no use at the moment. If we need it later, it could be created then: https://github.com/photoprism/photoprism/blob/f15ca1d7e5811abb2ef285e60942fcd0c57b20be/internal/photoprism/index_mediafile.go#L651-L669
- However, *if there is no other preview image available for videos*, they will always need a JPEG still image that can be used as the primary image for search results and as a cover image in the video player.

internal/meta/data.go
```diff
@@ -68,6 +68,7 @@ type Data struct {
 	Error         error         `meta:"-"`
 	json          map[string]string
 	exif          map[string]string
+	EmbeddedVideo string `meta:"EmbeddedVideo"`
```
lastzero: I'd move `EmbeddedVideo` up e.g. between `Albums` and `Error` so that the private properties come last.

internal/photoprism/mediafile.go
```diff
@@ -314,6 +317,86 @@ func (m *MediaFile) EditedName() string {
...
+			log.Debugf("Error running exiftool on video file: ", err)
+			if stderr.String() != "" {
+				return "", errors.New(stderr.String())
+			} else {
+				return "", err
+			}
+		}
+
+		// find the extracted video path
+		savedPath := filepath.Join(
```
lastzero: Looking at the code, I'm still not 100% sure what the destination filename used by Exiftool is: Is it a temporary name? Does it check if it already exists? What happens in case of filename conflicts? Would it make sense to let Exiftool write to the cache path or to the tmp path, so that it doesn't write directly to the sidecar path (and then only move the video to the sidecar path if it was successfully extracted)?

internal/photoprism/mediafile.go
```diff
@@ -314,6 +317,86 @@ func (m *MediaFile) EditedName() string {
...
+			Config().ExifToolBin(), fmt.Sprintf("-%s", metaData.EmbeddedVideo),
+			"-b", "-w",
+			path, m.FileName(),
+		)
+
+		var out bytes.Buffer
+		var stderr bytes.Buffer
+		cmd.Stdout = &out
+		cmd.Stderr = &stderr
+		cmd.Env = []string{fmt.Sprintf("HOME=%s", Config().CmdCachePath())}
```
lastzero: `HOME=Config().CmdCachePath()` is a hack we originally added for Darktable because it requires the home directory to be writable even if no (or only temporary) files are created there. To get the configured tmp file path, you can use `Config().TempPath()`.

lastzero: Note that it doesn't hurt to keep this, even though it may not be required for Exiftool. It MAY be needed because Docker containers can run with UIDs for which no home (or actual user account) exists in the container.

internal/photoprism/mediafile.go
```diff
@@ -314,6 +317,86 @@ func (m *MediaFile) EditedName() string {
...
+			)
+			return "", nil
+		} else if extension := strings.Split(
+			mimeType, "/",
+		)[len-1]; extension != "" {
+			// rename the extracted video file with the correct extension
+			dir, file := filepath.Split(savedPath)
+			newFileName := fmt.Sprintf("%s.%s", file, extension)
+			newSavedPath := filepath.Join(dir, newFileName)
+			err := os.Rename(savedPath, newSavedPath)
```
lastzero: `Rename` fails when you try to move a file across different filesystems. We added a wrapper in the `fs` package that reverts to Copy & Remove in this case:
https://github.com/photoprism/photoprism/blob/1507525ba4a3ebfc96cb2c65ee9493740bd2cf59/pkg/fs/move.go#L10-L34

lastzero:
I'll merge this now so everyone can test it with our development preview build:
https://docs.photoprism.app/release-notes/#development-preview

Thank you very much! :gem:

lastzero:
There seems to be one last issue that we need to address, see this comment in the feature request:

- https://github.com/photoprism/photoprism/issues/439#issuecomment-1687782351

Issue #2337, opened 2022-05-20
Config: Add support for listening on Unix sockets


tushev:
Hello,

PhotoPrism currently supports only HTTP port communication. 

It would be nice to have UNIX socket support as well, because sockets:
1. eliminate latency from network stack => faster
2. are subject to file system permissions => better access control
3. will never lead to problems such as [docker+ufw port exposure flaw](https://stackoverflow.com/questions/30383845/what-is-the-best-practice-of-docker-ufw-under-ubuntu)

I've created a makeshift implementation of this using custom Dockerfile and `socat`.

While I'm currently happy with it, it would be nice to have this implemented natively, without the need of custom Dockerfiles.
 Something like:
```yaml
PHOTOPRISM_HTTP_MODE: 'socket'
PHOTOPRISM_HTTP_SOCKET_PATH: '/var/run/photoprism/photoprism.sock'
```
Finally, `/var/run/photoprism` should be made available on the host (or another container) via bind mount (or volume).

Unfortunately, I'm not familiar with Go so I cannot implement this myself and create a pull request.
<hr>

Below are details of my implementation. I've modified `docker-compose.yaml` to use a custom Dockerfile. I also added `env` variables.

The implementation is based on `socat`, which creates a connection between socket file inside container and PhotoPrism on port 2342 (inside container as well). Then the directory with the socket file is being exposed outside as Docker bind mount.

Finally, I've pointed my **nginx** reverse proxy to newly created socket file with<br>`proxy_pass http://unix:/home/photoprism/.socket/photoprism.sock`<br>instead of `        	proxy_pass http://127.0.0.1:2342`.

This solves (2) and (3). However, there's still some network stack communication inside container - between socat and PhotoPrism.
(1) can be solved only by [implementing HTTP UNIX socket listener natively in Go](https://noknow.info/it/go/implemented_http_server_that_supports_tcp_and_unix?lang=en).

**docker-compose.yaml:**
```yaml
services:
  photoprism:
    build:
      context: ./images
      dockerfile: Dockerfile-photoprism-socket
      args:
        SOCKET_UID:   '${SOCKET_UID}'
        SOCKET_GID:   '${SOCKET_GID}'
    ...
    environment:
      ...
      SOCKET_UID:   '${SOCKET_UID}'
      SOCKET_GID:   '${SOCKET_GID}'   
      SOCKET_MODE:  '${SOCKET_MODE}'
    volumes:
      ...
[...]

PR #3595, opened 2023-07-31
Backend: Support listening on Unix Socket (#2337)

SilverBut:
This patch implements the Unix Domain Socket listening. Users can prevent unnecessary port mappings if they want.

TLS or AutoTLS will not work since there is no TLS layer when using the unix domain socket.

Acceptance Criteria:

- [ ] (N/A, no frontend modifications. Existing functions are working) Frontend components must be responsive to work and look properly on phones, tablets, and desktop computers; you must have tested them on all major browsers and different devices
- [ ] (N/A) In case you submit database-related changes, they must be tested and compatible with SQLite 3 and MariaDB 10.5.12+

### DIFF
--- a/internal/commands/status.go
+++ b/internal/commands/status.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"time"
 
@@ -31,6 +32,15 @@ func statusAction(ctx *cli.Context) error {
 	// running after Get, Head, Post, or Do return and will
 	// interrupt reading of the Response.Body.
 	client := &http.Client{Timeout: 10 * time.Second}
+
+	// make a dial function for unix socket
+	if unixSocketPath := conf.HttpHostAsSocketPath(); unixSocketPath != "" {
+		client.Transport = &http.Transport{
+			Dial: func(network, addr string) (net.Conn, error) {
+				return net.Dial("unix", unixSocketPath)
+			},
+		}
+	}
 
 	url := fmt.Sprintf("http://%s:%d/api/v1/status", conf.HttpHost(), conf.HttpPort())
 
--- a/internal/config/config_server.go
+++ b/internal/config/config_server.go
@@ -104,7 +104,8 @@ func (c *Config) HttpCachePublic() bool {
 
 // HttpHost returns the built-in HTTP server host name or IP address (empty for all interfaces).
 func (c *Config) HttpHost() string {
-	if c.options.HttpHost == "" {
+	// when unix socket used as host, make host as default value. or http client will act weirdly.
+	if c.options.HttpHost == "" || c.HttpHostAsSocketPath() != "" {
 		return "0.0.0.0"
 	}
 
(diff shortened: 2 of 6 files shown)


SilverBut:
Hi guys. I'm not sure if there should be an automated unit test, since this code changes works mainly about interacting with system. But if it's necessary I can add it somehow.

Also I noticed documentations are in a separate repository. Maybe I can make PR in that repository after this one has been reviewed?

Thanks.

lastzero:
How about we reuse the existing host config option for this so that the list of options doesn't grow even further (some users already complain that it gets too complicated)?

SilverBut:
> How about we reuse the existing host config option for this so that the list of options doesn't grow even further (some users already complain that it gets too complicated)?

Oh that's better! I was doubtful about it would raise misunderstandings....

What about this way: if `HTTP_HOST` starts with `unix:` and `HTTP_PORT` is explicitly set to 0, listening on the socket path given? 

I think port set to 0 is necessary because some user may already using a resolvable dns name as `HTTP_HOST` before.

lastzero:
@SilverBut It might be enough to check if the host contains `/` to specify a path? A little more research would be needed to see if this is a common practice and if there are potential drawbacks, e.g. for security.

Alternatively, it would be good to shorten the config option to `PHOTOPRISM_HTTP_SOCKET` so that it has a similar naming scheme as in other apps like Docker (`DOCKER_SOCKET`) and MariaDB (`socket`).

Edit: Detecting or assigning the value 0 for the port is not good for Go apps like PhotoPrism, as it is the default value and therefore should be set to the documented default (2342) in this case.

lastzero:
Since you already have the code for a separate configuration option, I would suggest that you shorten it to `PHOTOPRISM_HTTP_SOCKET` and then we can see how to move ahead. It would also be good to know more about your use case, as I've never connected a browser to a Unix socket and wonder if/how that works. We also use websocket connections, so those must work as well. Our config examples should generally not include this option, as many users would not understand it and could break their installation.

SilverBut:
@lastzero Good idea about checking `/`. RFC 1123 says `/` is illegal in hostname, so I think `HTTP_HOST.include("/") && HTTP_HOST.startsWith("unix:")` is safe enough and won't break any existing user by accident. I will remove new code for additional options reading.

And some thing about unix socket:

* Browser's won't connect to unix socket directly, as it would breaking lots of stuff (for example, how to check CORS policy). Instead, unix socket is a good option if application is hide behind some reverse proxy like Caddy or Nginx. 
* Security policy (like SELinux) can be easier implemented by controlling which application can r/w the socket file. Wrong configured port mapping may also cause unexpected access.
* By not exposing a port, no firewall rules or route rules is necessary. This can be helpful if user don't want to let docker touching network stuff. Also, not exposing port can prevent some weird issues caused by NAT, like [running out of local ports](https://blog.cloudflare.com/conntrack-tales-one-thousand-and-one-flows/).
* It could also bring slightly performance advance than listing on a local port and forward by Nginx, because lots of network stuff can be skipped. However, comparing with the real time used by real http handler, this performance advance can be ignored.

I agree with not putting this in config examples. But maybe we can mention it in the [config options](https://docs.photoprism.app/getting-started/config-options/) for documentation purpose? Also the [Advanced Nginx proxy setup](https://docs.photoprism.app/getting-started/advanced/nginx-proxy-setup/) can add a small paragraph about it.

I have tested and WebSocket works fine if reverse proxy is correctly configured with WebSocket support (just like plain HTTP does).

lastzero:
The docs at https://docs.photoprism.app/getting-started/config-options/ are automatically updated from the source code, so you only need to change the usage description.

SilverBut:
Changed to reflect new ideas, also added some tests about config.

lastzero:
Thank you very much! I'll merge this now :+1:

Issue #3168, opened 2023-02-04
Frontend: Upgrade Vuetify from v1.5 to v3.7


dennisameling:
### Description

**As a developer, I want to use the current stable versions of Vue and [Vuetify](https://vuetifyjs.com/en/getting-started/upgrade-guide/#setup) so that I have access to the latest features, user interface components, and bug fixes.**

### Acceptance Criteria

- [x] Vue and Vuetify MUST be upgraded to the latest stable versions, along with all required dependencies
- [x] All currently implemented functionality MUST continue to work as before
- [x] User interface performance and loading time MUST NOT be negatively affected by the upgrade
- [x] The upgraded frontend MAY look different based on the design changes introduced with Vuetify 3
- [x] The upgraded frontend MUST work in the latest Firefox, Safari and Chrome browser versions
- [x] The upgraded frontend MUST be fully responsive for compatibility with mobile and desktop devices
- [x] Existing unit and acceptance tests MUST run successfully without errors
- [ ] Existing documentation, e.g. in the developer guide, SHOULD be updated as needed
- [x] The browser console SHOULD not display any errors or warnings
- [x] Any remaining issues that are not blockers MUST be fully documented so that they can be resolved later

####  Frontend Templates and Components

The following user interface templates and components need to be checked for compatibility with Vue and Vuetify 3 and, if necessary, updated to ensure they continue to work as expected:

- [x] [Application Layout](https://github.com/photoprism/photoprism/blob/develop/frontend/src/app.vue)
  - [x] `v-app`
  - [x] `v-content`
- [x] [Components](https://github.com/photoprism/photoprism/tree/develop/frontend/src/component)
  - [x] [PNavigation](https://github.com/photoprism/photoprism/blob/develop/frontend/src/component/navigation.vue) in "component/navigation.vue"
  - [x] PNotify in "component/notify.vue"
  - [x] PLoadingBar in "component/loading-bar.vue"
  - [x] PScrollTop in "component/scroll-top.vue"
  - [x] PAuthHeader in "component/auth/header.vue"
  - [x] PAuthFooter in "component/auth/footer.vue"
  - [x] PPhotoViewer in "component/photo-viewer.vue"
  - [x] PVideoPlayer in "component/video/player.vue"
  - [x] PPhotoToolbar in "component/photo/toolbar.vue"
  - [x] PPhotoCards in "component/photo/cards.vue"
  - [x] PPhotoMosaic in "component/photo/mosaic.vue"
[...]

Issue #1131, opened 2021-03-10
UX: Enhance workflow with keyboard shortcuts


Zaijo:
As a photoprism user I want to use keyboard shortcuts to perform frequent actions in my typical workflow to work faster.

Suggestions:

- <kbd>space</kbd> to **select** a picture (to current selection) so I can decide to insert into album or archive later
- <kbd>d</kbd> or <kbd>r</kbd> quickly **archive** selected pictures (probably with a prompt)
- <kbd>a</kbd> open "add to **album** dialog" for selected pictures
- <kbd>g</kbd>: Download (Get) images
- <kbd>s</kbd>: Share images
- ctrl+click to open in new tab (https://github.com/photoprism/photoprism/issues/948)
- Use  arrows to navigate between photos in edit mode (https://github.com/photoprism/photoprism/issues/1593)
Also introduce a typical "keyboard shortcuts" overlay screen, that would offer quick reference.

This issue should gather all the shortcuts requests mentioned in e.g. #508

bernsel:
We would be happy to help fund this as well. What would it take to get it added to the roadmap? Or could we hire a developer, and pay them for their time to create this module for the project? With your input and guidance of course, so that whatever is developed fits the overall vision of the project.

lastzero:
See https://github.com/photoprism/photoprism/pull/743

abhchand:
To add another suggestion, it would be great to navigate through photos with `j` and `k` in addition to arrow keys. While the origins are related to `vim`, these shortcuts have found their way in to a ton of applications. 

I'm considering opening a PR for this if that's ok. Just wanted to throw it out as an idea for feedback as well.

Thanks!

CWempe:
I would like to use `ESC` to ~close the photoviewer or~ go back to the previous page.

I am using this often in other apps.

FollowTheWizard:
Have you considered using the access key option from the browser to provide a quick implementation?

Link to doucmentation:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/button?retiredLocale=de

In my opinion those options would already be great to have
**s**elect
**l**ike
**a**rchive
**d**ownload

artincube:
I'd love to see some. Next/Prev image in Faces and any other situation where possible. 
Then
select
edit
like
archive

anst84:
[...]

PR #3623, opened 2023-08-12
UX: Use arrow keys to navigate in edit dialog

tomplast:
**What does it implement / fix / improve? Why?**
I thought it would be useful (and intuitive) if you could use the keyboard arrows to navigate to previous and next image while in the photo edit dialog.

**Are the changes related to an existing issue?**
Nope.


graciousgrey:
Thank you very much for your contribution! We will have a look at this once the recently merged pull requests have been successfully released :)

tomplast:
@graciousgrey Any updates :) ?

lastzero:
Please note that we are out of the office until October 3 to get some rest after working on the new release. 🏝️💬

graciousgrey:
The first tests look good so far :) It works in all browsers I have tested. 
I just noticed that sometimes I have to click into the edit dialog before I can use the arrow keys.

graciousgrey:
@tomplast We found one thing we need to resolve before we can merge this: The app must only navigate to the next edit dialog when the focus is not on an input field.

tomplast:
@graciousgrey Functionality wise I think have fixed the issues you mentioned.

graciousgrey:
Thank you! I will test your changes as soon as possible :)

graciousgrey:
I just tested your latest changes. 

> I just noticed that sometimes I have to click into the edit dialog before I can use the arrow keys.

I have not observed this anymore :tada: 

>The app must only navigate to the next edit dialog when the focus is not on an input field.

I still see this behavior. The strange thing is that sometimes it works and sometimes it doesn't. One way around this could be to use a key combination that has no other function assigned to it, such as navigating in an input field. For example: shift key + arrow key.

I also realized that we need to prevent navigating to the next edit dialog when there are unsaved changes. For example, if a user has edited a field and not clicked Apply or Done, the changes will be lost when navigating to the next edit dialog. We could either prevent navigation without a message, or add a dialog that informs the user that there are unsaved changes and lets him save or discard them.

tomplast:
Sorry for the late reply, I will try to look into it the coming week.

graciousgrey:
No need to hurry :) We are currently busy with 2FA and OpenID Connect support, so we only have limited time to review changes over the next weeks.

lastzero:
I'd love to give this a try now that the frontend has been [upgraded from Vuetify v1.5 to v3.7](https://github.com/photoprism/photoprism/issues/3168#screenshots)! :ok_hand:

lastzero:
@tomplast Thanks again for your contribution! Based on your request and proof-of-concept, the above commit adds support for navigating with the arrow keys in the edit dialog.

To ensure that this only happens when no input element has focus, navigating with the keys is only possible when the active element is a div, the dialog is active, and the model data is unchanged.

Since we also support right-to-left languages like Arabic and Hebrew, another thing to keep in mind was to flip left and right when using the keys.

Related Issues:

- #1131

I'm sorry we couldn't take care of this sooner! But now it's finally done, so we can close the PR (since it's no longer needed) and set the status to Preview for final testing.

lastzero:
I've created a follow-up issue to take care of the remaining issues we found while testing (getting help from our community for this would be awesome):

- https://github.com/photoprism/photoprism/issues/4916

lastzero:
It seems that the arrow keys stopped working after pressing the Apply button because disabled elements cannot have focus, i.e. the browser automatically focused the `<body>` element instead without triggering an additional *focusout* event. The above changes handle this and other situations, and should prevent the edit and lightbox dialogs from losing focus when active. Any help testing this with our preview build would be much appreciated! ✅

PR #3636, opened 2023-08-16
Places: Fix markers not updating correctly

reubendowle:
This should fix the issue with the clusters not loading quickly etc.

https://github.com/photoprism/photoprism/issues/1187

I have not properly tested this as my photoprism install has corrupted itself after my PC ran out of disk space. I will test it properly tomorrow, although I am pushing this as a draft if someone else can do a quick test.

Acceptance Criteria:

### DIFF
--- a/frontend/src/page/places.vue
+++ b/frontend/src/page/places.vue
@@ -635,7 +635,14 @@ export default {
         }
       });
 
-      this.map.on('idle', this.updateMarkers);
+      // Following example for cluster updates from here: https://maplibre.org/maplibre-gl-js/docs/examples/cluster-html/
+      this.map.on('data', (e) => {
+          if (e.sourceId !== 'photos' || !e.isSourceLoaded) return;
+
+          this.map.on('move', this.updateMarkers);
+          this.map.on('moveend', this.updateMarkers);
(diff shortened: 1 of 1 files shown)


frontend/src/page/places.vue
```diff
@@ -635,7 +635,14 @@ export default {
         }
       });
 
-      this.map.on('idle', this.updateMarkers);
+      // Following example for cluster updates from here: https://maplibre.org/maplibre-gl-js/docs/examples/cluster-html/
+      this.map.on('data', (e) => {
+          if (e.sourceId !== 'photos' || !e.isSourceLoaded) return;
+
+          this.map.on('move', this.updateMarkers);
```
lastzero, line 642: Wouldn't this attach the event handlers every time the results are updated, for example when the user performs a search? In this case, I would expect the map to slow down with each search performed.

lastzero: Related Documentation: https://maplibre.org/maplibre-gl-js/docs/API/types/maplibregl.MapDataEvent/

Reads like that could happen. The example they provide might be a special case, in case the data is loaded once. If you don't have time to test it, I can do it later today.

PR #3648, opened 2023-08-18
Places: Fix cluster update callback function

reubendowle:
This is an attempt to fix the latest issues noted where the places photo clusters (https://github.com/photoprism/photoprism/issues/1187) can sometimes render in wrong locations etc.

It looks like the reason this sometimes occurs is that the cluster updates happen in an asyncronous callback, and that processing can be triggered multiple times in parallel in some cases if the processing takes longer than the time between multiple map events.

This is difficult to reliably reproduce, however I have been unable to reproduce any issues since making this change.

When testing, I consistently saw a cluster from USA appear when zooming into Tübingen:

image

With the fixed code, this consistently renders correctly:

image

Acceptance Criteria:

### DIFF
--- a/frontend/src/page/places.vue
+++ b/frontend/src/page/places.vue
@@ -59,6 +59,7 @@ export default {
       markersOnScreen: {},
       clusterIds: [],
       loading: false,
+      markerPromise: null,
       style: "",
       terrain: {
         'topo-v2': 'terrain_rgb',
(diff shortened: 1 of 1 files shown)


lastzero:
Thanks! I'll give that a try :)

From how the cluster update code looks, one problem could be that it is a "two-pass" approach where the features are queried multiple times, i.e. first the cluster ids and then the related images via `getMultipleClusterFeatures()`, so the data could have changed in between? 🤔

lastzero:
@reubendowle Feel free to email us if you would like to join our Contributors Chat! 💬

lastzero:
Works locally, although it's hard to see if everything is correct because I have too many pictures...

@reubendowle @heikomat I've started a new preview build so you can [test it in our demo](https://demo.photoprism.app/library/places). It's easy to see there if the number of pictures in the clusters adds up when zooming in and out! :nerd_face:

lastzero:
@reubendowle One thing I noticed is that when I zoom or move the map, the clusters don't update simultaneously and even after I stop doing anything, it can take a while for the new clusters to be visible and the old ones to disappear. With your changes, would it help to also listen for the `move` event instead of just `moveend`? Based on my tests, it hasn`t made any difference so far (just caused more computational overhead).

reubendowle:
> One thing I noticed is that when I zoom or move the map, the clusters don't update simultaneously and even after I stop doing anything, it can take a while for the new clusters to be visible and the old ones to disappear. With your changes, would it help to also listen for the move event instead of just moveend? Based on my tests, it hasn`t made any difference so far (just caused more computational overhead).

In theory if the time it takes to update the clusters is very short, then implementing move would be a good idea. This fires multiple times as the map is scrolling, zooming, etc, so would allow the clusters to stay closely in sync with the map itself.

But if the processing time is longer, then you might just end up taking longer to get to the final clusters when the zooming has stopped - if the processing of clusters for an intermediary map step delays the processing for the final map step.

I really don't know what the best answer is here. Perhaps on fast desktops move might be good, and perhaps on mobile it would be worse. But the only way to know properly would be with a lot of testing. So I think just updating once on moveend is probably good as a safer/conservative option unless we are sure the other way is better.

PR #3655, opened 2023-08-20
Add batch editing #271

vidaren:
This PR should solve issue #271 . – *except* editing of labels.
This PR implements batch editing: Editing of multiple properties for multiple photos in the details page in one operation:
- Make the changes to properties on one of the photos
- Click/Press and *hold* the APPLY or DONE button until a confirmation dialog appears
The long-press method was chosen because it requires no UI for small devices.

Acceptance Criteria:

- [X] Automated unit and/or acceptance tests are mandatory to ensure the changes work as expected and to reduce repetitive manual work
- [X] Frontend components must be responsive to work and look properly on phones, tablets, and desktop computers; you must have tested them on all major browsers and different devices

Tested on MariaDB only, but I believe GORM will handle the changes for SQLite too.


graciousgrey:
Thank you very much for your contribution! We look forward to working with you on the batch edit feature once the recently merged pull requests have been successfully released :)

It might not be ideal to reuse the existing edit dialog. Instead we had planned to add a separate batch edit dialog to avoid confusion.

The following use cases will likely require additional consideration:

**(a) Editing the date of a photo.**

Suppose a user selects two photos that currently have the following date set:
- Photo 1 day:01 month:08 year:2022
- Photo 2 day:05 month:07 year:2021

Therefore, he opens the edit dialog of photo 1 and changes day:01 to day:03 and year:2022 to year:2020 and long clicks Apply to apply the changes to all photos. Now the user expects both photos to have the following date: day:03 month:08 year:2020. Instead, he gets the following result:

- Photo 1 day:03 month:08 year:2020
- Photo 2 day:03 month:07 year:2020

**(b) Adding keywords to photos**

Suppose a user selects two photos that currently contain the following keywords:
- Photo 1 keywords: banana, yellow
- Photo 2 keywords: apple, red

Therefore, he opens the edit dialog of photo 1 and changes the keywords: banana, yellow to keywords: banana, yellow, fruit and long clicks Apply to apply the changes to all photos. As a user I would now expect the following 

- Photo 1 keywords: banana, yellow, fruit
- Photo 2 keywords: apple, red, fruit

Instead I get 

- Photo 1 Keywords: banana, yellow, fruit
- photo 2 keywords: banana, yellow, fruit

An additional benefit of having a separate Batch Edit dialog is that users can always see exactly which changes are applied to their images. This is not the case with the current implementation, as the user has to remember exactly which fields he just changed and which were already set before. 

Also, in a separate batch edit dialog we can add additional fields that are not present in the normal edit dialog, for example, a field to add labels and a field to remove labels.

vidaren:
@graciousgrey You raise some good points in (a) and (b). My implementation is a bit crude that way, not smart at all, only taking the new field values and storing them on all selected photos. It was written for my personal need, adding coordinates and country to groups of photos that don't have them, which it does well. I then included the rest of the fields on the Details page for completeness, without thinking much about the use cases you mention.

Removing date and keywords from the fields that can be batch-edited makes sense, perhaps: Just don't batch-save the changes made (not the best UX, I know.) At least the date does not need batch-editing, or editing at all, most of the time.

sclu1034:
The way I usually see bulk editing implemented is that in the edit dialog, every field gets additional, virtual values `<keep>` and `<clear>` (or variations thereof). And the logic is

- all fields in the bulk edit dialog start with a default value of `<keep>`
  - unless you have all data for all entries loaded, in which case you can also differentiate between:
    - if every entry has the same value in a field, pre-fill the field in the dialog with that value
    - otherwise pre-fill with `<keep>`
- on submit, for every dialog field
  - if the value is `<keep>`, each entry's individual value in that field is kept unchanged
  - if the value is `<clear>`, apply the default value for that field on every entry
  - otherwise apply the given value to every entry

And for array-like fields (like keywords and labels), bulk edit dialogs usually have two input fields as graciousgrey mentioned: One "add values" and one "remove values". So with example **(b)** from above, the logic would be

```
Add keywords: fruit
Remove keywords: yellow, green
```

Before

- Photo 1 keywords: banana, yellow
- Photo 2 keywords: apple, red

After

- Photo 1 keywords: banana, fruit
- Photo 2 keywords: apple, red, fruit

Other additional fields that a dedicated bulk edit dialog could have, would be things from the last tab (Orientation, Panorama, Favorite, etc.).
And while the "long press opens confirmation" flow works, a proper "apply to all" button in that dedicated dialog would be a lot more intuitive. In my experience, people don't just peruse whole documentation pages to discover something like "this one particular button actually has additional functionality if you press and hold".

graciousgrey:
I'm going to close this as our team will be working on this with a different UX approach (see related issue) than the one chosen in this pull request. Thank you for your contribution!

sclu1034:
@graciousgrey  Since the other issue is locked, and this thread here has been discussing UI already, I'll hijack it for feedback on your preview:

You mention additional options like "append" or "prepend" per text field. I find that the preview you shared is already filled enough with fields and that an additional dropdown or set of buttons _per field_ would add a lot of clutter.

Maybe you already have ideas for that in mind, but my proposal would be this:

When the form is opened, every text field is pre-filled with a placeholder like `{{value}}`. When the form is sent, every text field is run through the simple operation of `field.replace('{{value}}', original_value)` (once per entry in the batch). This way, you get all possible operations without any additional buttons or dropdowns:

- retaining the original value is the default, like `<keep>` in my comment above
- an empty field means clearing it, like `<clear>` above or the "delete" option you mention
- append and prepend, or a mix thereof, can be achieved by placing `{{value}}` anywhere (e.g. `my prefix-{{value}}-my suffix`)
- replacing the value is achieved by omitting `{{value}}` entirely

graciousgrey:
Thank you  for your feedback! We don't have a final solution for this in mind yet. As you say, adding dropdowns could clutter the UI, on the other hand the approach you mentioned requires the user to know the exact syntax. 

We plan to release the basic version of batch editing first. After that we will create follow up issues for enhancements.

PR #3657, opened 2023-08-21
Places: Fix clusters showing in the wrong location

reubendowle:
This should fix the issues with clusters being shown in the incorrect location as pointed out in https://github.com/photoprism/photoprism/issues/1187.

From by debugging, it appears that cluster Ids must be being re-used sometimes, because the cached cluster markers are not always valid. Previously a cluster marker rendered for one location was just being moved to a new geo location.

This now changes the code to remove caching of the cluster markers, and always regenerates them whenever the visible cluster IDs change.

Acceptance Criteria:

### DIFF
--- a/frontend/src/page/places.vue
+++ b/frontend/src/page/places.vue
@@ -55,7 +55,6 @@ export default {
       canSearch: this.$config.allow("places", "search"),
       initialized: false,
       map: null,
-      markers: {},
       markersOnScreen: {},
       clusterIds: [],
       loading: false,
@@ -543,7 +542,7 @@ export default {
 
       this.markerPromise = new Promise((resolve, reject) => {
         // Find clusters.
-        let features = this.map.querySourceFeatures("photos");
+        const features = this.map.querySourceFeatures("photos");
 
         const clusterIds = [...new Set(features
           .filter(feature => feature.properties.cluster)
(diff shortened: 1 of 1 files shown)


lastzero:
Did you check how this impacts performance?

lastzero:
I've merged your changes for testing them in our develop branch :+1: 

When I have a bit of time, I'll see if I can find additional optimizations.

reubendowle:
> Did you check how this impacts performance?

Subjectively it did not seem much different, but I have not done any proper performance testing.

reubendowle:
> When I have a bit of time, I'll see if I can find additional optimizations.

I put some console prints at a couple of points in this function, and on my PC the getMultipleClusterFeatures call is what takes almost all the time. The first call to this at a particular zoom level often takes about 300ms for me (and in some extreme cases it can take nearly 2 seconds), but once that returns data the rest of getMultipleClusterFeatures only takes about 2ms to complete.

I dug a bit deeper than that too. As far as I can see, it is the first call to getClusterFeatures() that blocks. Then the subsequent calls return quickly.  Once you have been to a location/zoom level, the calls seem to start returning much faster.

So it appears almost all of the time is just waiting for the map component to recalculate which items go into each cluster. I doubt there is much we can do to make any significant impact here on performance. Certainly it looks to me like trying to optimize the logic in updateMarkers is not going to help.

lastzero:
@reubendowle @heikomat Having no (known) bugs anymore is a big improvement! I just tested this with my personal library (~100k photos) and couldn't see much of a difference in performance.

From a usability perspective, marker updates on the map should generally be smoother, e.g. when zooming. I'm not sure if/how this can be done with our approach. Perhaps it is possible to refactor the code to use fewer loops or other API calls? It might be worth looking at how other teams have implemented similar solutions to get ideas for further optimization...

lastzero:
With these fixes, all markers - clusters and photos - finally work in my local environment and on our demo. I have started a new preview build so we can do more testing. It should be available within the next hour:

PR #3660, opened 2023-08-23
Samsung: Improved support for Motion Photos

gokcedilek:
As a follow-up to https://github.com/photoprism/photoprism/pull/3588
- check if a video sidecar file already exists before extracting the embedded video
- check if the original file was created by a supported camera make

### DIFF
--- a/internal/photoprism/mediafile.go
+++ b/internal/photoprism/mediafile.go
@@ -332,6 +332,17 @@ func (m *MediaFile) ExtractEmbeddedVideo() (string, error) {
 
 	// Get the embedded video field name from the file metadata.
 	if metaData := m.MetaData(); metaData.Error == nil && metaData.EmbeddedVideo != "" {
+		// Check if an embedded video sidecar already exists
+		if embeddedVideo := m.HasEmbeddedVideo(); embeddedVideo != "" {
+			return embeddedVideo, nil
+		}
+
+		// Check if the camera make is supported
+		if _, ok := entity.CameraMakes[metaData.CameraMake]; !ok {
+			return "", errors.New("unrecognised camera make")
+		}
+
+		// Extract the embedded video
 		outputPath := filepath.Join(Config().TempPath(), m.RootRelPath(), "%f")
 		cmd := exec.Command(Config().ExifToolBin(),
 			fmt.Sprintf("-%s", metaData.EmbeddedVideo), // TODO: Is this safe?
--- a/internal/photoprism/mediafile.go
+++ b/internal/photoprism/mediafile.go
@@ -333,13 +333,8 @@ func (m *MediaFile) ExtractEmbeddedVideo() (string, error) {
 	// Get the embedded video field name from the file metadata.
 	if metaData := m.MetaData(); metaData.Error == nil && metaData.EmbeddedVideo != "" {
 		// Check if an embedded video sidecar already exists
-		if embeddedVideo := m.HasEmbeddedVideo(); embeddedVideo != "" {
+		if embeddedVideo := m.SidecarFileName(fs.VideoMP4); embeddedVideo != "" {
 			return embeddedVideo, nil
-		}
-
-		// Check if the camera make is supported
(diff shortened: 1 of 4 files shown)


internal/photoprism/mediafile_meta.go
```diff
@@ -53,6 +53,15 @@ func (m *MediaFile) NeedsExifToolJson() bool {
 	return !fs.FileExists(jsonName)
 }
 
+// Checks if an .mp4 video file exists for this file in the sidecar path
+func (m *MediaFile) HasEmbeddedVideo() string {
+	sidecarFile := filepath.Join(Config().SidecarPath(), m.RootRelPath(), m.BasePrefix(false)+fs.ExtMP4)
```
lastzero: The `fs` package includes a function that returns the sidecar filename for a file type, if exists:
```go
fs.VideoMP4.FindFirst(m.FileName(), []string{Config().SidecarPath(), fs.HiddenPath}, Config().OriginalsPath(), false)
```

internal/photoprism/mediafile_meta.go
```diff
@@ -53,6 +53,15 @@ func (m *MediaFile) NeedsExifToolJson() bool {
 	return !fs.FileExists(jsonName)
 }
 
+// Checks if an .mp4 video file exists for this file in the sidecar path
+func (m *MediaFile) HasEmbeddedVideo() string {
```
lastzero: How about a more general method to find sidecar files of a certain type?

```go
func (m *MediaFile) SidecarFileName(t fs.Type) string {
  return t.FindFirst(m.FileName(), []string{Config().SidecarPath(), fs.HiddenPath}, Config().OriginalsPath(), false)
}
```

Note that the MP4 can also be in the originals path, e.g. if it was extracted manually or if the originals path is the same as the sidecar path.

internal/photoprism/mediafile.go
```diff
@@ -332,6 +332,17 @@ func (m *MediaFile) ExtractEmbeddedVideo() (string, error) {
 
 	// Get the embedded video field name from the file metadata.
 	if metaData := m.MetaData(); metaData.Error == nil && metaData.EmbeddedVideo != "" {
+		// Check if an embedded video sidecar already exists
+		if embeddedVideo := m.HasEmbeddedVideo(); embeddedVideo != "" {
+			return embeddedVideo, nil
+		}
+
+		// Check if the camera make is supported
+		if _, ok := entity.CameraMakes[metaData.CameraMake]; !ok {
```
lastzero: `entity.CameraMakes` is just a lookup table for normalizing the vendor string. It might be best to encapsulate the validation in the `meta` package, so that `metaData.EmbeddedVideo` is only set if the original media file can (optionally) contain an MP4 video. If you need a list of vendor strings (or other metadata properties) to check against, it would make sense to add these to the `meta` package as well, e.g. as string constants or a `map[string]bool` lookup table.

internal/photoprism/mediafile.go
```diff
@@ -332,6 +332,17 @@ func (m *MediaFile) ExtractEmbeddedVideo() (string, error) {
 
 	// Get the embedded video field name from the file metadata.
 	if metaData := m.MetaData(); metaData.Error == nil && metaData.EmbeddedVideo != "" {
+		// Check if an embedded video sidecar already exists
+		if embeddedVideo := m.HasEmbeddedVideo(); embeddedVideo != "" {
```
lastzero: Although the related MP4 video may also be in the originals path, e.g. if it was extracted manually or if the originals path is the same as the sidecar path, `ExtractEmbeddedVideo()` should only return a filename here if the MP4 is in the sidecar path, since `MediaFile.RelatedFiles()` will automatically find it in the originals path. Otherwise, if we return the name of any MP4 sidecar file, the same file may appear twice in `result.Files` - which would be wrong. Alternatively the handling in `RelatedFiles()` must be modified.

If this becomes too complex and confusing, just ignore it and I'll help you when I get back!

graciousgrey:
Thank you so much for your contribution! Since we recently released this feature, I'm closing this PR.

PR #3696, opened 2023-09-02
Backend: Suport systemd socket activation

PhracturedBlue:
systemd supports passing an already open TCP or unix socket to an application on startup (enabled via the LISTEN_FDs env variable).  Podman supports passing this socket from the host to the container as of version 3.4 (Docker does not support socket activation for containers).  Socket activation provides the benefit of only starting a service on demand (reducing overall boot time and potentially reducing memory usage), as well as improving security by allowing running containers with '--network=none' and still being able to be exposed.

This PR adds automatic socket-activation support.  If the 'LISTEN_FDS' variable is set, socket activation will automatically be used, otherwise the behavior remains unchanged.  Socket activation can work with unix domain sockets or TCP sockets, and will work with TLS if configured

Acceptance Criteria:

- [?] Automated unit and/or acceptance tests are mandatory to ensure the changes work as expected and to reduce repetitive manual work
- [N/A] Frontend components must be responsive to work and look properly on phones, tablets, and desktop computers; you must have tested them on all major browsers and different devices
- [N/A] In case you submit database-related changes, they must be tested and compatible with SQLite 3 and MariaDB 10.5.12+

I will post a separate PR for documentation updates.  There is currently no test infrastructure for the server component that I see, so I'm not sure what to do about writing tests.

This patch does not support AutoTLS.  In theory it should be possible to use autotls if 2 ports are socket-activated, but I don't see how AutoTLS can even work today as it opens 2 listeners on the same port, which I believe should fail.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,7 @@ module github.com/photoprism/photoprism
 
 require (
 	github.com/araddon/dateparse v0.0.0-20210429162001-6b43995a97de
+	github.com/coreos/go-systemd/v22 v22.5.0
 	github.com/disintegration/imaging v1.6.2
 	github.com/djherbis/times v1.5.0
 	github.com/dsoprea/go-exif/v3 v3.0.1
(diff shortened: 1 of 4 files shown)


lastzero:
@PhracturedBlue Thank you for your contribution! ⭐

My apologies for not getting back to you any sooner. Unfortunately, this proposal did not make it into our stable release, and we have since reworked much of the related code. I will therefore close this PR as it is no longer applicable.

PR #3709, opened 2023-09-06
Google: Initial support for Motion Photos

gokcedilek:
Issue #1739
- Search for the key `MotionPhoto` or `MicroVideo` in the exifdata
- If the key exists, extract the embedded video in the image: 
  - Find the byte offset of the `ftyp` header in the file
  - Extract the contents of the file starting from this byte into the video file  
  - Add the video file to the list of related files

PR #3722, opened 2023-09-13
Google: Add support for Motion Photos

### LINKED ISSUE #1739: Live Photos: Add support for Google Camera Motion Photos
Similar to how you implemented the Apple Live Photo (#177) and to the currently developing #439.

Google Camera has its way of creating Motion Photos: it makes a JPEG and then the MP4 inside the same file.

The "Motion Photo" is identified by:

    XMP:MicroVideo: 1 <- Before Android 11
    XMP:MotionPhoto: 1 <- After Android 11

Extracting the video is straightforward and doable in a few bash commands:

    srcfile=image
    offset=$(grep -F --byte-offset --only-matching --text ftyp "$srcfile")
    offset=${offset%:*}
    dd if=$srcfile of=$srcfile.mp4 bs=$((offset-4)) skip=1

*Note:* I tested it only with Android 12 Motion Pictures. Moreover,  versions <11 had the offset defined in an XMP.

rconn01:
This pull request adds support for google motion photos, which is resolves #1739 . It does so by using exiftool to find the video offset of the image. The approach here is a little hacky because it injects a field into the json sidecar file. This is done because the file contains a jpg followed by an mp4, so there are duplicate metadata keys. To get the offset we need the second `DirectoryItemLength` key. So exiftool is run a second time, not in json mode, so it can get that and inject it into the json. 

This is a little different than the approach mentioned in issue #1739 and was done to favor performance especially for smaller devices like raspberry pi. With this approach we don't need to walk all the bytes in the file looking for magic bytes and can seek right to the offset.

I've tested this on a pretty large collection of motion videos and seems to work. As part of the testing I noticed a few improvements to motion photos in general, which are also part of this MR. Specifically
1. making sure we always set the video part as the main photo
2. Changing the offset on preview images so a preview can always be generated. I ran into a number of cases where an empty preview would be generated resulting in the related video not getting indexed at all

The implementation here follows what was done for other live photo types, but I do have some general questions that might be worth changing in other change sets

1. Why are we putting the extracted video in `sidecar` vs `cache`. Since its fully recreatable like the thumbnails wouldn't cache be a more logical place?
3. Should we be generating preview images at all for the video part of live photos. Effectively the image part is already the preview, generating another jpg from the video part feels like wasted effort and space. This would be a little bit of a larger change but I think might be worth exploring (and happy to contribute it as another change set if we agree) 
Acceptance Criteria:


lastzero:
Thank you very much for your contribution! @gokcedilek has already worked on Google Motion Photos in #3709. From what I know, she found a way to extract the videos without using Exiftool. It therefore seems best if she reviews your PR and provides feedback :)

gokcedilek:
@rconn01 thank you for your implementation! I have also encountered the approach for reading the second `DirectoryItemLength` key, but we thought of not proceeding with this approach as it requires running exiftool a second time after the json is generated. My current PR extracts the video by searching for the MP4 byte header in the file, however I'll work on optimizing this approach so that we read in chunks or use binary search on the file, as pointed out by @lastzero. Please let us know if you have any suggestions regarding this issue.

rconn01:
What's the concern of running exiftool a second time only in the cases that motionPhoto is detected. There is no scanning of the file needed in that case as we know the exact offset to consume the file from. Exiftool is a pretty cheap operation. The big concern I can think of is going out of process, but that's already happening.

rconn01:
Ok so looking a little further, when you said extract videos without running Exiftool a second time @lastzero was that in response to the current impl for samsung photos, where it runs it to extract the data in binary mode?  I actually think that can totally be removed in favor of the impl that is here. Atleast when looking at the sample photo, as it has a `MicroVideoOffset` which this accounts for. I just did a quick test with the sample photo and removing all of this still works https://github.com/rconn01/photoprism/commit/36c71f3f33fefb16572e34e0f47f18375861e830 

Which means for Google motion v1 and samsung motion we are only running exiftool once. For google motion v2 we run it twice to get the offset key, but we are running it back to back in the same place just with two different flags so we can get more keys. And the result is persisted as the sidecar file. Personally I think this tradeoff is worth it vs searching for magic bytes in the file. Especially since we are already going out of process already. I can run some tests tonight to see if there are actual measurable numbers in performance difference, if there isn't then its probably a mute point.

Regardless which path is chosen for extraction, I do think my other changes are great improvements with all the testing I have done. Additionally I thing its worth discussing some of the questions I had in the MR description for future improvements as well and then we can align who picks those up (if we agree)

kvalev:
@rconn01 Its not really needed to run exiftool twice, by specifying `-all -struct` you can extract duplicate entries, see
https://github.com/kvalev/photoprism/pull/19/files#diff-8bcb41048113424bf9549028714449910c19e443c9a7dbc70670417964418e93R157

However @gokcedilek is right, at the end of the day you will need to find the mp4 file, because plenty of times the offset in the exif data is garbage, so you cannot rely on it.

rconn01:
Ahh I see, you get a structure like this
```"XMPToolkit": "Adobe XMP Core 5.1.0-jc003",
  "MotionPhoto": 1,
  "MotionPhotoVersion": 1,
  "MotionPhotoPresentationTimestampUs": 1033133,
  "HasExtendedXMP": "647E9AB0DF312CB8B2BC69B4E817E036",
  "Directory": [{
    "Item": {
      "Length": 0,
      "Mime": "image/jpeg",
      "Padding": 0,
      "Semantic": "Primary"
    }
  },{
    "Item": {
      "Length": 3773267,
      "Mime": "video/mp4",
      "Padding": 0,
      "Semantic": "MotionPhoto"
    }
  }],
```

That is certainly better than running it twice. we would just need to take the `MotionPhoto` semantic. We'd probably want to add some parsing to flatten anything that isn't `directoryItem` just so other things wouldn't break if there happens to be duplicates.

When you say the offset is garbage do we have examples of that. I've run across hundreds of legacy and current google motion photos and have not run into any issues where an mp4 wasn't extracted. 

If we add the parsing of the `-struct` format then we simply just have the few lines [here](https://github.com/photoprism/photoprism/pull/3722/files#diff-e5a39f92547299cd358543ff78e747595e351aa0f56a50d43c0034cb94465d8aR389-R390) and no extra anything

```
		videoStart := m.FileSize() - metaData.EmbeddedVideoOffset
		reader := io.NewSectionReader(srcFile, videoStart, m.FileSize())
```

gokcedilek:
I'm also curious to know about under which scenarios the offset in the exif data is garbage (@kvalev if you could share more information about that, that would be great:). I think it's a great idea to compare the performance improvements for going with the `DirectoryItemLength` approach versus the finding the MP4 header approach. I need to make improvements to my PR as I mentioned above, but I'm actually on a trip for the next two weeks and am not able to focus on this as much as I want to during this time. Also, just out of my own curiosity, for learning, when you mention the hundreds of tests, are you running these manually? Or do you have an automated testing script, other than the [unit tests](https://github.com/photoprism/photoprism/pull/3588/files#diff-4b9e845ea8615a5f14c5e74cbd56114a00c96c04eba7f38f14ef979868a1e069R2614) structure I created for my Samsung & Google motion photos PRs? If you do it might also be nice to share that as we could all contribute images to an automated test suite.

internal/photoprism/mediafile.go
```diff
@@ -378,6 +378,28 @@ func (m *MediaFile) ExtractEmbeddedVideo() (string, error) {
 
 			return dstPath, nil
 		}
+	} else if metaData.Error == nil && metaData.EmbeddedVideoOffset > 0 {
```
gokcedilek: The if case above, `metaData.EmbeddedVideo != ""` was something I added for Samsung motion photos. Which means we can utilize it here as well, we don't need a separate else if case with `metaData.EmbeddedVideoOffset`. If you look at my most recent [changes](https://github.com/photoprism/photoprism/pull/3709/files#diff-e5a39f92547299cd358543ff78e747595e351aa0f56a50d43c0034cb94465d8aL334), I changed this function a bit to accommodate both Samsung and Google photos as they can be extracted in the same way, we just need to decide if we will 1) use the directory item length approach or 2) use the approach for finding the MP4 header and then we don't need to have this duplicate else if case.

internal/photoprism/mediafile.go
```diff
@@ -378,6 +378,28 @@ func (m *MediaFile) ExtractEmbeddedVideo() (string, error) {
...
+		dstPath := filepath.Join(Config().SidecarPath(), m.RootRelPath(), m.BasePrefix(false)+".MP.mp4")
```
gokcedilek: Instead of hardcoding the `dstPath` extension, see the approach we followed here for Samsung motion photos: https://github.com/photoprism/photoprism/pull/3588/files#diff-e5a39f92547299cd358543ff78e747595e351aa0f56a50d43c0034cb94465d8aR369

internal/meta/data.go
```diff
@@ -14,61 +14,62 @@ const (
...
+	Lat                 float32       `meta:"-"`
+	Lng                 float32       `meta:"-"`
+	Altitude            float64       `meta:"GlobalAltitude,GPSAltitude"`
+	Width               int           `meta:"ImageWidth,PixelXDimension,ExifImageWidth,SourceImageWidth"`
+	Height              int           `meta:"ImageHeight,ImageLength,PixelYDimension,ExifImageHeight,SourceImageHeight"`
+	Orientation         int           `meta:"-"`
+	Rotation            int           `meta:"Rotation"`
+	Views               int           `meta:"-"`
+	Albums              []string      `meta:"-"`
+	EmbeddedVideo       string        `meta:"-"`
```
gokcedilek: Not directly related to this PR, but just to share, in my PR I also changed this field to boolean: https://github.com/photoprism/photoprism/pull/3709/files#diff-5e54ffefc2cd1c77d57258d92f2c45a9d9cebf1ad8b70519309cd32eae287de1R370 thinking that we don't actually need the field name, we just need to know if the video exists, as it is used here: https://github.com/photoprism/photoprism/pull/3709/files#diff-b92b8f546ec67d3a9b9e70bb31cd50416ed9489173fa35e7687aa123fd13e7c9R60, this is the case for both the MP4 header approach and the `DirectoryItemLength` approach I guess so we could keep it in mind

kvalev:
> I'm also curious to know about under which scenarios the offset in the exif data is garbage (@kvalev if you could share more information about that, that would be great:).

The most common problem I encountered was that the mp4 offset was larger than the actual filesize. I dont know exactly why it happens, I personally didnt encounter it, but other users have.

rconn01:
ok, I have implemented without an execution to exiftool twice, and updated this PR. I've tested across many google motion photos and it seems to work fine. This impl works fine for the 1 samsung photo in the examples repo. This impl is cleaner than the original one I proposed.

as for
> Also, just out of my own curiosity, for learning, when you mention the hundreds of tests, are you running these manually?

Yes manually, I ran the server and upload a couple hundred photos and let them index. I then sport checked they were all playable. Also ran a script with ffmpeg against the side car files to ensure they were all playable mp4.

As for which direction to go. I think its either this one or the searching for magic bytes, the original proposal from me is not worth it

internal/photoprism/mediafile.go
rconn01: This PR is now updated to handle samsung/google the same, let me know if you still have a concern/suggestion for the way its done now

internal/photoprism/mediafile.go
rconn01: ok,  I can look to add the mime type check if we pick to go this way. Out of curiosity do we have any examples where its not actually a mp4? how critical is the check/logic and file move

lastzero:
@rconn01 Hello Raymond :) Thanks for all your feedback and ideas! Unfortunately @gokcedilek is attending a software conference right now and we should have our upcoming release ready for QA testing today. Since the only thing missing is Samsung/Google Motion Photo support, it seems best for me to finish it today based on your two pull requests and the information you've gathered. Would that be okay with you?

rconn01:
Sure, I don't mind its a community effort. Curious what else you'd like done. Happy to be a tester here as I have a very large library of Google motion photos

Also not sure if you've had a chance to look at my remaining questions. Don't think they're necessary as part of this feature set, but would be good to understand you're stance for potential future contributions

lastzero:
@rconn01 I'll be happy to answer questions after I've had time to recover from the upcoming release :)

Also, feel free to reach out on Matrix.org! You'll find me in our [Community Chat](https://link.photoprism.app/chat).

rconn01:
Sure no problem, just curious what there is to finish, this should be fully working, can help make updates as needed let me know

internal/ffmpeg/preview.go
```diff
@@ -5,7 +5,7 @@ import "time"
 // PreviewTimeOffset returns an appropriate time offset depending on the duration for extracting a preview image.
 func PreviewTimeOffset(d time.Duration) string {
 	// Default.
-	result := "00:00:00.001"
+	result := "00:00:00.000"
```
lastzero, line 8: This change is interesting! Did it cause any issues when viewing live/motion photos in the UI?

If I remember correctly, there were very few files that had a black still image if you specified exactly `00:00:00.000` here. But since that was a long time ago, I can't provide you with any details...

rconn01: I didn't notice any issues in the UI. Without this change though a number of photos failed to index because ffmpeg resulted in 
```
Output file is empty, nothing was encoded (check -ss / -t / -frames parameters if used)
```

This caused the check for the side car file to fail and the index to not complete. Too a while to find, as the error returned from index was not logged (which is why added the log). I couldn't find a reason why some files hit that issue and others didn't.

This is what ultimately led to my question though. Do we even need to generate a preview image on these mp4s. The original image is effectively the preview image. This would be a larger change though so left it for a future change set

lastzero: Would you be able to share some affected files for testing? I've never this happening before.

rconn01: almost all my photos have faces that I am not comfortable sharing, however I was able to find one to share. I have provided the raw image and the extracted mp4 that gives the issue https://www.dropbox.com/scl/fo/h954riwxmguqntzegq6aj/h?rlkey=jx69sw2i77r8rlzezp68gjb80&dl=0

rconn01: fwiw, I just downloaded the video portion straight from google photos, with the "Download Video" option. It also has the same issue as the one extracted from the jpg. Meaning I don't think its related to how we are doing extraction. 


And to share more detail why this is problematic. FFmpeg wouldn't produce an output file, which cause this to error https://github.com/photoprism/photoprism/blob/8b6de380d0ef4486972a5168f496e75fe81ec351/internal/photoprism/convert_image.go#L101 because the file doesn't exist on disk

lastzero: As a general observation, Google and Samsung seem to add the start frame and metadata at the END of their embedded video files - contrary to their own explicit recommendation on how you should do it. This prevents HTTP streaming with chunked encoding and makes extracting a still image overly complicated as you have to read the entire file, which can take a long time if you are using remote network storage.

graciousgrey:
Thank you so much for your contribution! Since we recently released this feature, I'm closing this PR.

PR #3735, opened 2023-09-17
Docker: Add Lunar-specific Darktable installation

qlyoung:
When on Lunar, add the Lunar-specific repo for Darktable and install it from there, instead of falling back to the system package version (which is not the latest release).

Like the current installation code, the code for 23.04 was pulled straight from [here](https://software.opensuse.org/download.html?project=graphics:darktable&package=darktable).

---

Please note: I don't think there are tests for this, but if I'm mistaken, please point me to them and I'm happy to add them.

Acceptance Criteria:

### DIFF
--- a/scripts/dist/install-darktable.sh
+++ b/scripts/dist/install-darktable.sh
@@ -32,6 +32,11 @@ case $DESTARCH in
       curl -fsSL https://download.opensuse.org/repositories/graphics:darktable/xUbuntu_22.04/Release.key | gpg --dearmor | tee /etc/apt/trusted.gpg.d/graphics_darktable.gpg > /dev/null
       apt-get update
       apt-get -qq install darktable
+    elif [[ $VERSION_CODENAME == "lunar" ]]; then
+      echo 'deb http://download.opensuse.org/repositories/graphics:/darktable/xUbuntu_23.04/ /' | sudo tee /etc/apt/sources.list.d/graphics:darktable.list
+      curl -fsSL https://download.opensuse.org/repositories/graphics:darktable/xUbuntu_23.04/Release.key | gpg --dearmor | sudo tee /etc/apt/trusted.gpg.d/graphics_darktable.gpg > /dev/null
+      sudo apt update
+      sudo apt install darktable
     elif [[ $VERSION_CODENAME == "bullseye" ]]; then
       apt-get update
       apt-get -qq install -t bullseye-backports darktable


lastzero:
Have you tested it on ARM as well? Is it getting security updates? Note that we are upgrading to Ubuntu 23.10 next month. So it may not be worth updating individual packages before we upgrade to a newer base image anyway.

qlyoung:
Just to add some context to this change, I want to point out that I did exactly what you did for Jammy in this commit:
https://github.com/photoprism/photoprism/commit/d2c6eac3bf

> Have you tested it on ARM as well? 

No. Based on the script contents and history, I don't think that is needed? The repo addition and installation is gated under the AMD64/x86 case of the script. [In the ARM case](https://github.com/photoprism/photoprism/blob/9fa7563f0cc35c81063639b6b0132e4409c9b2c4/scripts/dist/install-darktable.sh#L46) darktable is installed from the system repository.

This change certainly isn't breaking anything for ARM, if that's what you're concerned about.

> Is it getting security updates? 

I assume the answer is the same as the code you added for 22.04 in the change I linked.

> Note that we are upgrading to Ubuntu 23.10 next month. So it may not be worth updating individual packages before we upgrade to a newer base image anyway.

The special case exists for Jammy; it should exist for Lunar; when 23.10 is out, the script will fallback to the system repo package. When a newer release is available on OBS for 23.10 that should be added to the script as well.

qlyoung:
To add a little context for why it's important to have the most recent version of darktable, it's because darktable will not correctly handle `.xmp` files containing edit stacks that were created with a later version of darktable. When photoprism uses an older version of `darktable-cli` to produce sidecar conversions of raws, it results in generated sidecars that are improperly graded.

For example, applying a `.xmp` created with darktable 4.4.2 using `darktable-cli` from 4.2.1 results in this:
out-421_s

Applying the same `.xmp` with `darktable-cli` from 4.4.2 produces the correct output:
out-442_s

Another example, 4.2.1 and then 4.4.2:
out-dt421_s
out-dt442_s

lastzero:
Thanks for your feedback! I understand that a newer version is better for some use cases like yours. On the other hand, we need to provide a consistent and secure product - not one where the images generated depend on the processor architecture or where we are responsible for checking the installed binaries for security issues instead of leaving that to the base distribution. We are a small team and have very few resources.

From what I remember, the solution for Jammy was a temporary workaround we needed to get some formats to work at all. We didn't want it to stay that way, so now we use the distribution version instead.

For now, you can install the latest Darktable version in your own Docker image. If I have some time next month and upgrade to a new base image anyway, I'll be happy to take a look and see what we can do to make newer Darktable versions available to our users.

qlyoung:
> On the other hand, we need to provide a consistent and secure product - not one where the images generated depend on the processor architecture or where we are responsible for checking the installed binaries for security issues instead of leaving that to the base distribution.

I agree with these goals. Given the presence of an identical solution for Jammy ("newer darktable needed to support some formats"), I assumed the same solution for a similar problem ("newer darktable needed to support newer xmp formats") for Lunar would be acceptable. If you are now saying only packages distributed via the distro repository are viable, that's fine.

> For now, you can install the latest Darktable version in your own Docker image. 

Okay, I will do that.

> If I have some time next month and upgrade to a new base image anyway, I'll be happy to take a look and see what we can do to make newer Darktable versions available to our users.

Thanks!

lastzero:
If we can get external packages that are regularly updated and available for both AMD64 and ARM64, that would be a viable solution. Not making the same version available to ARM64 users, i.e. anyone who has a Mac or a Raspberry, is not great, as you will probably agree.

qlyoung:
My belief is that Darktable uses OBS to build every release. OBS builds for AMD64, ARM64 and PPC64LE, so every platform you listed should be covered.

http://download.opensuse.org/repositories/graphics:/darktable/xUbuntu_23.04/

What other information would you need to see to be satisfied that the packages are available and updated? I can assist with gathering it.

If this is sufficient, would you agree that updating this patch to add the same code to install from the above repository to the ARM section of the script is good enough? Or would more be required? The downside of that is that it would require adding a new case to add the appropriate repo URL for the ubuntu distribution version every time you change base images.

lastzero:
One question would be whether these packages support everything that the distribution packages support, e.g. GPU acceleration for rendering RAW images. It would also be good to know how these packages affect the size of the Docker image, e.g. if they include custom shared or static library versions or if they use the installed distribution libraries.

qlyoung:
> One question would be whether these packages support everything that the distribution packages support, e.g. GPU acceleration for rendering RAW images.

I don't see any indicator that says otherwise, and I don't know why they would disable hwaccel in their OBS packages versus system packages. If it's that critical, I can go ask them?

> It would also be good to know how these packages affect the size of the Docker image, e.g. if they include custom shared or static library versions or if they use the installed distribution libraries.

The AMD64 `.deb` is 6 mb, which seems negligible regardless of their linking strategy...I'm sure the uncompressed version is a similar size

lastzero:
> I don't see any indicator that says otherwise, and I don't know why they would disable hwaccel in their OBS packages versus system packages. If it's that critical, I can go ask them?

Support for some file formats, especially patented formats like HEIC/HVEC, may depend on the compiler flags used. I'm not saying that's the case here, but I know it is with FFmpeg, which is why they only distribute the source code.

In some cases it is also very important for security that the distribution libraries are used, for example with the recently disclosed vulnerability in WebP: https://floss.social/@photoprism/111064346355243107

> The AMD64 `.deb` is 6 mb, which seems negligible regardless of their linking strategy...I'm sure the uncompressed version is a similar size

From my understanding, the deb package could - theoretically - also install additional software and libraries from opensuse.org. So just because the package is only 6 MB in size it doesn't necessarily mean that it only installs 6 MB.

lastzero:
@qlyoung Could you please ask the package maintainers specifically about the following:

- [ ] If these packages do not support certain hardware (like GPUs that depend on the distribution's driver support) or file types (such as Apple ProRAW DNG or HEIC) that might otherwise be supported by the distribution packages, e.g. due to patented codecs?

- [ ] Whether they download and/or link to other libraries, so that updating those libraries in the base distribution - e.g., after a [security vulnerability becomes known](https://floss.social/@photoprism/111064346355243107) - would not fix the issue in Darktable?

- [ ] What the minimum processor instruction set is for ARM64 and AMD64, e.g. whether AVX or AVX2 support is required in Intel/AMD CPUs?

- [ ] How often the packages are updated, especially in case major security vulnerabilities are found in Darktable (or other related packages additionally installed from opensuse.org)?

**I will also add the same questions to issue #3741 so you can answer them there in preparation for the upgrade. Thank you very much! :pray:**

qlyoung:
Added it to my FOSS backlog, will hopefully get to it soon

lastzero:
Awesome! We can assign this issue to you if you comment on it, so others can see that you are working on it:

- #3741

PR #3737, opened 2023-09-18
Fedora: Improve Docker build environment compatibility

strugee:
The problem is that in SELinux enforcing distributions, access to `docker.sock` is restricted in containers even if DAC permissions supposedly allow access. This causes endless log lines when invoking `docker-compose up` similar to the following:

    time="2023-09-18T04:10:56Z" level=error msg="Failed to retrieve information of the docker client and server host: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get \"http://%2Fvar%2Frun%2Fdocker.sock/v1.24/version\": dial unix /var/run/docker.sock: connect: permission denied" providerName=docker

The solution is to run the Traefik container in privileged mode. See https://stackoverflow.com/a/30368817/1198896.

Acceptance Criteria:

(I'm just checking boxes here for CI purposes even though almost all of these are N/A since this is a dev environment change.)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -153,6 +153,7 @@ services:
   ## Docs: https://doc.traefik.io/traefik/
   traefik:
     image: photoprism/traefik:latest
+    privileged: true
     ports:
       - "80:80" # HTTP (redirects to HTTPS)
       - "443:443" # HTTPS (required)
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,6 +15,7 @@ services:
     security_opt:
       - seccomp:unconfined
       - apparmor:unconfined
+    privileged: true
     ports:
       - "2342:2342"   # Default HTTP port (host:container)
       - "2443:2443"   # Default TLS port (host:container)


strugee:
I've just pushed an additional patch since I encountered more SELinux problems. We now run the main container in privileged mode too, to work around similar problems.

lastzero:
Are you actually running Docker on Fedora or Podman, which is more common on Red Hat-based distributions? I know Podman often requires more permissions, especially if you use ports < 1024.

strugee:
I started running rootless Podman. However I ran into lots of issues with that (port 80 being privileged was one, but the biggest was that there _is_ no `docker.sock` for Traefik) so I decided to see if I could get a dev environment setup working with Docker and then circle back to making Podman work.

The bug fixed here happened with Docker running fully outside a Toolbox/Podman container.

sgflt:
https://www.redhat.com/sysadmin/privileged-flag-container-engines
https://blog.cthudson.com/2023-11-02-running-traefik-with-podman/

lastzero:
@strugee Thanks for suggesting this change! After seeing (more than one) heated argument against it from Podman (i.e. Fedora or Redhat) users, I'd rather not merge this as the debates distract us from other work:

- https://github.com/photoprism/photoprism/discussions/4846

I will therefore go ahead and close this PR :/

PR #3828, opened 2023-10-17
Adds Galaxy S10 to cameraModels

lukasdotcom:
My brother has the S10 and it shows up as Samsung SM-G973U2 right now.


lastzero:
Thank you!

Issue #3763, opened 2023-09-24
Albums: Add option to sort albums/states chronologically by date of assigned photos


graciousgrey:
The "newest first" filter sorts by the year/month/day property of an album. For manually managed albums and state albums, no album year/month/day is defined. In this case "newest first" is the same as "recently added".

**Acceptance Criteria:**
- When sorting albums or states by "newest first", the albums with the most recently taken photo must be displayed first.
- When sorting albums or states by "Oldest first", the albums containing the photo with the oldest date taken must be displayed first.

Xirt:
I checked my database and noticed that the field "album_year", "album_month" and "album_day" are empty for all my albums, resulting in aforementioned issue. Wouldn't it make more sense to populate these fields based on the photos in present in the album (i.e. update on additions / removals)? Afterwards, the filters should automatically start working.

An alternative would be to expose these fields in the GUI so they can be populated/updated by a user.

_Note that the first "fix" would require a decision on whether the newest or oldest photo information should be used for populating the fields._

graciousgrey:
At the moment only moment and month albums do have a defined year, month and date. 

It's the  scope of this issue to add additional logic for manually managed albums. 

We need to determine and save both the taken date of the oldest photo in the album and the taken date of the newest photo. This information must also be updated if photos are added or removed from the album. It might also make sense to exclude dates of private photos. This should be done automatically so that the user does not have to define dates for albums. So it is a pure backend task.

raudraido:
+1

zbejas:
I would also like to add that some images that have IDs for filenames, for example, `00129-something` will get a date added of 9th of December when scanned. 
This messes up any sorting, and manually changing photo filenames would be a pain.
I propose a sorting method that is tied to the Creation/Update date of the file.

rosinscj:
+1

ALEEF02:
+1 bump!
Would like to use this feature in tandem with AI ranking I have on an external API to force a sort of the photos by my own criterion.

PR #3832, opened 2023-10-17
Albums: Add option to sort albums/states chronologically by date of assigned photos

gokcedilek:
https://github.com/photoprism/photoprism/issues/3763

- [x] When a photo is added, update the album oldest / newest fields if needed
- [x] When a photo is removed, re-compute the album oldest / newest if needed and update the fields
- [x] When a user updates the date of a photo that is the newest or oldest in an album, check if it is still the newest or oldest, re-compute the album oldest / newest if needed and update the fields
- [ ] Make sure the code and the queries are optimized

### DIFF
--- a/internal/entity/album.go
+++ b/internal/entity/album.go
@@ -901,13 +901,13 @@ func (m *Album) RemovePhotos(UIDs []string) (removed PhotoAlbums) {
 			photo.PhotoUID != "" {
 			takenAt := photo.TakenAt
 			if isOldest := takenAt.Equal(m.AlbumOldest); isOldest {
-				if oldestPhoto, err := AlbumOldestOrNewest(m.AlbumUID, true); err == nil {
+				if oldestPhoto, err := AlbumOldestOrNewest(m.AlbumUID, true, ""); err == nil {
 					// update the oldest of the album
 					updatedAlbumOldest = oldestPhoto.TakenAt
 				}
 			}
 			if isNewest := takenAt.Equal(m.AlbumNewest); isNewest {
-				if newestPhoto, err := AlbumOldestOrNewest(m.AlbumUID, false); err == nil {
+				if newestPhoto, err := AlbumOldestOrNewest(m.AlbumUID, false, ""); err == nil {
 					// update the newest of the album
 					updatedAlbumNewest = newestPhoto.TakenAt
 				}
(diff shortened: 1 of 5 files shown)


lastzero:
My apologies for the late feedback on your changes! I hope you had a good start into the new year :)

Since empty albums should not have an oldest or newest date, it might be a good idea to change the entity time values to pointers (so that they can be set to nil if needed):

```go
AlbumOldest	 *time.Time	 `json:"Oldest" yaml:"Oldest"`
AlbumNewest 	 *time.Time	 `json:"Newest" yaml:"Newest"`
```

Based on the currently implemented functionality, it would seem best to (also) update the new album date fields in the following place (or move this function to the entity package)?

https://github.com/photoprism/photoprism/blob/e03dbe5d16ac47c27552369bca2b12bac1ea279d/internal/query/albums.go#L107-L124

This function is automatically executed in regular intervals [by the meta worker](https://github.com/photoprism/photoprism/blob/develop/internal/workers/meta.go) so that the updates happen automatically in the background:

https://github.com/photoprism/photoprism/blob/e03dbe5d16ac47c27552369bca2b12bac1ea279d/internal/photoprism/moments.go#L264

That's particularly useful for albums that are based on search filters, e.g. the "state", "folder" and "moment" types (for manually managed albums, an additional update statement can be run in the entity model so that these albums are updated immediately after performing changes).

The SQL query needed to determine and update the dates in `UpdateAlbumDates()` would be different for each album type (e.g. manual or state). To cover all types, a similar approach to the cover query below could theoretically be taken:

https://github.com/photoprism/photoprism/blob/e03dbe5d16ac47c27552369bca2b12bac1ea279d/internal/query/albums.go#L32-L33

However, from what I know, we only need to take care of the "album", "state" and "folder" types, and for those it should be possible to determine the oldest and newest date without running a regular search query for each individual album, which might otherwise cause quite a bit of overhead:

https://github.com/photoprism/photoprism/blob/e03dbe5d16ac47c27552369bca2b12bac1ea279d/internal/query/albums.go#L61

Happy to answer any follow-up questions you may have 👍

lastzero:
Thanks for starting to work on this, @gokcedilek! We really appreciate your contributions.

Unfortunately, since you can't move forward, I'll have to close this PR so we can focus on the other issues and pull requests. 😞

Issue #4030, opened 2024-01-24
Videos: Improve Intel QSV Hardware Support and Performance


xrad:
#### 1. What is not working as documented?

[DISCLAIMER: I know dev time is scarce, that this is a very specific case and that it may not even be specific to photoprism. But it happens and is reproducable in PP's docker compose container and I could not find anything about this issue anywhere else.]

Qsv accelerated ffmpeg does not work and falls back to software coding. In the docker logs this manifests as:
```
3.100\n  libswscale      7.  1.100 /  7.  1.100\n  libswresample   4. 10.100 /  4. 10.100\n  libpostproc    57.  1.100 / 57.  1.100\nlibva info: VA-API version 1.19.0\nlibva info: Trying to open /usr/lib/x86_64-linux-gnu/dri/iHD_drv_video.so\nlibva info: Found init function __vaDriverInit_1_18\nlibva info: va_openDriver() returns 0\n[AVHWDeviceContext @ 0x562abf67fe80] Error creating a MFX session: -9.\nDevice creation failed: -1313558101.\nFailed to set value '/dev/dri/renderD128' for option 'qsv_device': Unknown error occurred\nError parsing global options: Unknown error occurred\n"
root-photoprism-1  | time="2024-01-23T13:31:48Z" level=warning msg="h264_qsv: failed to transcode 
```

#### 2. How can we reproduce it?

In the container specified below, you can reproduce the issue on the command line. To further simplify, I only focus on decoding:

```
root@231128:/photoprism/storage$ ffmpeg -hwaccel qsv -c:v h264_qsv -i input.mp4 -c:v libx264 -f mp4 output.mp4

...
libva info: VA-API version 1.19.0
libva info: Trying to open /usr/lib/x86_64-linux-gnu/dri/iHD_drv_video.so
libva info: Found init function __vaDriverInit_1_18
libva info: va_openDriver() returns 0
[AVHWDeviceContext @ 0x555f7c623cc0] Error creating a MFX session: -9.
Device creation failed: -1313558101.
No device available for decoder: device type qsv needed for codec h264_qsv.
Stream mapping:
  Stream #0:1 -> #0:0 (h264 (h264_qsv) -> h264 (libx264))
  Stream #0:0 -> #0:1 (aac (native) -> aac (native))
Device setup failed for decoder on input stream #0:1 : Unknown error occurred
```

Could "VA-API version 1.19.0" vs "__vaDriverInit_1_18" be a culprit?

[...]

PR #4052, opened 2024-02-04
Scripts: add libmfx1 to installed packages for Intel QSV hardware sup…

xrad:
Add libmfx1 to apt-get as discussed in #4030


lastzero:
@xrad Based on this, the libmfx1 package should be installed automatically by libmfx-dev (which has libmfx1 as its only dependency):
- https://packages.ubuntu.com/noble/libmfx-dev

Can we close this then? Note that we prefer the `-dev` packages whenever possible, as their name does not change when there is a new version or a different distribution/base image is used, which often saves a lot of maintenance time.

xrad:
Sure, closing.

PR #4155, opened 2024-04-06
Draft: Add ability to search for similar images

sazarkin:
This is POC for similarity search inspired by this PR https://github.com/photoprism/photoprism/pull/2005

What it does:

- Uses TensorFlow CLIP (ViT-B-16) model to create embeddings ([512]float32) vector and stores it db
- Uses FAISS https://github.com/facebookresearch/faiss library to maintain in-memory index for this embeddings. Because MariaDB does not support yet index like this.
- Adds UI screen to show similar images 

image

Saved CLIP model: *edited*

What it can do:
- Use this embedding for text search as proposed in https://github.com/photoprism/photoprism/pull/2005
- Upload any image like google images to search

UI is not in very good state, index init is just inserted at the start command, libfaiss is not added into docker image.
This is not in ready state, and I want to receive comments if this approach even desirable in upstream.


graciousgrey:
Thank you very much for that! We will review it and give you feedback as soon as we have finalized our next stable release :)

lastzero:
@sazarkin We like the approach, although it would be good not to depend on yet another C library. I assume there are a number of mature Go vector search/database libraries that we could use instead.

That would also help with face recognition, as the similarity search there is implemented with a library I wrote myself. So there is probably a lot that could be improved and optimized.

Would that be something you can help with?

PR #4199, opened 2024-04-19
MariaDB: Change image name in Docker Compose config example for ARMv7

GetTuh:
### - What does it implement / fix / improve? Why?
Armv7 support , since [Linuxserver's image is not supporting armhf anymore](https://hub.docker.com/r/linuxserver/mariadb)
This mariadb image is recently updated and running on linux alpine. 
### - Are the changes related to an existing issue?
no

Acceptance Criteria:

- [ x] Features and enhancements must be fully implemented so that they can be released at any time without additional work
- [ x] Automated unit and/or acceptance tests are mandatory to ensure the changes work as expected and to reduce repetitive manual work
- [ x] Frontend components must be responsive to work and look properly on phones, tablets, and desktop computers; you must have tested them on all major browsers and different devices
- [ x] Documentation and translation updates should be provided if needed
- [ x] In case you submit database-related changes, they must be tested and compatible with SQLite 3 and MariaDB 10.5.12+

### DIFF
--- a/setup/docker/armv7/docker-compose.yml
+++ b/setup/docker/armv7/docker-compose.yml
@@ -106,7 +106,7 @@ services:
   ## Database Server (recommended)
   ## see https://docs.photoprism.app/getting-started/faq/#should-i-use-sqlite-mariadb-or-mysql
   mariadb:
-    image: linuxserver/mariadb:latest
+    image: yobasystems/alpine-mariadb:latest
     restart: unless-stopped
     stop_grace_period: 5s
     security_opt:


lastzero:
Thank you very much! Did you check if the names of the environment variables are the same? We currently use the following for the official MariaDB image:
```yaml
environment:
      MARIADB_AUTO_UPGRADE: "1"
      MARIADB_INITDB_SKIP_TZINFO: "1"
      MARIADB_DATABASE: "photoprism"
      MARIADB_USER: "photoprism"
      MARIADB_PASSWORD: "insecure"
      MARIADB_ROOT_PASSWORD: "insecure"
```

GetTuh:
The names are the same as before for linuxsystem's but [different than the one you posted:](https://github.com/yobasystems/alpine-mariadb?tab=readme-ov-file#docker-compose-example)

```mysql:
  image: yobasystems/alpine-mariadb
  environment:
    MYSQL_ROOT_PASSWORD: hguyFtgfR4r9R4r76
    MYSQL_DATABASE: wordpressdb
    MYSQL_USER: wordpressuser
    MYSQL_PASSWORD: hguyFt6S95dgfR4ryb
```
 I've had the server running since yesterday without problems.

lastzero:
I see... it's a shame that they don't support the same variables as the original image, especially `MARIADB_AUTO_UPGRADE`, as that could avoid upgrade issues and confusion among inexperienced users:
- https://docs.photoprism.app/getting-started/troubleshooting/mariadb/#auto-upgrade

lastzero:
@GetTuh Note that I changed the mount point of the data volume in the config example after checking the image documentation (the linuxserver image used `/config` instead of the default `/var/lib/mysql`):
- https://dl.photoprism.app/docker/armv7/docker-compose.yml

**If you are using `/config` with the new image, we strongly recommend that you create a backup dump (as described in our documentation) and then reinitialize the MariaDB service with the new settings.**

PR #4203, opened 2024-04-21
WIP: Replace GORM with GORMv2

albrechtf:
I mark this currently as a Draft, although the Go tests look quite good on my end. I will have to do a lot more tests, especially with MariaDB and using the frontend.

Notable changes with GORM v2:

* I had to change type Map, which was (contrary to its description) not a type **alias**, to a type alias now, because Gorm otherwise complains about "unsupported data" in all kinds of operations (that took me only five hours to find).
* Gorm now no longer supports the `association_` tags. I am really not sure by what to replace some of them. As long as it is just about column naming, replacements are `joinForeignKey` etc., and that was easy to replace.
* Same for Preload tag. This now has to be done via code. I removed them when I encountered them, but some may still be there (without effect). As I did not add equivalent code as a replacement, this could decrease performance.
* `DropTableIfExists` is no longer supported by the Migrator. I replaced it with `DropTable` and just a log of any errors.
* Gorm now e.g. supports soft-deletion out of the box. I did not use that, as this is already implemented manually for some object types. This is something for a later PR (by another person, probably).
* `sqlite3` is now `sqlite` in Gorm (sounds like a good idea to me). I only changed the strings (and directory name), but not the constant names.

Other notes:

* I am really not familiar with Go, so I am not sure about the changes in the go.mod file. That is just what `go mod tidy` did for me :smile: 

Acceptance Criteria:

- [? yeah think so] Automated unit and/or acceptance tests are mandatory to ensure the changes work as expected and to reduce repetitive manual work


lastzero:
@albrechtf Thanks a lot for working on this! I really appreciate it and look forward to taking a look at your changes when I have some time, either tomorrow or when I'm back. I'll only be in the office until Wednesday to get some rest after the release.

The [GORM v1 magic behind Associations](https://v1.gorm.io/docs/associations.html) had stopped me out and made me revert all my changes for v2 when I tried to do this myself because I didn't have the time to find a good solution that would work for all existing use cases. I didn't really like the Associations to begin with, but ended up using them anyway so as not to work against the framework and have working software that we can release.

We wouldn't be sad at all if the existing Associations could be replaced with something else, as long as we don't have to write lots of manual queries (via copy & paste) instead and the new solution is reliable enough to be released... 

Similarly, [Preload is framework magic](https://v1.gorm.io/docs/preload.html) that we would like to refactor/replace if this is possible without risking hard to find bugs that we then have to identify and fix before we can release the next stable version.

The other changes you mentioned seem straightforward and relatively low risk! 👍

lastzero:
If you are interested in more detailed feedback, I would be happy to take a look at your code/changes early next week? Anything other than the fixtures you want me to check in particular? Note that, unfortunately, I won't be able to provide you with a ready to use solution/replacement for the Associations and Preload functionality available in Gorm v1.

lastzero:
You also mentioned an album fixture that has a duplicate UID? If you let me know which one it is, I can quickly resolve that for you.

albrechtf:
Hi @lastzero , sorry for the long time without feedback. Hard to find time to work on this.
I have made good progress in adjusting most of the SQL to support both MySQL and PostgreSQL. There are some statements left, mostly around file indices with some... interesting calculations. I will contact you in chat about this.
Also, I fixed many of the gorm2 relations. Now, the automagically generated structure in PostgreSQL looks quite good.

Some tests are still failing, but looks better wrt duplicates.

lastzero:
@albrechtf I would like to apologize for the recent reorganization of our internal Go packages and hope it does not cause you too much pain due to potential merge conflicts?

It was mostly just sorting the existing packages into subdirectories based on the functionality they provide. So now you can find all the code related to entity models as well as database queries and migrations in the `entity` package directory, or a subdirectory of it, e.g. `migrate`, `search` and `query` (note that the name of the packages hasn't changed):

internal-entity

This was necessary to reduce the amount of scrolling, which already required a large screen before and finally became excessive with the new packages needed for OIDC support. Also, an intern will be joining us this month to work on AI related features. I think it will be much easier for him to find the code/packages this way.

keif888:
@lastzero I found 2 duplicate fixture records when checking the SQL that was run in my testing.
album_fixtures.go has a duplicate AlbumUID as6sg6bipotaab19 on line 274
file_fixtures.go has a duplicate FileUID fs6sg6bw45bn0007 on line 373


**I assume the intent of this pull is an end result where there are no functional changes after GORM2 is implemented.**

I have created a new branch from develop and reapplied the changes from this pull to it manually.  There were to many conflicts when I attempted to merge the origin\develop in, so I chose to manually apply albrechtf's changes.  Hopefully I got them all.
https://github.com/keif888/photoprism/tree/gorm2

**I have changed all structs that had VARBINARY to use bytes to address an index size issue in MariaDB.  This works with sqlite and should work with Postgres. **

A number of the standard tests are currently failing on sqlite, so I decided to start from the "just loaded for testing state" and test from there.
https://github.com/keif888/photoprism/tree/gorm2/internal/entity/dbtest

I have forced the test load fixtures to ignore associations where required via Omit(clause.Associations) to prevent GORM2 walking the tree and inserting all the child records which is different to GORM1.

I have enabled SoftDelete on Files as that was required to fix some of the tests that are breaking.

I haven't added a PULL here as this one is already here, and I don't believe mine is ready to include, as the unit tests suite is failing.  and I'm not sure where @albrechtf is up to?

keif888:
@lastzero I have got my branch passing the existing unit tests, and on my development build, it is able to upload, and then retrieve photos, albums, and approve.  I haven't done extensive UI testing.
SoftDelete is fully implemented.

Current issues:
- I haven't changed any pre loads that didn't break unit tests
- I haven't checked if there are any pre loads that have been removed and not implemented as code
- There are more records in the test database than with gormv1
- Gorm V1 will execute BeforeCreate and BeforeSave on a Save that becomes a Create.  Gorm V2 does not do this.  **IT WILL NOT TRIGGER A BeforeCreate IF A SAVE BECOMES A CREATE.**  This is likely to have undetected consequences as most creates are done by calling the Save from what I have seen in the code.

lastzero:
@keif888 Thank you very much for your work and feedback on this! :partying_face: 

I have updated our fixtures so that the UIDs no longer conflict.

> I have changed all structs that had VARBINARY to use bytes to address an index size issue in MariaDB. This works with sqlite and should work with Postgres. 

What (actual) data type will Gorm v2 choose for MariaDB in this case?

There were no index size issues with Gorm v1, right? If so, the VARBINARY type might have been magically changed to VARCHAR by the framework, which would explain the larger index size (with 4 bytes per Unicode character instead of just 1 as required for storing ASCII).

> Gorm V1 will execute BeforeCreate and BeforeSave on a Save that becomes a Create. Gorm V2 does not do this. IT WILL NOT TRIGGER A BeforeCreate IF A SAVE BECOMES A CREATE. This is likely to have undetected consequences as most creates are done by calling the Save from what I have seen in the code.

Good catch! The `BeforeCreate()` function should mainly be used to generate a UID (and sometimes also a shorter `RefID` for logging purposes) before inserting new records, for example:

```go
// BeforeCreate creates a random UID if needed before inserting a new row to the database.
func (m *Label) BeforeCreate(scope *gorm.Scope) error {
	if rnd.IsUnique(m.LabelUID, LabelUID) {
		return nil
	}

	return scope.SetColumn("LabelUID", rnd.GenerateUID(LabelUID))
}
```

To work around this (potential) issue, you could add `BeforeSave()` functions that call `BeforeCreate()` so that they both do the same thing - unless this can lead to weird situations where, for example, one of the `Save()` implementations incorrectly assumes that a record already exists because a UID was generated by the callback before the check?

> There are more records in the test database than with gormv1

Have you been able to identify which entities or tables are affected? These additional records might have been added due to changes in the way the Associations or Create/Save functions work?

**In order for me to help with the remaining issues, it would be good if you could open a PR, as this will make it easier to compare and comment on specific changes.**

keif888:
> > I have changed all structs that had VARBINARY to use bytes to address an index size issue in MariaDB. This works with sqlite and should work with Postgres.
> 
> What (actual) data type will Gorm v2 choose for MariaDB in this case?

Gorm v2 is choosing varbinary when utilising the Gorm type bytes.  I double checked that as this way there will be no database migration issues as it's the same data type as per Gorm v1, and it's database independent (so should work in Postgres etc.).  

> 
> There were no index size issues with Gorm v1, right? If so, the VARBINARY type might have been magically changed to VARCHAR by the framework, which would explain the larger index size (with 4 bytes per Unicode character instead of just 1 as required for storing ASCII).

There were no index size issues in Gorm v1 because it is using VARBINARY.  

> > Gorm V1 will execute BeforeCreate and BeforeSave on a Save that becomes a Create. Gorm V2 does not do this. IT WILL NOT TRIGGER A BeforeCreate IF A SAVE BECOMES A CREATE. This is likely to have undetected consequences as most creates are done by calling the Save from what I have seen in the code.
> 
> Good catch! The `BeforeCreate()` function should mainly be used to generate a UID (and sometimes also a shorter `RefID` for logging purposes) before inserting new records, for example:
> 
> ```go
> // BeforeCreate creates a random UID if needed before inserting a new row to the database.
> func (m *Label) BeforeCreate(scope *gorm.Scope) error {
> 	if rnd.IsUnique(m.LabelUID, LabelUID) {
> 		return nil
> 	}
> 
> 	return scope.SetColumn("LabelUID", rnd.GenerateUID(LabelUID))
> }
> ```
> 
> To work around this (potential) issue, you could add `BeforeSave()` functions that call `BeforeCreate()` so that they both do the same thing - unless this can lead to weird situations where, for example, one of the `Save()` implementations incorrectly assumes that a record already exists because a UID was generated by the callback before the check?
> 

I will work through the BeforeSave() and do my best to catch any issues.

> > There are more records in the test database than with gormv1
> 
> Have you been able to identify which entities or tables are affected? These additional records might have been added due to changes in the way the Associations or Create/Save functions work?

The extra records are due to child records being included in fixtures, but not in the main fixture for the entity.  The Photo fixture is the main (and maybe only) cause as it creates extra Labels and Keywords.  I haven't tracked each individual record down after I proved the first couple were coming from fixture files.  
> 
> **In order for me to help with the remaining issues, it would be good if you could open a PR, as this will make it easier to compare and comment on specific changes.**

Will do a PR tomorrow.

lastzero:
@albrechtf Thank you for your contribution! I'll go ahead and close this PR as it has been superseded:

- https://github.com/photoprism/photoprism/pull/4831

PR #4323, opened 2024-06-07
UX: Update counters and covers in the background

tschechniker:
I have a library of about 150K of photos. Even with an optimized MariaDB every Approve/Archive Batch action takes about 4 seconds. I debugged and found that from these 4 seconds 3.5 seconds are used to regenerate counters. This can easily be done in the background. 

Now if i move a photo to the archive it takes milliseconds again :smile: 

Acceptance Criteria:

- [x] Automated unit and/or acceptance tests are mandatory to ensure the changes work as expected and to reduce repetitive manual work ( already there )

### DIFF
--- a/internal/api/batch.go
+++ b/internal/api/batch.go
@@ -72,10 +72,10 @@ func BatchPhotosArchive(router *gin.RouterGroup) {
 		}
 
 		// Update precalculated photo and file counts.
-		logWarn("index", entity.UpdateCounts())
+		go func() { logWarn("index", entity.UpdateCounts()) }()
 
 		// Update album, subject, and label cover thumbs.
-		logWarn("index", query.UpdateCovers())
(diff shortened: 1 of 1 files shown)


lastzero:
@tschechniker Thank you so much! I apologize for not getting back to you sooner. Unfortunately, I've only just noticed this great suggestion.

However, while it seems like a generally good idea to update counters and covers in the background, it might affect the frontend in unexpected ways - so I'd rather not merge this without having time to properly test it. We'll make sure it doesn't get forgotten (again)! :warning:

lastzero:
@tschechniker Thanks again for your contribution! Based on your proof-of-concept, the above commit updates counters and covers in the background where possible. I will therefore go ahead and close this PR as it is no longer needed.

I'm very sorry we couldn't get this released sooner. :disappointed:

PR #4372, opened 2024-07-04
Config: Improve parsing of boolean environment variables

diamondburned:
This commit changes the existing code to also accept "true", "1" and other
truthy values as specified by [strconv.ParseBool] from environment variables,
rather than assuming that only "true" is truthy. It also warns the user if a
boolean value is invalid.

For more information on [strconv.ParseBool], see:
https://pkg.go.dev/strconv#ParseBool

This change is not related to an existing issue.

---

Acceptance Criteria:

### DIFF
--- a/internal/config/env.go
+++ b/internal/config/env.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/photoprism/photoprism/pkg/list"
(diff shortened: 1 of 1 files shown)


lastzero:
Thanks! You should be able to use the following function instead of adding a new one:
https://github.com/photoprism/photoprism/blob/ad581aff4b8d0444efb975e8030a5313d2cc5115/pkg/txt/bool.go#L7-L16

diamondburned:
> Thanks! You should be able to use the following function instead of adding a new one:
> 
> https://github.com/photoprism/photoprism/blob/ad581aff4b8d0444efb975e8030a5313d2cc5115/pkg/txt/bool.go#L7-L16

Is there a reason why this function exists instead of `strconv.ParseBool`? It also doesn't seem to warn the user on invalid inputs, should that change be made here as well?

lastzero:
The function `config.Env()` is only used in very few cases, e.g. to set the log level before the regular config options are parsed/initialized:
https://github.com/photoprism/photoprism/blob/ad581aff4b8d0444efb975e8030a5313d2cc5115/internal/config/config.go#L108-L116

I don't believe it's a good idea to report an error at this point, because that would lead to a chicken-and-egg problem, e.g. if you use a flag because you want to suppress error or warning messages. Those using the flags should notice if they have an effect, and the flags that are also part of the regular config options (most/all) would trigger an error at a later stage anyway if they have unsupported values.

diamondburned:
Huh, I see. I initially had the impression that `Env()` is used for all boolean flags such as `PHOTOPRISM_DETECT_NSFW`. Shouldn't we unify `Env()` together with `urfave/cli`'s parser then? That would probably be a breaking behavior, and I feel like this is more of a documentation issue than anything...

lastzero:
@diamondburned My sincere apologies for not getting back to you any sooner! I'm happy to report that I've just implemented your suggestion with additional tests and our existing parser function. I will therefore go ahead and close this PR as it is no longer needed. ✅

PR #4382, opened 2024-07-08
FFmpeg: Fix Intel Quick Sync Video (QSV) hardware transcoding

hicasper:
When using ffmpeg 7 and intel_qsv to transcode videos, it will shows
```
[AVHWDeviceContext @ 0x5fd57abc2f80] No VA display found for device /dev/dri/renderD128.
Device creation failed: -22.
Failed to set value '/dev/dri/renderD128' for option 'qsv_device': Invalid argument
Error parsing global options: Invalid argument
```

So I remove `qsv_device` arg, as https://github.com/photoprism/photoprism/issues/1337#issuecomment-1088648066 mentioned, let ffmpeg to decide which device to use, it works well on both ffmpeg 6 and 7

---

Acceptance Criteria:


lastzero:
@hicasper Thanks a lot for submitting this fix! I'm very sorry that I only noticed it now. I'll test and apply your changes once I have a test environment set up to verify that they (still) work with the current release. :clapper:

lastzero:
I've just verified that it works. However, I was only able to transcode videos on my Linux Mint host (and not in our Docker development environment), which still has ffmpeg 6.1.1 installed (the version of ffmpeg 7.0.2 I have does not support hardware transcoding).

The above commit therefore changes the ffmpeg command flags for Intel QSV as suggested, and also adds hardware transcoding tests for Intel and Nvidia.

Note that the FFmpeg docs should also be updated, as the problematic command flag is still mentioned there:

- https://trac.ffmpeg.org/wiki/Hardware/QuickSync#Transcode

lastzero:
Since the changes were applied manually due to conflicts, I'm going to close this PR (instead of merging it) and then set the status to preview. Thanks again! This was very helpful.

PR #4449, opened 2024-08-05
Check if the current kernel is recent enough for libvips

### LINKED ISSUE #4339: Synology: Fatal glibc error "cannot get entropy for arc4random"
#### 1. What is not working as documented?
After updating [240420] to the latest docker version [240531] I tried to index a specific folder which I recently added.
Then after it finished I recognised that I can not see these newly added pictures.
In the log I saw that it was skipped due to the problem: Fatal glibc error: cannot get entropy for arc4random.

I think it's related to an to old kernel version of my Synology NAS (Linux NAS 3.10.108 #64570).

#### 2. How can we reproduce it?

0. Be sure to use a system with a kernel version of "Linux NAS 3.10.108 #64570" (or maybe even older)
1. Install the 240420 version on docker
2. Click on settings and try to index the pictures of a new added folder 
3. See error "Fatal glibc error: cannot get entropy for arc4random" in the logs

#### 3. What behavior do you expect?
No error

#### 4. What could be the cause of your problem?
I think the root cause is a too old kernel version.
This has also been reported on other software: https://github.com/jlesage/docker-freefilesync/issues/8#issuecomment-1696377382, https://github.com/MatthewVance/unbound-docker/issues/147)

#### 5. Can you provide us with example files for testing, error logs, or screenshots?
If needed I can add logs.

#### 6. Which software versions do you use?

(a) PhotoPrism version: 240531
(b) Database Type & Version: ?
(c) Operating System Types & Versions: Linux -> DSM 7.2 from Synology NAS
(d) Browser Types & Versions: Chrome (current version)
(e) Ad Blockers, Browser Plugins, and/or Firewall Software: uBlockOrigin

#### 7. On what kind of device is PhotoPrism installed?
Synology NAS DS716+

#### 8. Do you use a Reverse Proxy, Firewall, VPN, or CDN?
Yes, reverse proxy

#### 9. Docker-compose for installation
[docker-compose.txt](https://github.com/user-attachments/files/15890344/docker-compose.txt)

petertrr:
Closes #4339

Acceptance Criteria:

### DIFF
--- a/scripts/dist/entrypoint.sh
+++ b/scripts/dist/entrypoint.sh
@@ -80,8 +80,8 @@ fi
 # Check if the current kernel is recent enough, see https://github.com/photoprism/photoprism/issues/4339
 MIN_KERNEL='3.17'
 OLDER_KERNEL=$(printf "$MIN_KERNEL\n$(uname -r)" | sort -V | head -1)
-if [ ! "$OLDER_KERNEL" = "$MIN_KERNEL" ]; then
-  PHOTOPRISM_DISABLE_VIPS=true
+if [[ ! "OLDER_KERNEL" = "MIN_KERNEL" ]]; then
+  export PHOTOPRISM_DISABLE_VIPS=true
 fi
 
 # display additional container info for troubleshooting
--- a/scripts/dist/entrypoint.sh
+++ b/scripts/dist/entrypoint.sh
@@ -77,6 +77,13 @@ else
   umask 0002
 fi
 
+# Check if the current kernel is recent enough, see https://github.com/photoprism/photoprism/issues/4339
+MIN_KERNEL='3.17'
+OLDER_KERNEL=$(printf "$MIN_KERNEL\n$(uname -r)" | sort -V | head -1)
+if [ ! "$OLDER_KERNEL" = "$MIN_KERNEL" ]; then
+  PHOTOPRISM_DISABLE_VIPS=true
+fi
+
 # display additional container info for troubleshooting
 echo "file umask....: \"$(umask)\" ($(umask -S))"
 echo "home directory: ${HOME}"
--- a/scripts/dist/entrypoint.sh
+++ b/scripts/dist/entrypoint.sh
@@ -80,7 +80,7 @@ fi
 # Check if the current kernel is recent enough, see https://github.com/photoprism/photoprism/issues/4339
 MIN_KERNEL='3.17'
 OLDER_KERNEL=$(printf "$MIN_KERNEL\n$(uname -r)" | sort -V | head -1)
-if [[ ! "OLDER_KERNEL" = "MIN_KERNEL" ]]; then
+if [[ ! "$OLDER_KERNEL" = "$MIN_KERNEL" ]]; then
   export PHOTOPRISM_DISABLE_VIPS=true
 fi
 


scripts/dist/entrypoint.sh
```diff
@@ -77,6 +77,13 @@ else
   umask 0002
 fi
 
+# Check if the current kernel is recent enough, see https://github.com/photoprism/photoprism/issues/4339
+MIN_KERNEL='3.17'
+OLDER_KERNEL=$(printf "$MIN_KERNEL\n$(uname -r)" | sort -V | head -1)
+if [ ! "$OLDER_KERNEL" = "$MIN_KERNEL" ]; then
+  PHOTOPRISM_DISABLE_VIPS=true
+fi
+
```
rmarinho:
```suggestion
# Check if the current kernel is recent enough, see https://github.com/photoprism/photoprism/issues/4339
MIN_KERNEL='3.17'
OLDER_KERNEL=$(printf "$MIN_KERNEL\n$(uname -r)" | sort -V | head -1)
if [[ ! "OLDER_KERNEL" = "MIN_KERNEL" ]]; then
  export PHOTOPRISM_DISABLE_VIPS=true
fi

```

rmarinho (commented):
I test on my synology NAS  with docker with this little changes and it worked fine.

lastzero:
@petertrr Thank you very much! While the check should help to avoid the error, how about we provide a separate image for older kernels, e.g. based on Debian or the previous LTS version of Ubuntu Linux?

This should make it possible to keep using `libvips`, which provides full color profile support and probably also requires less memory... provided that anyone using PhotoPrism on a NAS (with an old kernel) is able to change the image tag and knows how to do it?

lastzero:
Related Discussions:
- https://lore.kernel.org/all/20220726195822.1223048-1-Jason@zx2c4.com/t/
- https://lore.kernel.org/all/?q=arc4random

rmarinho:
The workaround has been working for me. we should merge this

lastzero:
@rmarinho Besides better color reproduction for non-SRGB images, the advantages of libvips are faster indexing and lower memory usage, which might be especially important for users of old hardware with old kernels. However, this workaround would also disable libvips in combination with a version of glibc that does NOT cause errors when used with an old kernel.

That's why I asked. I understand that the error disappears this way, but disabling libvips and not being able to turn it back on (because the workaround is always applied) also has disadvantages...

Also note that you don't have to wait for this to get merged if there are problems, as you can manually set `PHOTOPRISM_DISABLE_VIPS` to `"true"` in your Docker Compose config file.

lastzero:
@rmarinho @petertrr According to this report, `rawtherapee-cli` and `darktable-cli` also fail with a recent version of glibc (e.g. in our latest Docker image) and a very old kernel (such as used on Synology devices):

- https://github.com/photoprism/photoprism/issues/4339#issuecomment-2455194098

**This means that the workaround will NOT solve all problems and we should instead provide an alternative image, e.g. based on Ubuntu Jammy, which comes with an OLD version of glibc that is compatible with OLD kernels.**

lastzero:
We've created a new legacy image for older devices:

- https://hub.docker.com/r/photoprism/photoprism/tags?page=1&name=legacy

It is based on Ubuntu Jammy 22.04 LTS, so you should be able to use it with older devices and operating systems. Note, however, that bundled applications and tools such as Darktable and RawThereapee may also be older, so newer file formats (especially for RAW images) may NOT be supported. If you need the latest software, please upgrade your NAS device or use PhotoPrism on a computer with a recent kernel, as we do not have the resources to work around this (also considering that our software is mostly free).

You can test the preview build by specifying the following image name (no final release yet as we are waiting for your feedback):

```
photoprism/photoprism:preview-legacy
```

Our documentation explains how to update the image you are using:

- https://docs.photoprism.app/getting-started/updates/

lastzero:
Since we have not received any problem reports regarding the `photoprism/photoprism:preview-legacy` build, a `photoprism/photoprism:legacy` image based on Ubuntu Jammy 22.04 LTS is now available as part of our stable release. I am therefore closing this PR.

Please note, however, that it is only offered as a (temporary) convenience to give you more time to upgrade, and that we cannot provide any support for it, especially since some incompatibilities with newer file formats are to be expected due to the older dependencies that ship with Jammy! 🐛

See also:

- https://github.com/photoprism/photoprism/issues/4339#issuecomment-2673765576

PR #4490, opened 2024-08-29
Backend: Fix constant conditions and possible panic

songzhibin97:
Acceptance Criteria:

### DIFF
--- a/internal/entity/album.go
+++ b/internal/entity/album.go
@@ -591,10 +591,6 @@ func (m *Album) UpdateTitleAndState(title, slug, stateName, countryCode string) 
 		return nil
 	}
 
-	if title != "" {
-		m.SetTitle(title)
-	}
-
 	return m.Updates(Map{"album_title": m.AlbumTitle, "album_slug": m.AlbumSlug, "album_location": m.AlbumLocation, "album_country": m.AlbumCountry, "album_state": m.AlbumState})
 }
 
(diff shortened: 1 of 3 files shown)


internal/entity/album.go
```diff
@@ -591,10 +591,6 @@ func (m *Album) UpdateTitleAndState(title, slug, stateName, countryCode string)
 		return nil
 	}
 
-	if title != "" {
```
lastzero, line 594: Why is the title removed/ignored here?

internal/entity/marker.go
```diff
@@ -358,7 +358,11 @@ func (m *Marker) SyncSubject(updateRelated bool) (err error) {
 		UpdateColumns(Map{"subj_uid": m.SubjUID, "subj_src": SrcAuto, "marker_review": false}).Error; err != nil {
 		return fmt.Errorf("%s (update related markers)", err)
 	} else if res.RowsAffected > 0 && m.face != nil {
-		log.Debugf("markers: matched %s with %s", subj.SubjName, m.FaceID)
+		subjName := ""
+		if subj != nil {
+			subjName = subj.SubjName
+		}
+		log.Debugf("markers: matched %s with %s", subjName, m.FaceID)
```
lastzero: Based on the code above, subj should never be nil here. However, I've added a Subject.String() function which can handle nil values and also makes the code easier to read, see https://github.com/photoprism/photoprism/commit/fc5d8cb9a9c5e4254ddc08bbcf3dee426505d0c7.

internal/entity/query/sessions.go
```diff
@@ -11,7 +11,7 @@ import (
 
 // Session finds an existing session by its id.
 func Session(id string) (result entity.Session, err error) {
-	if l := len(id); l < 6 && l > 2048 {
+	if l := len(id); l < 6 || l > 2048 {
```
lastzero, line 14: Good find! I've fixed this in https://github.com/photoprism/photoprism/commit/1e3f9f7984cddc3d8c7e02f440ae7c9b55366eeb.

lastzero:
@songzhibin97 Thanks again for submitting these fixes! Both seem to be resolved now, so I will go ahead and close this PR as it is no longer needed. :white_check_mark:

Issue #668, opened 2020-11-30
Why not using WEBP instead of JPEG? 


ivangjxyz:
I see that the web views of the files are transformed to .JPG, but I'd find it even cooler if .WEBP was used, or at least to have an option. 

Many big websites like IKEA are using .WEBP images already, so I don't see that compatibilities are any issue any more... and the thing is that... It's better at compressing images, in size-wise **AND** quality-wise. 

[https://developers.google.com/speed/webp](https://developers.google.com/speed/webp)

prog32:
Just note

**WebP** has global **coverage ~90%**, reduces JPEG **size by ~30%**.
https://caniuse.com/?search=webp

Even more modern format **AVIF** (keyframe from AV1), has **global coverage ~25%**, reduces JPEG **size by ~50%** (dose not support progressive mode - image as JPEG/WebP)
https://caniuse.com/avif

lastzero:
Maybe later, won't do such a change shortly before a releasy. Also keep in mind that while not ALL devices support WebP, you need both. So you save nothing and storage is cheap anyway, especially in comparison in the time & money it costs us to change everything.

tbrtje:
All desktop and most android browsers already support WebP and Safari gained WebP-support in Version 14 on Desktop and iOS 14 on mobile, so there are no compatibility issues. You dont need JPEG anymore. 
Storage isnt an issue, youre right, but bandwith often is.

lastzero:
Guess "most" is the keyword here... we'll have an eye on this. After 3 years of hard, unpaid work, we're happy PhotoPrism is finally stable. Give us a break 👍

benmccann:
This could probably be an option. It's really only IE11 and MacOS 10 and older that don't support webp. If I'm running a personal PhotoPrism instance then I know I'm not accessing it in that manner and am only using devices that support webp and so I'd rather disable jpeg to improve indexing speed and save disk space.

lastzero:
What about Photoshop? At best, there seem to be external plugins for WebP support. Haven't seen it in the standard format dropdown in 2021. That's a no go for me. Besides, my mother might also struggle to open a webp file when I send it to her via email.

ivangjxyz:
But I think the idea here is for the web interface to be faster and lighter. Obviously, using JPG for the "download" button still obviously makes the most sense. 

I think it is best to load 50 lighter _thumbnails_ and be able to download "that one picture" in JPG. 

[...]

PR #4495, opened 2024-08-31
4494/filtering based on quality

TheRanomial:
`func (w *Convert) isHDRVideo(filename string) (bool, error) {
	ffprobeBin := "ffprobe" // Default to using ffprobe from PATH

	// Check if FFprobe binary path is explicitly set in config
	if w.conf.FFmpegBin() != "" {
		// Assume FFprobe is in the same directory as FFmpeg
		ffprobeBin = strings.Replace(w.conf.FFmpegBin(), "ffmpeg", "ffprobe", 1)
	}

	cmd := exec.Command(ffprobeBin, "-v", "error", "-select_streams", "v:0", 
		"-show_entries", "stream=color_space,color_transfer,color_primaries", 
		"-of", "default=noprint_wrappers=1", filename)
	
	output, err := cmd.CombinedOutput()
	if err != nil {
		return false, fmt.Errorf("ffprobe failed: %v", err)
	}

	outputStr := string(output)
	return strings.Contains(outputStr, "bt2020") || strings.Contains(outputStr, "smpte2084") || strings.Contains(outputStr, "arib-std-b67"), nil
}`

`if f.IsAnimated() && !f.IsWebP() && w.conf.FFmpegEnabled() {
		timeOffset := ffmpeg.PreviewTimeOffset(f.Duration())

		isHDR, err := w.isHDRVideo(f.FileName())
		if err != nil {
			log.Warnf("jpeg: failed to detect HDR status: %s", err)
			// Fallback to non-HDR processing if detection fails
			isHDR = false
		}

		ffmpegArgs := []string{
			"-y", "-ss", timeOffset, "-i", f.FileName(),
			"-vframes", "1",
		}

		if isHDR {
			// Apply HDR to SDR conversion filters only for HDR content
			ffmpegArgs = append(ffmpegArgs, 
				"-vf", "zscale=t=linear:npl=100,format=gbrpf32le,zscale=p=bt709,tonemap=tonemap=hable:desat=0,zscale=t=bt709:m=bt709:r=tv,format=yuv420p",
				"-q:v", "2", // Higher quality for HDR content
			)
		} else {
			// For non-HDR content, just set the quality
			ffmpegArgs = append(ffmpegArgs, "-q:v", "5")
		}

		ffmpegArgs = append(ffmpegArgs, jpegName)
		// Improved FFmpeg command for better HDR to SDR conversion
		result = append(result, NewConvertCommand(
			exec.Command(w.conf.FFmpegBin(), ffmpegArgs...)),
		)
	}`

### DIFF
--- a/internal/photoprism/convert_image_jpeg.go
+++ b/internal/photoprism/convert_image_jpeg.go
@@ -30,9 +30,15 @@ func (w *Convert) JpegConvertCommands(f *MediaFile, jpegName string, xmpName str
 
 	// Extract a still image to be used as preview.
 	if f.IsAnimated() && !f.IsWebP() && w.conf.FFmpegEnabled() {
-		// Use "ffmpeg" to extract a JPEG still image from the video.
+		timeOffset := ffmpeg.PreviewTimeOffset(f.Duration())
+		
+		// Improved FFmpeg command for better HDR to SDR conversion
 		result = append(result, NewConvertCommand(
-			exec.Command(w.conf.FFmpegBin(), "-y", "-ss", ffmpeg.PreviewTimeOffset(f.Duration()), "-i", f.FileName(), "-vframes", "1", jpegName)),
+			exec.Command(w.conf.FFmpegBin(), "-y", "-ss", timeOffset, "-i", f.FileName(),
+				"-vframes", "1",
+				"-vf", "zscale=t=linear:npl=100,format=gbrpf32le,zscale=p=bt709,tonemap=tonemap=hable:desat=0,zscale=t=bt709:m=bt709:r=tv,format=yuv420p",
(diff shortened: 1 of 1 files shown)


lastzero:
@TheRanomial Considering the effort to work around this problem when using 8-bit JPEG images to display 10 or 12-bit HDR content, it seems better to invest the time in generating (at least) these thumbnails in the newer AVIF format, which is much better suited for displaying HDR content:

- https://github.com/photoprism/photoprism/issues/668

Unfortunately, since we need to work on batch editing and AI features first, this will have to wait until that is done. The implemented logic is therefore no longer needed, so I will go ahead and close this PR.

Thanks a lot for your contribution! :star:

PR #4506, opened 2024-09-06
CI: Add codespell config and fix typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -4,6 +4,6 @@ skip = .git*,*.svg,go.sum,package-lock.json,*.css,.codespellrc,locales,stopwords
 check-hidden = true
 # ACRONYMS and generally mixed case/Capitalized words
 # Also ignore some German
-ignore-regex = ^(SMethod=shal|biuld: build)$|https://\S+|\b([A-Z][a-zA-Z]+|ist auf)\b
+ignore-regex = ^(SMethod=shal|biuld: build)$|https://\S+|\b([A-Z][a-zA-Z]+|ist auf)\b|"(datin|alis|hel)"|\bfixe == false\b
 # common variables or other tricky cases
 ignore-words-list = renderd,nd,folde,crate,ue,fo,ist,alo,mot,te,admiraal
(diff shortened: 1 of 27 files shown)


yarikoptic:
is there a way to trigger some testing?  otherwise -- this is an easy to review PR, don't get scared by number of files, quite often it is just one typo fix per file.

lastzero:
@yarikoptic I apologize for not getting back to you any sooner!

The reason we've been so reluctant with this is because of security concerns regarding GitHub Actions (especially when adding/modifying them through pull requests). See, for example:

- https://medium.com/@simardeep.oberoi/github-actions-supply-chain-attack-how-tj-actions-changed-files-leaked-secrets-and-what-you-need-9e2677e07f68

That being said, we'd be happy to merge the spelling fixes without the action configuration and/or add an additional Make target that [runs a spell check](https://github.com/golangci/golangci-lint/blob/f15e6c4865d5b55c04e192a6ac49c4117ff0b1c7/pkg/golinters/misspell/misspell.go) locally, just like eslint. It would be worth looking at https://github.com/golangci/golangci-lint for this as well, since `golangci-lint` might be a better and more versatile choice for our needs.

Would that be possible? I'll be happy to help with this and answer any questions :)

yarikoptic:
sure -- concerns are well understood I must say especially if you have/use some secrets (I might have missed but I do not think you do yet). Anything is fine with me. I can easily redo and drop CI action.  Just let me know if you would have time now to review so we do not breed conflicts and require re-doing?

yarikoptic:
actually, "easier done than said" so here is an updated -- dropped CI, rebased, added a few ignore pragmas, less changes to review

lastzero:
Excellent, thank you very much! I'll go ahead and merge this.

lastzero:
@yarikoptic This is now included in our [preview build](https://docs.photoprism.app/release-notes/#development-preview)! Note that I had to fix some minor issues, for example using `#` for comments is not allowed in Go. I assume that using `//` instead shouldn't affect the codespell annotation.

yarikoptic:
Oh, so sorry that my Python rotten mind saw everything as Python code. Yes, it should work with a proper to the language comment

PR #4549, opened 2024-09-24
Videos: Add log-statement to convert_video_avc.go

Akashic101:
Just a small change, this prints out that no conversion is needed when a video already had been previously converted. Currently running `docker compose exec photoprism photoprism convert` is a very quiet command with basically no output if already run at a previous point, leaving me unsure how much progress has been achieved. This small log fixes that by providing a quick log about where the converter is right now

Acceptance Criteria:


lastzero:
@Akashic101 My apologies for not getting back to you any sooner!

We should be able to merge this if you change the log message as follows:

```
convert: %s is AVC encoded, no transcoding required
```

lastzero:
@Akashic101 I'm happy to report that your proposal has been implemented as part of the above commit. I will therefore close this PR as it is no longer needed. Thank you for your contribution! :+1:

PR #4556, opened 2024-09-27
Frontend: Upgrade Vuetify from v1.5 to v3.7

AsikNasik:
Acceptance Criteria:


frontend/package.json
```diff
@@ -119,5 +119,9 @@
   },
   "browserslist": [
     ">0.25% and last 2 years"
-  ]
+  ],
+  "devDependencies": {
+    "@mdi/font": "^7.4.47",
+    "eslint-plugin-vuetify": "^2.4.0"
```
lastzero, line 128: You're welcome to move the other eslint packages to devDependencies as well, if that has an advantage?

So far my understanding is that it makes no difference for our use case, as **npm** is only used for development, while the production build only contains precompiled JS/CSS files and therefore never contains any dev dependencies (whether they are listed as such or not)...

frontend/package.json
```diff
@@ -119,5 +119,9 @@
   },
   "browserslist": [
     ">0.25% and last 2 years"
-  ]
+  ],
+  "devDependencies": {
+    "@mdi/font": "^7.4.47",
```
lastzero: Due to some breaking changes in the [material icons font](https://www.npmjs.com/package/material-design-icons-iconfont) that we use, we added the required CSS and font files to our own repository under [`/frontend/src/css/vendor/icons`](https://github.com/photoprism/photoprism/tree/develop/frontend/src/css/vendor/icons) (AFAIK, the original package no longer included a plain CSS file or the font paths in it were invalid, so the icons broke in the UI after updating the package):
- https://www.npmjs.com/package/material-design-icons-iconfont
- https://jossef.github.io/material-design-icons-iconfont/

Instead of using/importing the original npm package, `/frontend/src/css/vendor/iconsmaterial-design-icons.css` is imported in our main CSS file:
https://github.com/photoprism/photoprism/blob/0b391d26ecf3989becef2b8280a15b0014e709f2/frontend/src/css/app.css#L26

If Vuetify 2/3 use `@mdi/font` instead of `material-design-icons-iconfont` then that's fine with us as both should support/include similar icons since they are also based on Material Design:
- https://fonts.google.com/icons

Note, however, that the naming of some icons MAY be different and that you will therefore probably need to change some templates so they use the new Material Design font icons:
- https://www.npmjs.com/package/@mdi/font
- https://pictogrammers.com/library/mdi/

Also, would it make sense to list `@mdi/font` under the regular `dependencies` in the `package.json` file, since the fonts are always required? See the comment below.

lastzero (commented):
Here is some initial feedback for you, mainly regarding the icons. Since these are important e.g. for the navigation, it might make sense to wait with further tests until your work on them is finished?

frontend/package.json
AsikNasik: Okay, I will check it later if it has any sense to do it. For now I propose to focus to the migration and after I will do some investigation about this question. What do you think about it?

frontend/package.json
AsikNasik: Yes, you are right.
I will move `@mdi/font` package to the `dependencies` because we always need them.

frontend/src/css/themes.css
AsikNasik, line 1: Will we keep this comment or will delete in the future?

frontend/src/css/themes/default.css
```diff
@@ -1,5 +1,5 @@
 /* Component Color Defaults */
-
+/*
```
AsikNasik: Will we keep this comment or will delete in the future?

dannyadair:
What an awesome effort!

Issue #4074, opened 2024-02-22
WebDAV: Adding TrueNAS as a service for syncing files does not work


Pursch:
Hi!

**1. What is not working as documented?**
Sync Access via WebDAV running on a Truenas server seems to be broken. No matter if I choose up- or download, I get the same error messages.

Error with activated WebDAV Basic Auth:
`sync: 401 Unauthorized: <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN"> <html><head> <title>401 Unauthorized</title> </head><body> <h1>Unauthorized</h1> <p>This server could not verify that you are authorized to access the document requested. Either you supplied the wrong credentials (e.g., bad password), or your browser doesn't understand how to supply the credentials required.</p> </body></html>`

There is no error, if I use the same settings from the command line within the app:
```
root@231128:/photoprism$ curl -X PROPFIND -H "Depth: 1" -u user:password http://192.168.0.2:30034/share/
<?xml version="1.0" encoding="utf-8"?>
<D:multistatus xmlns:D="DAV:">
<D:response xmlns:lp2="http://apache.org/dav/props/" xmlns:lp1="DAV:">
<D:href>/bilder/</D:href>
<D:propstat>
<D:prop>
<lp1:resourcetype><D:collection/></lp1:resourcetype>
<lp1:creationdate>2024-02-22T16:35:07Z</lp1:creationdate>
<lp1:getlastmodified>Thu, 22 Feb 2024 16:35:07 GMT</lp1:getlastmodified>
<lp1:getetag>"27-611fb06b4072d"</lp1:getetag>
<D:supportedlock>
<D:lockentry>
<D:lockscope><D:exclusive/></D:lockscope>
<D:locktype><D:write/></D:locktype>
</D:lockentry>
<D:lockentry>
<D:lockscope><D:shared/></D:lockscope>
<D:locktype><D:write/></D:locktype>
</D:lockentry>

```
Accessing the URL by webbrowser works fine too.

Error without WebDAV Auth:
```
sync: HTTP multi-status request failed: 200 OK
service-folder: HTTP multi-status request failed: 200 OK
```
**2. How can we reproduce it?**
Got to Settings -> Services -> Connect+
Add a Webdav server running as Truenas App -> Save
Configure Sync, no folder selection is possible -> Save
-> A triangle is shown on the sync symbol.
Lookup error messages.

I configured a second service to connect to a Nextcloud server without problems.

**3. What behavior do you expect?**
[...]

Issue #3541, opened 2023-07-11
WebDAV: Add fallback for servers that only allow PROPFIND requests with a Depth of 1


leruaa:
#### 1. What is not working as documented?

I'm trying to connect to my pCloud account via WebDAV, and I get the following error: "Could not connect, please try again". In the logs, I only see:

```
time="2023-07-11T06:20:18Z" level=error msg="could not connect"
```

#### 2. How can we reproduce it?

Unfortunately, WebDAV is only avaliable to pCloud paid customer, but I'm able to connect with Windows 10 using the following address: `https://ewebdav.pcloud.com`, my pCloud login and password.

1. Go to 'Settings > Services'
2. Click on 'Connect'
3. Enter the url `https://ewebdav.pcloud.com`, a pCloud login and password.

#### 3. What behavior do you expect?

I should be able to connect to my pCloud account

#### 4. What could be the cause of your problem?

N/A

#### 5. Can you provide us with example files for testing, error logs, or screenshots?

N/A

#### 6. Which software versions do you use?

(a) PhotoPrism Architecture & Build Number: Build [230625-17242fb07]

(b) Database Type & Version: SQLite,

(c) Operating System Types & Versions: Linux

(d) Browser Types & Versions: Firefox

#### 7. On what kind of device is PhotoPrism installed?

N/A

#### 8. Do you use a Reverse Proxy, Firewall, VPN, or CDN?

N/A

graciousgrey:
WebDAV in general works, so the problem is probably pCloud specific. 

Are there other pCloud users who can reproduce this problem?

seb-ol:
> WebDAV in general works, so the problem is probably pCloud specific.
> 
> Are there other pCloud users who can reproduce this problem?

Just tried from multiple client, Qnap Windows and on Ubuntu, I have error connexion quite often. it's really painfull to use. Really slow throught webdav on windows while working fine with pcloud client. and .. cannot sync more than 1Gig of data before error on QNAP. 
Hope it's just temporary ..

Snijder:
I have the same issue, unfortunately photoprism does not log anything about why the connection fails.
The only thing I've noticed is that the username for pcloud webdav is an email and the pcloud webdav is not very fast. 

When accessing via the browser a very simple index shows up:
![afbeelding](https://github.com/photoprism/photoprism/assets/10807997/5513b95f-83c0-44ac-8cc8-8e6ee8dfb271)

[...]

PR #4608, opened 2024-11-03
WebDAV: Add depth header to service discovery request for Hetzner compatibility

esteve:
I noticed that I couldn't connect to my Hetzner storage box via webdav, as it does not allow `PROPFIND` requests without `Depth` set as `0` or `1`. When a request is created via `NewRequest` it doesn't have a `Depth` field, which Hetzner interprets as `infinity`:

```
$ curl -X PROPFIND https://XXXX:XXXXX@XXXXXXXXX/webdav/
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
<html>
<head>
<title>403 Forbidden</title>
</head><body>
<h1>Forbidden</h1>
<p>PROPFIND requests with a Depth of "infinity" are not allowed for /webdav/.</p>
</body></html>
```

I believe this is the same issue as https://github.com/photoprism/photoprism/issues/4074 https://github.com/photoprism/photoprism/issues/3541

I've tested the change locally and it works. However photoprism still uses `PROPFIND` with `Depth = infinity` somehwere else, I'd suspect it's because of https://github.com/photoprism/photoprism/blob/c096382dbd6e891c2408466bce7642f39d2fd14c/internal/workers/sync_refresh.go#L28 , but this is the first Go code I've ever written and I'm not familiar with the `http` and `webdav` packages.

### DIFF
--- a/internal/service/remote.go
+++ b/internal/service/remote.go
@@ -44,6 +44,7 @@ const (
 
 func HttpOk(method, rawUrl string) bool {
 	req, err := http.NewRequest(method, rawUrl, nil)
+	req.Header.Add("Depth", "1")
 
 	if err != nil {
 		return false


gOOvER:
i have the same problem with hetzner and i would be happy, when this could be merged asap

lastzero:
Thank you so much for submitting this PR! My apologies for not getting back to you sooner. I've been helping our team with the frontend update, which is now complete, so I can finally check the open PRs and merge them :)

lastzero:
@esteve The above changes allow you to set custom request headers in the service type configuration, rather than adding a `Depth` header to all requests.

However, you are right that this may not be enough for you to use the service, since our WebDAV client currently only makes a single, recursive request to get a list of files and folders on the remote server:

- https://github.com/photoprism/photoprism/issues/4074
- https://github.com/photoprism/photoprism/issues/3541

This could be solved by implementing a fallback that makes multiple requests to get the contents of the remote folders for synchronization; for example, you could implement the Directories() function differently and use it when certain errors are detected in line 148 (see the source code of earlier releases for inspiration):
https://github.com/photoprism/photoprism/blob/c096382dbd6e891c2408466bce7642f39d2fd14c/internal/service/webdav/client.go#L142-L165

lastzero:
This can now be tested with our development preview build:

esteve:
@lastzero thank you for merging this PR 🙂

Issue #4698, opened 2025-01-07
Viewer: Improve playback of videos, animations & live photos


lastzero:
**As a user with many videos, animations and/or live photos, I would like to see the following improvements implemented for the new [hybrid photo/video viewer](https://github.com/photoprism/photoprism/issues/1307) in `component/viewer.vue` so that I can safely enjoy them in the next stable release:**

- [x] Check if there is a more convenient and/or secure way to render the video slide, then perform security tests to ensure that no code can be injected, e.g. create an [`HTMLVideoElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement) object, set the properties based on the media type/video duration, and then return it instead of the plain HTML as implemented in the proof-of-concept.
- [x] Live Photos and Animations (e.g. GIFs) should be looped and played automatically (autoplay attribute).
- [x] If the browser can naively handle the video file format, don't default to the AVC video URL, as this may require transcoding, which is slow and resource-intensive. For this, the `Util.videoUrl()` function has a second argument for the codec (might need to be added to the server response, which is something we can help with).

**Additional notes and requirements:**

- [x] Live photos, animations and/or videos should NOT loop [while a slideshow is running](https://github.com/photoprism/photoprism/issues/4702) (to be implemented), otherwise you might get stuck on that slide. However, it seems to be a good idea to play them automatically while a slideshow is running (the next slide should only open when the video/animation is finished/paused).
- [x] Be sure to additionally test the Viewer (with all media types) in [Library > Originals](https://docs.photoprism.app/user-guide/library/files/) and in the [Places](https://docs.photoprism.app/user-guide/organize/places/) view, as it works with limited data fields to improve query performance. So it's possible that something is missing for media/file types other than regular images.

**Once this is released, the following enhancements can be worked on and shipped in a future release:**

1.  We MAY consider [using the `.m3u8` file format](https://github.com/photoprism/photoprism/issues/4179) for specifying the stream URL(s) if it improves the user experience in any way (develop a simple/static proof-of-concept first to see if/how it works).
2. The server SHOULD (additionally) provide a video/animation still from time index 0 that can be used as a poster (the current thumbnail is taken later for longer videos, since the first frame is often black).

**Related Issues:**

- https://github.com/photoprism/photoprism/issues/1307
- https://github.com/photoprism/photoprism/issues/3372

lastzero:
Two additional notes on this issue:

[...]

PR #4708, opened 2025-01-08
Viewer: Improve loading of videos, animations, and live photos #4698

omerdduran:
Related Issue:
- #4698


AsikNasik (approved):

frontend/src/component/viewer.vue
```diff
@@ -412,31 +409,49 @@ export default {
...
-            };
-          } else {
-            return {
-              html: `<video class="pswp__video" controls playsinline poster="${imgSrc}" preload="metadata"><source src="${videoSrc}" /></video>`,
-            };
-          }
+          return {
+            type: "html",
+            html: `<div style="width: 100%; height: 100%;">Loading...</div>`,
+            w: window.innerWidth, // Use window width for video
```
lastzero: Note that this is not the case when the (yet to be finalized) sidebar is visible. Is there any reason to pass the width and height here, and if so, what effect does it have (since the content will be rendered by our callback anyway)?

frontend/src/component/viewer.vue
```diff
@@ -412,31 +409,49 @@ export default {
...
-              html: `<video class="pswp__video" autoplay controls playsinline poster="${imgSrc}" preload="auto"><source src="${videoSrc}" /></video>`,
-            };
-          } else {
-            return {
-              html: `<video class="pswp__video" controls playsinline poster="${imgSrc}" preload="metadata"><source src="${videoSrc}" /></video>`,
-            };
-          }
+          return {
+            type: "html",
+            html: `<div style="width: 100%; height: 100%;">Loading...</div>`,
```
lastzero: It would be good not to use inline styles unless absolutely necessary. If you do need styles, please assign classes that follow the conventions - in this case it could be `class="pswp__content"` or `class="pswp__html"` - and then do the styling in `src/css/viewer.css`. Since `pswp__content` also seems to be used for image slides, it may already have some styles attached to it, but you will have to test if this is good or if we want to use something else like `pswp__html` to avoid this.

frontend/src/component/viewer.vue
```diff
@@ -446,6 +461,73 @@ export default {
       // Publish event to be consumed by other components.
       this.$event.publish("viewer.opened");
     },
+
+    getVideoProperties(model) {
```
lastzero: This should not be necessary if we add the required data to the response, right? Otherwise, this will perform a new request for each slide, which can be slow and thus not good for UX. I can add the values for you if you give me a list.

frontend/src/component/viewer.vue
```diff
@@ -412,31 +409,49 @@ export default {
...
-          src: imgSrc, // Thumbnail image URL.
-          width: model.Thumbs[s].w, // Actual thumbnail image width (x).
-          height: model.Thumbs[s].h, // Actual thumbnail image height (y).
+          src: imgSrc,
+          width: model.Thumbs[thumbSize].w,
+          height: model.Thumbs[thumbSize].h,
         };
       });
 
+      lightbox.on("contentLoad", async (e) => {
```
lastzero: `async` can be removed when `getVideoProperties()` is not needed anymore, right?

frontend/src/component/viewer.vue
```diff
@@ -412,31 +409,49 @@ export default {
...
-            return {
-              html: `<video class="pswp__video" controls playsinline poster="${imgSrc}" preload="metadata"><source src="${videoSrc}" /></video>`,
-            };
-          }
+          return {
+            type: "html",
+            html: `<div style="width: 100%; height: 100%;">Loading...</div>`,
+            w: window.innerWidth, // Use window width for video
+            h: window.innerHeight, // Use window height for video
+            msrc: imgSrc,
```
lastzero: Same question for "msrc" as for "w" and "h": What effect does it have (since the content is rendered by our callback anyway)? Wouldn't it make (more) sense to pass the model data e.g. in a "model" property (once the `getVideoProperties()` callback is no longer needed)?

frontend/src/component/viewer.vue
```diff
@@ -446,6 +461,73 @@ export default {
...
+
+    createVideoElement(videoSrc, posterSrc, autoplay = false, shouldLoop = false) {
+      // Create video element
+      const video = document.createElement("video");
+
+      // Set essential attributes
+      video.className = "pswp__video";
+      video.controls = true;
+      video.playsInline = true;
+      video.style.cssText = "position: absolute; top: 0; left: 0; width: 100vw; height: 100vh; object-fit: contain; background: black;";
```
lastzero: It would be good to not use inline styles unless absolutely necessary. If you need styles, please do the styling in `src/css/viewer.css`.

frontend/src/component/viewer.vue
```diff
@@ -446,6 +461,73 @@ export default {
...
+        video.autoplay = true;
+      }
+
+      // Create and append source element
+      const source = document.createElement("source");
+      source.src = videoSrc;
+
+      video.appendChild(source);
+
+      // Convert the element to HTML string for PhotoSwipe
```
lastzero: In other places PhotoSwipe seems to accept dom nodes as return value. Are you sure this isn't possible, or would we need to use a different filter in this case? Just to make sure we have the proper callbacks in place and haven't missed anything... Otherwise, the video element looks fine (without having tested it in a browser, of course).

lastzero (changes requested):
Please check my comments and let me know what values to add to the model data so you don't need the additional API requests. Thank you! :)

frontend/src/component/viewer.vue
lastzero: The function to use to get the viewport size is `this.getLightboxViewport()`, if you need it.

lastzero:
I just added `Type`, `Lat`, `Lng`, and `Duration` to the viewer response data, so you should no longer need the extra backend request to display videos, animations, and/or live photos - see https://github.com/photoprism/photoprism/commit/cedb30692948e15c1292c385ec9d63bb5c332a62. Please merge this from `upgrade/vuetify-3` into your branch!

lastzero:
This commit additionally adds the media `Codec` to the response data, so you can use it to play the original format if the browser supports it (see my comments in `viewer.vue`): https://github.com/photoprism/photoprism/commit/1a938a387f8f8f9a7022d62ef06b9efe45cda859

lastzero:
Please also note the following:

- [ ] Live photos, animations and/or videos should NOT loop while a slideshow is running (to be implemented), otherwise you might get stuck on that slide. However, it seems to be a good idea to play them automatically while a slideshow is running (the next slide should only open when the video/animation is finished/paused).
- [ ] The backend will automatically flag short videos up to 3 seconds as live photos, so you do NOT need to check the duration in the frontend code to enable looping - checking if it is a live photo or animation should be enough (TBD). In a future release, the metadata might additionally include "loop" and "mute" flags, allowing users to decide for each video how it should be played.
- [ ] Be sure to additionally test the Viewer (with all media types) in [Library > Originals](https://docs.photoprism.app/user-guide/library/files/) and in the [Places](https://docs.photoprism.app/user-guide/organize/places/) view, as it works with limited data fields to improve query performance. So it's possible that something is missing for media/file types other than regular images.

lastzero (approved):
Thank you very much! I will go ahead and merge this so we can do more testing and add improvements if needed.

Issue #4700, opened 2025-01-07
Viewer: Add toggle to flag/unflag pictures as favorites


lastzero:
**As a PhotoPrism user, I would like to see the following implemented for the new [Hybrid Photo/Video Viewer](https://github.com/photoprism/photoprism/issues/1307) in `component/viewer.vue` before releasing it:**

- [x] If the current user has the necessary permissions, add a toggle to the buttons at the top to flag/unflag the current picture as a favorite. Note that the new version uses the star icon instead of a heart for this.

Since the controls were rendered with the VueJS template in the previous release, and the icon should update automatically based on the favorite state of the current picture (`this.model`), you need to **find the best (and most maintainable/reliable) way to implement this** with the new version of PhotoSwipe, e.g. via JS, by setting a CSS class, or by rendering the buttons with VueJS again (which may require some more difficult changes):

![favorite-toggle](https://github.com/user-attachments/assets/e9bfe292-7da2-4aca-abed-c9a07a0ac799)

If you are unsure, please test, discuss, and suggest alternative implementations before committing to a specific approach!

**Related Issues:**

- https://github.com/photoprism/photoprism/issues/4699
- https://github.com/photoprism/photoprism/issues/4701
- https://github.com/photoprism/photoprism/issues/1307

lastzero:
@AsikNasik @omerdduran I tried to improve the contrast of the SVG toggle icon with a CSS filter, but the result was not as good as the icon outline implementation used by PhotoSwipe - so I guess we'll just go with that and add the outline to the specified paths if there's more than one path, for example:

```html
<use 
  class="pswp__icn-shadow pswp__icn-favorite-on" 
  xlink:href="#pswp__icn-favorite-on"></use>
<path 
  d="..." 
  id="pswp__icn-favorite-on"
  class="pswp__icn-favorite-on" />
<use
  class="pswp__icn-shadow pswp__icn-favorite-off"
  xlink:href="#pswp__icn-favorite-off"></use>
<path
  d="..."
  id="pswp__icn-favorite-off"
  class="pswp__icn-favorite-off" />

```

Styles for this in `/frontend/src/css/viewer.css`:

```css
.p-viewer__lightbox.is-favorite .pswp__button--favorite-toggle .pswp__icn-favorite-off {
    display: none;
}

.p-viewer__lightbox:not(.is-favorite) .pswp__button--favorite-toggle .pswp__icn-favorite-on {
    display: none;
}
```

[...]

PR #4709, opened 2025-01-08
Viewer: add a Toggle to flag/unflag pictures as Favorite  #4700

AsikNasik:
These changes add the following feature:

- #4700.

Main changes:

- add buttons like/like-outline (flag/unflag) in the viewer
- add functionality to toggle pictures as Favorite

### DIFF
--- a/frontend/src/component/viewer.vue
+++ b/frontend/src/component/viewer.vue
@@ -1,6 +1,6 @@
 <template>
   <div v-if="visible" ref="container" class="p-viewer" tabindex="-1" role="dialog">
-    <div ref="lightbox" class="p-viewer__lightbox" :class="{ slideshow: slideshow.active, sidebar: sidebarVisible }"></div>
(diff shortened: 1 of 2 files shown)


omerdduran (approved):

lastzero:
This looks like it should work. However, I think we could make it more generic, e.g. by using vue to automatically add an "is-favorite" and "is-selected" class to the lightbox container, similar to how the cards and mosaic views work (and then use that to show/hide the matching icons).

What do you think? Also, once we have that, maybe the HTML for the button could be rendered in a way that avoids having to effectively add two buttons for the same action?

lastzero:
PhotoSwipe should allow you to return the full HTML for the button, e.g. by setting the "html" property or returning a node object, instead of letting it render the final HTML by specifying only the button's SVG path and size.

lastzero:
Thank you very much! I will go ahead and merge this so we can test it and proceed with adding the remaining buttons.

lastzero:
Initial tests look good. Seems to be working as expected! :sparkles:

Issue #4704, opened 2025-01-07
Viewer: Dynamically set image src, width and height to match current zoom level and viewport size


lastzero:
**As a user with high resolution photos, I would like the new [hybrid photo/video viewer](https://github.com/photoprism/photoprism/issues/1307) to automatically load a higher resolution thumbnail based on the current zoom level and screen resolution.**

- [x] Check the [PhotoSwipe 5 documentation](https://photoswipe.com/adjusting-zoom-level/) for the best way to implement this feature request, e.g. use a callback function or provide a srcset to let the browser decide which image to load (if possible/supported).
- [x] Once you know how to do it, give an update (e.g. comment on this issue) and implement it as a proof-of-concept for further UX testing and testing with different browsers/devices.
- [x] The final version must be fully compatible with all previously supported browsers and devices.

Related Issues:

- https://github.com/photoprism/photoprism/issues/247
- https://github.com/photoprism/photoprism/issues/4703
- https://github.com/photoprism/photoprism/issues/4702
- https://github.com/photoprism/photoprism/issues/4701
- https://github.com/photoprism/photoprism/issues/4700
- https://github.com/photoprism/photoprism/issues/4699
- https://github.com/photoprism/photoprism/issues/1307

lastzero:
The PhotoSwipe event to watch for may be one of the following:

```js
lightbox.on('contentResize', ({ content, width, height }) => {
  // content will be resized
  // can be default prevented
  console.log('contentResize', content, width, height);
});
lightbox.on('imageSizeChange', ({ content, width, height, slide }) => {
  // content.element is image
  console.log('imageSizeChange', content, width, height, slide, slide.index);
});
```

See the available documentation for further information:

- https://photoswipe.com/events/#slide-content-events
- https://photoswipe.com/adjusting-zoom-level/

The thumbnail sizes available from our backend API are documented here:

- https://docs.photoprism.app/developer-guide/api/thumbnails/#thumbnail-sizes

Related source code:

https://github.com/photoprism/photoprism/blob/d286e4317418d20acd55f3f47daeb9d815ecaece/frontend/src/component/viewer.vue#L282-L289

PR #4722, opened 2025-01-15
Feature/dynamically set image src, width and height to match current zoom level and viewport #4704

AsikNasik:
This pull request enhances the image viewer component by dynamically loading higher quality thumbnails when the image zoom level changes.

As users zoom in, the function automatically switches to higher resolution thumbnails to maintain visual clarity. It calculates the current zoomed dimensions and determines the optimal thumbnail size needed, loading a higher quality version only when necessary. To optimize performance and prevent redundant network requests, the function includes checks to ensure higher quality versions aren't reloaded if they're already being displayed.

Acceptance Criteria:


frontend/src/component/viewer.vue
```diff
@@ -294,6 +306,10 @@ export default {
           alt: model?.Title,
         };
 
+        // Save the start width and height of the image
+        this.imageThumbsSize.width = model.Thumbs[thumbSize].w;
```
lastzero, line 310: It looks as if `imageThumbsSize` if a global component property, while the "itemData" callback is called repeatedly by PhotoSlide and not only for the currently active slide. So `imageThumbsSize` might not be set to the width and height of the current thumbnail, but simply to the last one PhotoSlide processed (the order is out of our control).

As I understand it, you could instead add an initial size property to the slide data (if knowing the current slide size, which is already set as width and height, is not enough), for example:

```js
const img = {
  src: model.Thumbs[thumbSize].src,
  width: model.Thumbs[thumbSize].w,
  height: model.Thumbs[thumbSize].h,
  initialWidth: model.Thumbs[thumbSize].w,
  initialHeight: model.Thumbs[thumbSize].h,
  alt: model?.Title,
};
```

Note, however, that the user is able to change the browser window size (or shrink the viewport by expanding the new sidebar) at any time. So I'm not exactly sure if knowing the original/initial size is of much/any value, as it's not necessarily the right/best size to use (or return to)?

frontend/src/component/viewer.vue
```diff
@@ -1052,6 +1068,66 @@ export default {
...
+      if (zoomLevel < 0.95) return;
+
+      const currentWidth = Math.round(currSlide.width * zoomLevel);
+      const currentHeight = Math.round(currSlide.height * zoomLevel);
+
+      // Find the best matching thumb size based on zoomed dimensions
+      let bestThumbSize = null;
+      let bestThumbArea = Infinity;
+
+      Object.entries(model.Thumbs).forEach(([size, thumb]) => {
```
lastzero, line 1092: I'll see if we can improve the backend API so that the image URL stays the same for larger thumbnail types if the actual size doesn't increase due to a smaller original. In that case, you could use the existing function to get the thumb type (as in itemData) and then only load/change the image if the URL has changed, right?

omerdduran: Yes, we can do that

lastzero: The backend API has been updated, so that the response will now look as in this example:

viewer-response

PR #4738, opened 2025-01-21
Viewer: Press space bar to play/pause videos & hide zoom action

omerdduran:
Add space bar control for playing/pausing videos and disable zoom icon for video content. 

While most browsers handle space bar control through video focus, Safari requires an additional event handler for consistent behavior.


lastzero:
Thank you very much! As a general thought: Is there a more generic/abstract way to handle keyboard events that would integrate well with our Vue 3 application?

For this, it's important not to simply attach global listeners that always react. Instead, the keyboard event callbacks must only be triggered when the element/context is active/has focus. As a starting point, here is what I found on Google:

- https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Frameworks_libraries/Vue_refs_focus_management#the_focus_management_problem
- https://vuejs.org/guide/essentials/event-handling.html
- https://github.com/pdanpdan/vue-keyboard-trap
- https://vuetifyjs.com/en/features/accessibility/#focus-management-and-keyboard-interactions

Unfortunately, I won't be able to test this solution and suggest possible improvements today because I need to take a break and tomorrow is my birthday. So I've asked @AsikNasik to help review this PR if she has time. Also, be sure to read the article on [developer.mozilla.org](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Frameworks_libraries/Vue_refs_focus_management#the_focus_management_problem) that I linked above, it seems to point in the right direction...

I look forward to both of your insights/suggestions! :hugs:

frontend/src/component/viewer.vue
```diff
@@ -247,6 +248,25 @@ export default {
       this.idleTimer = false;
       this.hasTouch = false;
 
+      // Add Safari-specific space key handler
+      if (this.isSafari) {
+        lightbox.on('bindEvents', () => {
+          lightbox.pswp.element.addEventListener('keydown', (e) => {
+            if (e.code === 'Space' || e.key === ' ') {
```
AsikNasik: From my point of view it is better to use the `keyCode` to check the necessary element, what do you think? 
In this case we can add the comment to this line, for example: 32 - Space, 13 - Enter, docs - https://www.toptal.com/developers/keycode

lastzero:
I've just added keyboard event handling based on your PR. Also, the zoom button is now hidden and the caption element style is adjusted when a video is playing. 🎬

lastzero:
I'll move ahead and close this PR since the approach has been manually added to the "upgrade/vuetify-3" branch. Thank you very much!

PR #4922, opened 2025-04-03
AI: Improve TensorFlow for C build images and scripts

raystlin:
This pull request fixes create_archive.sh to include symlinks as symlinks and not as copies of the same file. Also the base image for tensorflow compiler docker image is changed to jammy to compile against an older version of glibc.


lastzero:
Thanks a lot! Let me know when you expect no more changes so I can proceed to merge and test it :+1:

lastzero:
@raystlin Since I haven't heard anything to the contrary, I'll go ahead and merge this! Feel free to open a new one if needed.

lastzero:
I've added more build targets and `--copt=-O2` flags as this reduces the size of the binaries:

```
-rw-r--r--  1 root root 113M Apr  6 07:41 libtensorflow-amd64-2.18.0.tar.gz-with-copt-O2
-rw-r--r--  1 root root 131M Apr  6 07:07 libtensorflow-amd64-2.18.0.tar.gz-with-copt-O3
```

PR #4930, opened 2025-04-05
Viewer: Add Interactive Map to Sidebar Info

omerdduran:
This PR adds an interactive MapLibre GL map to the sidebar info component to display location coordinates, along with related CSS styling for layout and map display.


frontend/src/component/sidebar/info.vue
```diff
@@ -91,17 +98,118 @@ export default {
...
+      marker: null,
+      mapLoaded: false,
+      loadingMapLibre: false,
     };
   },
   computed: {
     model() {
       return this.modelValue;
     },
+    megapixels() {
```
lastzero: I don't believe you need this, since I've moved the calculation to the model?

frontend/src/component/sidebar/info.vue
```diff
@@ -91,17 +98,118 @@ export default {
...
+  watch: {
+    "model.Lat"() {
+      this.loadMapAndInit();
+    },
+    "model.Lng"() {
+      this.loadMapAndInit();
+    },
+  },
+  mounted() {
+    if (this.model.Lat && this.model.Lng) {
```
lastzero: There might be a more elegant way than adding two model watchers and calling `loadMapAndInit()` (up to) 3 times? This looks like it could use refactoring :)

lastzero (changes requested):
Please take a look at my comments and try to make the map loading more elegant :)

frontend/src/component/sidebar/info.vue
lastzero: For example, you could create a reusable component for the map (which is only rendered when there are coordinates) and load MapsLibreGL there in `mounted()`.

lastzero (approved):
Thank you very much! This looks good at first sight, so I'll go ahead and merge it :+1:

PR #4931, opened 2025-04-06
Tensorflow 2.18.0 arm64 Dockerfile to Jammy (issue 222)

raystlin:
As stated in issue 222, the Dockerfile for arm64 should be changed to jammy, so as to support more versions of glibc.


lastzero:
Thanks a lot! I'll go ahead and merge this, so we can test it :)

PR #4978, opened 2025-05-01
Setup: Update DigitalOcean Docker configuration

omerdduran:
This PR includes updating the Docker image used on Digital Ocean

### DIFF
--- a/setup/docker/cloud/README.txt
+++ b/setup/docker/cloud/README.txt
@@ -8,7 +8,7 @@ SOFTWARE INCLUDED
 
 PhotoPrism latest, AGPL 3
 Docker CE 23.0, Apache 2
-Traefik 2.9, MIT
+Traefik 3.3, MIT
 MariaDB 10.10, GPL 2
 Ofelia 0.3, MIT
 Watchtower 1.5, Apache 2
--- a/setup/docker/cloud/README.txt
+++ b/setup/docker/cloud/README.txt
@@ -1,6 +1,6 @@
 ========================================================================
 PhotoPrism for Cloud Servers
-Based on Ubuntu 22.04 LTS (Jammy Jellyfish)
+Based on Ubuntu 25.04 LTS (Plucky Puffin)
 ========================================================================
 
 SOFTWARE INCLUDED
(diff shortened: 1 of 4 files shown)


setup/docker/cloud/compose.yaml
```diff
@@ -1,6 +1,6 @@
 #  ========================================================================
 #  PhotoPrism for Cloud Servers
-#  Based on Ubuntu 22.04 LTS (Jammy Jellyfish)
```
lastzero, line 3: Note that the Ubuntu version here refers to the version listed in the `digitalocean.json` file:

```json
"builders": [
    {
      "type": "digitalocean",
      "image": "ubuntu-22-04-x64"
    }
]
```

Since it is not necessarily relevant to this `compose.yaml` file, it may be best to simply remove this information to avoid confusion.

setup/docker/cloud/README.txt
```diff
@@ -1,17 +1,17 @@
...
 ------------------------------------------------------------------------
 
 PhotoPrism latest, AGPL 3
-Docker CE 23.0, Apache 2
-Traefik 2.9, MIT
-MariaDB 10.10, GPL 2
-Ofelia 0.3, MIT
-Watchtower 1.5, Apache 2
+Docker CE latest, Apache 2
+Traefik 3.3, MIT
```
lastzero: Traefik v3.4 has just been released. So it would be good to upgrade it here too before merging these changes.

setup/docker/cloud/README.txt
```diff
@@ -1,17 +1,17 @@
 ========================================================================
 PhotoPrism for Cloud Servers
-Based on Ubuntu 22.04 LTS (Jammy Jellyfish)
+Based on Ubuntu 25.04 LTS (Plucky Puffin)
```
lastzero: As explained in more detail below, the Ubuntu version here refers to the version listed in the `digitalocean.json` file. If we have time, we should update the server base image to the LTS version, in this case 24.04. Only the SD card image and the Docker container image (should) use the non-LTS versions, so at the moment 25.04 (without LTS in the name).

lastzero: Just like in the `compose.yaml` file, it might be best to remove the Ubuntu version information here to avoid confusion!?

lastzero:
@omerdduran Thank you very much! :whale:

@graciousgrey After merging this, we need to remember to update the app image on DigitalOcean.

Issue #127, opened 2019-07-24
Loading a Custom TensorFlow Model


Kazakuri:
It'd be nice to have some sort of documented procedure for loading a different TensorFlow model into PhotoPrism.

I don't have much experience with machine learning so I may be missing the mark with some of my assumptions but I tried to look into it a bit and wrote down my findings here.

--------------------------------------------------------

From my initial investigation it looks like NASNet is [downloaded when building the Docker image](https://github.com/photoprism/photoprism/blob/de1a02694c82843a51bc8f5b91a07fec11f4414c/scripts/download-nasnet.sh), and then the model is [loaded from a static location in the application config](https://github.com/photoprism/photoprism/blob/801097c3688a0d86f293ed872ada0503b64484bf/internal/config/config.go#L411).

By default that's `~/.local/share/photoprism/resources/nasnet` so to replace the TensorFlow model you can just mount that folder with a new asset or otherwise change the `assets-path` in the config file.

From there I start to get a bit lost, when loading the model it seems like it [loads a graph that's tagged with the name `photoprism`](https://github.com/photoprism/photoprism/blob/a995bb87de5c4b9fd529364a6d4a19e2dc49c41c/internal/photoprism/tensorflow.go#L162) and labels from `labels.txt` in the same folder. It might be helpful to make the name of the tag that's loaded part of the configuration to make changing out models easier.

[The documentation](https://godoc.org/github.com/tensorflow/tensorflow/tensorflow/go#LoadSavedModel) I found for `LoadSavedModel` doesn't tell me much, but my assumption is that it's only capable of loading protocol buffer models, so models must be trained and saved as .pb files or somehow converted to a .pb file.

The other thing I've noticed when looking through the code is that [PhotoPrism uses models trained at 224x224](https://github.com/photoprism/photoprism/blob/a995bb87de5c4b9fd529364a6d4a19e2dc49c41c/internal/photoprism/tensorflow.go#L244), so I assume that if a custom model was trained at a different size then those numbers would need to be changed as well.

---------------------------------------------

As a summary of what I've gathered it seems like these are the steps required:

1. Get your model into the correct format (.pb).
    - I'm not sure how possible this is, but maybe provide resources on how to do this for other common model types? Ex. h5, cntk, pth/pkl
2. Update the name of the graph being loaded from the model.
    - Either by updating the source code, or ideally a config file change.
3. Update the resolution that the images are scaled to.
    - Either by updating the source code, or ideally a config file change.
[...]

PR #5011, opened 2025-05-14
AI: Allow configuration of custom classification models

raystlin:
This commit adds the possibility of using custom models for classification via `vision.yml` configuration file.
The code will inspect the models and try to find their variable layout, inputs, number of outputs and so on, however, if the model returns `logits` instead of probabilities it must be specified on the configuration to add a `softmax` layer after the last layer of the model.

**Notes**
* NSFW and Faces can be configured and replaced, but the rest of the code is not really prepared for them.
* Please, take into account that `download-nasnet.sh` script has been modified to include `labels21k.txt` to `nasnet`  path, so as to have a default implementation of 21k labels, although the rules for them have not been written yet.
* Also the test case `internal/ai/classify/model_external_test.go` has been set as optional via an environment var, as it is not an acceptance but one to perform on test photos against external models. 

**Related Issues**

-  #127


scripts/download-nasnet.sh
```diff
@@ -9,6 +9,7 @@ MODEL_ZIP="/tmp/photoprism/nasnet.zip"
 MODEL_HASH="f18b801354e95cade497b4f12e8d2537d04c04f6  $MODEL_ZIP"
 MODEL_VERSION="$MODEL_PATH/version.txt"
 MODEL_BACKUP="storage/backup/nasnet-$TODAY"
+MODEL_21K_LABELS_URL="https://dl.photoprism.app/tensorflow/vision/labels-imagenet21k.txt"
```
lastzero, line 12: It probably makes sense to comment this out (or remove it) until 21k labels are actually supported? As I mentioned in my last issue comment, I believe it would be easier to generate CLIP embeddings and implement a search with those than to generate 21,000 different labels suitable only for regular text-based searches?

lastzero:
@raystlin Thanks a lot! That's awesome :)

After taking a (very) quick look at your changes, I have a few questions regarding their intended scope and functionality:

- It seems like the newly added `PhotoInput` abstracts the input thumbnail configuration, right?
- When you pass an image: Are there any constraints regarding size or formats?
- Should it automatically resize the image to match the configured dimensions?
- If the input image is not square or does not match the configured size or aspect ratio, can/should multiple crops automatically be created to cover the entire area?

Also, I'd like to ask the following just to be sure:

- When you run `make test-go` in a terminal, do all the existing and new unit tests pass? Have you observed any instabilities or performance issues?
- How is the test coverage, especially with regard to the new functionality? Should more tests be added before or after these changes are merged?
- Did you run `make fmt-go` to reformat the source files as needed?

internal/ai/vision/model.go
```diff
@@ -16,15 +17,15 @@ var modelMutex = sync.Mutex{}
...
-	Service       Service   `yaml:"Service,omitempty" json:"Service,omitempty"`
-	Path          string    `yaml:"Path,omitempty" json:"-"`
-	Tags          []string  `yaml:"Tags,omitempty" json:"-"`
-	Disabled      bool      `yaml:"Disabled,omitempty" json:"disabled,omitempty"`
+	Type          ModelType             `yaml:"Type,omitempty" json:"type,omitempty"`
+	Name          string                `yaml:"Name,omitempty" json:"name,omitempty"`
+	Version       string                `yaml:"Version,omitempty" json:"version,omitempty"`
+	Prompt        string                `yaml:"Prompt,omitempty" json:"prompt,omitempty"`
+	Resolution    int                   `yaml:"Resolution,omitempty" json:"resolution,omitempty"`
+	Meta          *tensorflow.ModelInfo `yaml:"Meta,omitempty" json:"meta,omitempty"`
```
lastzero, line 33: Having a `vision.yml` configuration example with your changes applied could help with testing. This would clarify what it should look like and enable us to easily compare it with the existing format and available options.

scripts/download-nasnet.sh
raystlin: Yes, you are right. I modified the script so as to have some kind of classification (although the rules were not being applied), but if we are thinking of alternative ways of doing the ruling, it makes more sense to remove this.

internal/ai/vision/model.go
raystlin: Of course. I missed this one. This is the example I used for the transformer one, but I have to change it on the PR.
```yaml
---
Models:
- Type: labels
  Name: transformer
  Version: Mobile
  Resolution: 224
  Meta:
    Input:
        Interval:
            Start: -1.0
            End: 1.0
    Output:
      OutputsLogits: true
- Type: nsfw
  Name: Nsfw
  Resolution: 224
- Type: face
  Name: FaceNet
  Resolution: 160
- Type: caption
  Resolution: 224
  Service:
    Uri: http://photoprism-vision:5000/api/v1/vision/caption
    FileScheme: https
    RequestFormat: url
    ResponseFormat: vision
Thresholds:
  Confidence: 10
```

raystlin:
- `PhotoInput` abstracts the entry needed by the model: the input layer name, the expected height and width, the expected interval for input values ([0,1] or [-1,1] or whatever), the index of the output and so on. I try to deduce as much as I can from the model deserialization, but those assumptions can be overridden via `vision.yml`. Please, also notice that I left resolution value at Models level, this is because I assumed by the comment https://github.com/photoprism/photoprism/issues/127#issuecomment-2781700031 that this parameter was going to be used to generate thumbnails in the exact size needed by the model. I think I misunderstood it, but we can always do a transformation like the one done on NSFW to solve it.
- Supported file formats are the same ones previously supported: jpg and png (I did not change the image decoding part of the code). 
- Constraints are not enforced. As I previously said, I thought the input was being enforced to its correct size during the thumbnail generation. If this assumption is wrong, I am afraid I have to work on that part.
- `make test-go` runs and passes the tests. I have not seen performance or stability issues.
- I have added the file `internal/ai/classify/model_external_test.go` to test/compare the new model against the test data used for `nasnet`. The test downloads the models (which is really slow), runs them and, sometimes, the labels generated for some tests are not the expected ones. Because of this, I added an enviroment variable as a condition to run them.
- I ran `make fmt-go` but I missed one file: `internal/ai/classify/model_external_test.go`. I will add it to the commit.

lastzero:
@raystlin Sorry for not getting back to you sooner! I was traveling, but am now back in the office. Please let me know if you consider this stable enough for release. I will then merge your changes, perform final testing, and create a new preview build :)

raystlin:
I wanted to add further checking to the input image size (and probably add conversions if the size is not the expected one), but I have been lately quite busy and I left this on my backlog. I will try to have it finished by the end of the week.

lastzero:
@raystlin Thanks for the info! I'll wait for your go. We can then merge and test this after our next stable release, which is planned for next week. :+1:

raystlin:
I am afraid I am having less time to code than I expected, but I promise to have this fully tested and finished just after the next stable release, so that you can have plenty of time to merge and test.

lastzero:
Note that I updated the default configuration to work with the [photoprism-vision](https://github.com/photoprism/photoprism-vision) service.

raystlin:
After reviewing the source code of Keras, I decided to add the following optional parameters to the input tensor definition:
- ResizeOperation: By default Center-Crop was being applied when the input image size did not match the expected input tensor size, now padding and simple resizing (without preserving the aspect ratio) is also possible.
- Channel order can now be defined. By default, RGB is used, but it is configurable.
- Input Intervals can be defined for each input channel. This was done because sometimes the model required having a different per-channel adjustment based on the mean and std dev values of the training data.

After this, only updating the commit to the changes on the upstream remains to be done.

lastzero:
@raystlin I added some new fields to the model struct so that the Ollama API service can be used directly.

raystlin:
Ok. This week I am out, but I will finish this PR during the weekend.

raystlin:
I think this is ready for your testing.

lastzero:
Thank you, I will test it as soon as possible!

lastzero:
@raystlin All the unit tests are green, so that's great!

As a general note, the inline comments should end with a period because they are used to automatically generate API documentation (not a big issue, though, because I can easily update them):

Screenshot from 2025-08-04 09-39-21<br>

For our end user and developer documentation, it would also be helpful to have more information on the **newly added configuration options**, as well as a **list of the models** that you have tested and found to work?

Thank you very much! :hugs:

lastzero:
@raystlin I've merged your changes with a few edits for further testing and documentation. Thank you so much for working on this! :heart: 

It would be good to have some specific configuration and usage examples ready for end users before releasing these improvements in the stable version. @graciousgrey and @omerdduran are happy to help with this.

raystlin:
Sure, I will do it this afternoon. I will also bear it in mind for future
contributions.

lastzero:
You can now test these changes with the updated preview build:

PR #5015, opened 2025-05-19
Fix: timezone handling in sidebar component

omerdduran:
The backend stores the TakenAtLocal value as the local time of the specified timezone, like Berlin's time for Europe/Berlin. If we interpret TakenAtLocal with the timezone (DateTime.fromISO(time, {zone: timezone})), it results in a double time offset. To fix this, we should interpret TakenAtLocal as UTC to avoid time shifts, add timezone information only for display (using keepLocalTime: true), and thus preserve the correct local time while showing the timezone.

Related Issues:

- https://github.com/photoprism/photoprism/issues/5018

### DIFF
--- a/frontend/src/component/sidebar/info.vue
+++ b/frontend/src/component/sidebar/info.vue
@@ -120,10 +120,14 @@ export default {
         return this.$gettext("Unknown");
       }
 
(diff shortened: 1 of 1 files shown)


lastzero:
Thank you very much! I'll merge this for further testing :+1:

PR #5022, opened 2025-05-23
Frontend: Update Chinese translation 

equals-chan:
Fixed some ambiguous translations.


lastzero:
Thank you very much! We appreciate your contribution. It would make things easier for us, however, if you could edit the relevant strings directly at https://translate.photoprism.app/projects/photoprism/, as described in the developer guide at https://docs.photoprism.app/developer-guide/translations-weblate/.

Please let us know if this works for you and whether you need any help! :)

equals-chan:
Ok, I'll edit it right on the page, thank you!

PR #5043, opened 2025-06-02
UX: Add a map for selecting locations in the Edit dialog

omerdduran:
This PR adds an interactive map to make setting photo locations easier. Clicking the new map button next to the latitude/longitude fields opens a dialog where users can click or drag a marker to set the location. Manual coordinate entry is still supported, as is reverse geocoding, which shows location details.

Related Issues:

- https://github.com/photoprism/photoprism/issues/465

Issue #465, opened 2020-08-14
Places: Set geolocation on map or from saved places


Linkandzelda:
As a user, I want to be able to quickly add a geotag to a photo that doesn't have one. Ideally, I could either open a map and select the location directly or view a list of saved locations consisting of latitude and longitude coordinates with a dropdown menu for quickly selecting the desired location and updating the photo's metadata.

![Image](https://github.com/user-attachments/assets/0e01f097-6b1c-4e91-b7f0-2d21ec396d85)

lastzero:
We'll certainly add that after our initial release.

jschmaltz69:
Further to this, the ability to batch edit multiple photos to the same approximate location would also be helpful. It also seems PhotoPrism does not read locations assigned to a photo's metadata by digiKam (or at least, I have not been able to make it work thus far).

rickysarraf:
It does. I've been fixing gps coordinates that way for weeks now.

I'll try to share a screenshot once I'm at my computer

s3nt fr0m a $martph0ne, excuse typ0s

Qix-:
# ⚠️  IssueHunt is abandoned.

If you are putting money into it, **you are losing it**. Nobody will be able to receive that money as IssueHunt is not responding to withdrawal requests.

Just a heads up. It's best to just forget about it.

andyvalerio:
Hello folks, 
not sure what's the status of this but I created a Chrome extension that allows to select the location of the photos from Google Maps. 
I did it mostly for myself but I published it on the Chrome Web Store as well under the name "[Geolocation Plugin for PhotoPrism](https://chrome.google.com/webstore/detail/geolocation-plugin-for-ph/oggmpodnbdcmfiognbkkeffacpeaifch)". 

While we wait for a more complete solution in PhotoPrism you're very welcome to use this, at least in Chrome. 😀

Linkandzelda:
Amazing plugin! Thanks so much @andyvalerio . The plugin is good enough that the system could be merged into PhotoPrism itself.

andyvalerio:
> Amazing plugin! Thanks so much @andyvalerio . The plugin is good enough that the system could be merged into PhotoPrism itself.

Thanks! It's nothing complicated but it gets the job done. It should be easy to add a list of "saved places" or "favorite places" as well but I really didn't have time to implement it for this first version, I might give it a try in a few weeks when I come back from vacation.

mtojay:
[...]

PR #5069, opened 2025-06-25
UX: Improve GPS position input and dialog components

omerdduran:
The changes made here include the improvements mentioned in this comment 

> - [X] Use a single input field for latitude and longitude in the photo edit dialog, similar to the one in the new selection dialog.
> - [X] Use a flex/grid-based layout for the form instead of v-card components, similar to other existing dialogs.
> - [X] This way, it should be easy to position the Cancel and Apply buttons at the bottom of the right column in desktop browsers, as in other dialogs.
> - [X] Currently, our other dialogs do not have a close button in the upper-right corner when using a desktop browser. For now, I would remove it unless we decide to implement it everywhere, which could require a lot more work.

Original comment https://github.com/photoprism/photoprism/issues/465#issuecomment-3002430996

#### Related Issues and PR

- #465 
- #5043

### DIFF
--- a/frontend/src/component/photo/edit/details.vue
+++ b/frontend/src/component/photo/edit/details.vue
@@ -710,7 +710,7 @@ export default {
       const lng = this.view.model.Lng;
 
       if (lat !== null && lng !== null && !(lat === 0 && lng === 0) && !isNaN(lat) && !isNaN(lng)) {
-        this.coordinateInput = `${parseFloat(lat).toFixed(6)}, ${parseFloat(lng).toFixed(6)}`;
+        this.coordinateInput = `${parseFloat(lat)}, ${parseFloat(lng)}`;
       } else {
         this.coordinateInput = "";
       }
--- a/frontend/src/component/photo/edit/details.vue
+++ b/frontend/src/component/photo/edit/details.vue
@@ -187,45 +187,24 @@
               ></v-text-field>
             </v-col>
             <v-col cols="12" md="6">
-              <v-text-field
-                v-model="coordinateInput"
+              <p-coordinate-input
+                :latitude="view.model.Lat"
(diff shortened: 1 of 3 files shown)


frontend/src/component/photo/edit/details.vue
```diff
@@ -671,6 +680,41 @@ export default {
...
+      this.view.model.Lng = 0;
+      this.view.model.PlaceSrc = "manual";
+      this.updateCoordinateInput();
+    },
+    updateCoordinateInput() {
+      const lat = this.view.model.Lat;
+      const lng = this.view.model.Lng;
+
+      if (lat !== null && lng !== null && !(lat === 0 && lng === 0) && !isNaN(lat) && !isNaN(lng)) {
+        this.coordinateInput = `${parseFloat(lat).toFixed(6)}, ${parseFloat(lng).toFixed(6)}`;
```
lastzero: Note that we need to be careful with rounding because it can cause pictures to be placed incorrectly on the map, depending on the specific latitude and longitude. Shortening the coordinates in the input must not affect the data sent to the backend. In any case, the original data should be preserved if no changes are made?

lastzero:
@omerdduran I'm happy to merge your PR and perform detailed testing — not just a quick look at the code — when you're done with your changes!

Please test the layout on desktop and mobile devices to ensure the look and feel are similar to existing dialogs, e.g. those under Settings > Account. It would also be great to have the position input as a reusable component :)

omerdduran:
I made some minor changes to the CSS. Now it looks just the way I wanted. You can test it and merge it. Whenever you have some free time @lastzero

omerdduran:
Since this component will be used again in the future for the batch edit feature, it makes more sense to move all the features to a separate component. I created a new component called "coordinate-input.vue."

PR #5097, opened 2025-07-07
Places: refactor location dialog to reuse PMap component

omerdduran:
## Summary
This pull request introduces new features to the PMap component. It adds new props, including `interactive`, `draggable`, `showControls`, and `clickable`. Additionally, new events such as `marker-moved`, `map-clicked`, `update:lat`, and `update:lng` have been implemented. To enhance user experience, new methods like `removeMarker()` and `flyTo()` have also been introduced. All these changes have been made while maintaining backward compatibility for existing PMap usage.

### DIFF
--- a/frontend/src/component/location/dialog.vue
+++ b/frontend/src/component/location/dialog.vue
@@ -29,7 +29,7 @@
           <div class="flex-grow-1 position-relative mb-4 mb-md-0">
             <p-map
               ref="map"
-              :coordinates="[currentLat, currentLng]"
+              :latlng="[currentLat, currentLng]"
               :zoom="12"
               :style="style"
               :interactive="true"
@@ -103,14 +103,12 @@
 
             <div class="flex-grow-1">
               <p-location-input
-                :lat="currentLat"
-                :lng="currentLng"
+                :latlng="[currentLat, currentLng]"
                 density="comfortable"
                 :enable-undo="true"
                 :auto-apply="true"
                 :label="locationLabel"
-                @update:lat="currentLat = $event"
-                @update:lng="currentLng = $event"
+                @update:latlng="onLatLngUpdate"
                 @changed="onLocationChanged"
                 @cleared="onLocationCleared"
               ></p-location-input>
@@ -152,7 +150,7 @@ export default {
       type: Boolean,
       default: false,
     },
-    coordinates: {
+    latlng: {
       type: Array,
       default: () => [0, 0],
     },
@@ -161,11 +159,11 @@ export default {
       default: "embedded",
     },
   },
-  emits: ["update:coordinates", "close", "confirm"],
+  emits: ["update:latlng", "close", "confirm"],
   data() {
     return {
-      currentLat: this.coordinates[0],
-      currentLng: this.coordinates[1],
+      currentLat: this.latlng[0],
+      currentLng: this.latlng[1],
       location: null,
       locationLoading: false,
       searchQuery: "",
@@ -187,8 +185,8 @@ export default {
   watch: {
     visible(show) {
       if (show) {
-        this.currentLat = this.coordinates[0];
-        this.currentLng = this.coordinates[1];
+        this.currentLat = this.latlng[0];
+        this.currentLng = this.latlng[1];
       }
     },
   },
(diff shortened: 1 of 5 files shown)


frontend/src/component/location/dialog.vue
```diff
@@ -128,21 +139,20 @@
...
     PLocationInput,
+    PMap,
   },
   props: {
     visible: {
       type: Boolean,
       default: false,
     },
-    latlng: {
+    coordinates: {
```
lastzero: I suggest naming it "latlng", as this is not only shorter, but also indicates the order of the values it contains.

frontend/src/component/location/dialog.vue
```diff
@@ -98,8 +109,8 @@
                 :enable-undo="true"
                 :auto-apply="true"
                 :label="locationLabel"
-                @update:lat="setLat"
-                @update:lng="setLng"
+                @update:lat="currentLat = $event"
```
lastzero: Would it be possible to use a single array value here as well, so that we only need one event handler?

frontend/src/component/photo/edit/details.vue
```diff
@@ -417,7 +417,7 @@
     </v-form>
     <p-location-dialog
       :visible="locationDialog"
-      :latlng="[view.model.Lat ? Number(view.model.Lat) : 0, view.model.Lng ? Number(view.model.Lng) : 0]"
+      :coordinates="[view.model.Lat ? Number(view.model.Lat) : 0, view.model.Lng ? Number(view.model.Lng) : 0]"
```
lastzero: As mentioned above, it would be better to name the parameter `latlng`, as this is shorter and also indicates the type and order of the values.

frontend/src/component/location/input.vue
```diff
@@ -140,11 +137,11 @@ export default {
...
-    lat() {
-      this.updateCoordinateInput();
-    },
-    lng() {
-      this.updateCoordinateInput();
+    latlng: {
+      handler() {
+        this.updateCoordinateInput();
+      },
+      deep: true,
```
lastzero: Have you checked whether the deep flag is strictly needed here? I was hoping that the Array Proxy would automatically trigger the event when one of the values changed, but of course that may not be the case...

frontend/src/component/map.vue
```diff
@@ -96,33 +128,90 @@ export default {
...
+    removeMarker() {
+      if (this.marker) {
+        this.marker.remove();
+        this.marker = null;
+      }
+    },
+    // Public method to fly to coordinates
+    flyTo(lat, lng, zoom = this.zoom) {
+      if (this.map) {
+        this.map.flyTo({
```
lastzero: Does this restore the fly animation, when the coordinates change? I replaced it with a simple position change in the version released on Monday, as I couldn't find a way to speed up the animation in the time available:

- Could you please check which map API parameters or options can be used to render the fly animation more quickly, particularly for locations that are very far away? Ideally, the animation speed would depend on the value in `settings.maps.animate`, as shown below.

- When map animations are disabled under *Settings > General*, no animation should be used at all:

https://github.com/photoprism/photoprism/blob/9ad5a533af81f9429643baecea025713357a3b88/frontend/src/page/settings/general.vue#L400-L403

**That should be all that's left to do/check before we can merge this. Thank you very much!** :hugs:

frontend/src/component/sidebar/info.vue
```diff
@@ -66,7 +66,7 @@
           >
           </v-list-item>
           <v-list-item v-if="featPlaces" class="mx-0 px-0">
-            <p-map :lat="model.Lat" :lng="model.Lng"></p-map>
+            <p-map :latlng="[model.Lat, model.Lng]"></p-map>
```
lastzero, line 69: Excellent! That looks much better now :)

omerdduran:
Now the location selection component respects to the animation speeds in the settings. Also, I didn't need the deep feature, I thought I might need it at first. If you can update the batch edit branch after merging, I can now add this feature to the batch edit dialog. @lastzero

lastzero:
Great! Are your changes complete and tested, so we can merge them?

omerdduran:
Yes, I tested it and everythings look good from my side

PR #5100, opened 2025-07-11
Setup: Update mariadb version

omerdduran:
I tested the this version with Raspberry pi5 and everything works correctly. You can also test with pi4 if you want. The reason for the small change I made to the file creation system is that it used to create an extra, unnecessary folder, but now it doesn't create it anymore.

### DIFF
--- a/setup/nas/raspberry-pi/cloud-init/user-data
+++ b/setup/nas/raspberry-pi/cloud-init/user-data
@@ -71,7 +71,10 @@ runcmd:
   
   # Create mount points for external drives
   - printf "\e[0;35m[SETUP] Creating mount points for external drives...\e[0m\n"
-  - mkdir -p /mnt/{a,b,c,d}
+  - mkdir -p /mnt/a
+  - mkdir -p /mnt/b
+  - mkdir -p /mnt/c
+  - mkdir -p /mnt/d
   - chown -R 1000:1000 /mnt
   - printf "\e[0;35m[SETUP] ✓ External drive mount points created\e[0m\n"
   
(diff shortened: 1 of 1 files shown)


setup/nas/raspberry-pi/cloud-init/user-data
```diff
@@ -71,7 +71,10 @@ runcmd:
   
   # Create mount points for external drives
   - printf "\e[0;35m[SETUP] Creating mount points for external drives...\e[0m\n"
-  - mkdir -p /mnt/{a,b,c,d}
+  - mkdir -p /mnt/a
```
lastzero, line 74: `mkdir -p /mnt/a /mnt/b /mnt/c /mnt/d` should also work. Have you tested whether external disks are automatically mounted so that they can be indexed and browsed?

setup/nas/raspberry-pi/cloud-init/user-data
```diff
@@ -278,11 +281,11 @@ write_files:
 
         mariadb:
           restart: always
-          image: arm64v8/mariadb:10.10
+          image: arm64v8/mariadb:11
           security_opt:
             - seccomp=unconfined
             - apparmor=unconfined
-          command: mysqld --innodb-buffer-pool-size=256M --transaction-isolation=READ-COMMITTED --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci --max-connections=512 --innodb-rollback-on-timeout=OFF --innodb-lock-wait-timeout=120
+          command: mariadbd --innodb-buffer-pool-size=256M --transaction-isolation=READ-COMMITTED --character-set-server=utf8mb4 --collation-server=utf8mb4_unicode_ci --max-connections=512 --innodb-rollback-on-timeout=OFF --innodb-lock-wait-timeout=120
```
lastzero: As shown in our other `compose.yaml` examples, you shouldn't need to include "mariadbd" or "mysqld" in the "command" value, only the command flags:
https://github.com/photoprism/photoprism/blob/1fa7b455a1519fe0c3c28812a1d2c1416a5b35a9/setup/docker/compose.yaml#L132

setup/nas/raspberry-pi/cloud-init/user-data
omerdduran: Yes, I tested it and it works. but maybe we can give a info to use file system compatible with ubuntu

setup/nas/raspberry-pi/cloud-init/user-data
omerdduran: Yes, as you said, there was no need, I deleted it

omerdduran:
I tested again with the final version and everything works @lastzero

PR #5164, opened 2025-08-13
AI: Add TensorFlow model shape detection

raystlin:
This commit allows us to load models that have different inputs than BHWC. The process will try to deduce the input shape from the metadata, but, if it is not able or if a configuration is given, the latest will be used.

Also the allocations during the creation of the tensor have been unified to reduce them.


lastzero:
Thanks! That's awesome! I assume you successfully ran all the unit tests, so I'll go ahead and merge this for manual testing :)

lastzero:
New development preview builds are now available for testing:

PR #5177, opened 2025-08-27
UX: Improve "Add to Album" dialog to support multiple albums

omerdduran:
- Update clipboard handlers to accept single UID or array and apply to all
- Update Add to Album dialog to reuse the upload dialog’s multi-select combobox


lastzero:
Thanks a lot! Have you ensured that the same album cannot be added to the list twice (even if the backend would ignore it)?

lastzero:
It would be great to have this ready by next week :)

PR #5178, opened 2025-08-27
Thumbs: Embed Adobe RGB ICC profile in JPEGs with an InteropIndex tag

akdor1154:
### Description

Many digital cameras can save JPEGs in AdobeRGB. Of these, a decent subset do not embed the full ICC Profile in the saved JPEGs; instead they set the exif tag InteropIndex=R03 as a special value that means "this photo is in AdobeRGB".

Problem: Vips doesn't support that, and browsers (at least Chrome and Firefox, I don't have a Mac on hand to test Safari) don't support that either. The images are treated as sRGB incorrectly in practice as a result - tl/dr my images in Photoprism look washed out.

This PR gets Photoprism to sub in a real ICC profile for such images. This makes Vips deal with them correctly, and makes them display correctly in browsers. While it's not really Photoprism's fault that all these other systems lack support for this tag, Photoprism is in a privileged position where it can just Make Things Work without much hassle.

As an example, here's an image with just the InteropIndex tag:
interop_profile

And here is the same one with a full ICC profile included:
embedded_profile

Known TODOs:
 - Licences! The Adobe-compatible profile I pulled from Debian's icc-profiles-free package, so I believe it's fine to use, but unsure if this needs any more documentation?
 - The test image (incl in this pr) is [CC-BY-SA](https://commons.wikimedia.org/wiki/Melbourne#/media/File:Melbourne_skyline_sor.jpg), will probably add attribution to exif description?
 - Unsure the best way to bundle the ICC profile - I feel like it should go in `assets` but the `colors` `pkg` has no access to stuff in there.  In the meantime I copied the method govips uses, but this looks a bit dodgy.

### Acceptance Criteria

- [ ] New features or enhancements are fully implemented and do not break existing functionality, so that they can be released at any time without requiring additional work
- [ ] Documentation has been / will be updated, especially as it relates to new configuration options or potentially disruptive changes
- [ ] Database-related changes have been successfully tested with SQLite 3 and MariaDB 10.5.12+


lastzero:
Thank you very much! We'll review and merge this as soon as possible :)

internal/thumb/vips.go
```diff
@@ -76,6 +77,33 @@ func Vips(imageName string, imageBuffer []byte, hash, thumbPath string, width, h
 		size = vips.SizeBoth
 	}
 
+	iiFull := img.GetString("exif-ifd4-InteroperabilityIndex")
```
lastzero: Does this "scan" the whole file for the string, or is it a cached lookup, so that no significant slowdown is expected when generating thumbnails?

lastzero: Once that's clarified, I'll be happy to merge this PR :+1:

akdor1154: Poking at vips, it looks to be a hash table lookup: https://github.com/libvips/libvips/blob/master/libvips/iofuncs/header.c#L1374-L1375

The foreach above is going over vips' own hardcoded meta items: https://github.com/libvips/libvips/blob/97be5689fee8ac4a672bdd3be3a689ad5592e862/libvips/iofuncs/header.c#L104

So I think ok?

akdor1154:
@lastzero could I also get you to sanity check if any of the TODOs I hid under the images in the PR desc need my attention? Thanks :)

internal/thumb/vips.go
lastzero: Thanks! This looks plausible, so I'll go ahead and merge the changes. However, it would be good to run more tests afterwards to check for any performance impact or other issues that we haven't thought of yet.

lastzero:
@akdor1154 I didn't notice the TODOs. Thanks for the hint! Here are my thoughts:

- If the resulting image looks good to you, I would trust that this is the right profile for Adobe RGB.
- Would it be possible to pass the ICC profile data directly to Vips without using a temporary file? If a file is necessary, we could place it in a subdirectory of the `assets/` folder, e.g., `assets/profiles/icc` or `assets/static/icc` (this would also make it accessible via HTTP/HTTPS in case web clients need it). I've added an `ICCProfilesPath` variable so you can access the profile from the thumbs package. It would also allow users to update the profile file there if necessary.
- You can either move the example image(s) to the [`assets/examples`](http://github.com/photoprism/photoprism/tree/develop/assets/examples) directory and add a `NOTICE` file along with it for attribution, which would allow tests in other packages to use them, or you can keep them in the package `testdata/` directory and add the `NOTICE` file there. I've added an `ExamplesPath` variable in case you want to use the `assets/examples` directory. If not, we can remove the variable again or add additional package tests that use the existing example images.

I'll wait for you to resolve these issues before merging this PR.

internal/thumb/vips.go
```diff
@@ -76,6 +77,33 @@ func Vips(imageName string, imageBuffer []byte, hash, thumbPath string, width, h
...
+	if iiFull != "" {
+		ii := iiFull[:3]
+		log.Tracef("read exif and got interopindex %s, %s", ii, iiFull)
+		fallbackProfile := ""
+		switch ii {
+		case "P98":
```
lastzero: Should this be "R98" instead of "P98"? See:

- https://www.imagekit.com/IK9VCLHelp/source/controlreference/imagekitcontrol/file/exif/propertyinteroperabilityindex.htm
- https://regex.info/blog/photo-tech/color-spaces-page7

lastzero:
Debian seems to reference the license as `zlib/libpng`:

- https://metadata.ftp-master.debian.org/changelogs//main/i/icc-profiles-free/icc-profiles-free_2.0.1%2Bdfsg-1.1_copyright

To be safe, you could mention or link to the license in a `NOTICE` file located in the same directory as the ICC profiles.

internal/thumb/vips.go
```diff
@@ -76,6 +77,33 @@ func Vips(imageName string, imageBuffer []byte, hash, thumbPath string, width, h
...
+			// we could logically embed an srgb profile in the image here, but
+			// there's no value in doing so; everything assumes srgb anyway.
+		case "R03":
+			// adobe rgb
+			fallbackProfile, err = colors.GetAdobeRGB1998Path()
+			if err != nil {
+				log.Errorf("%s in %s (get fallback profile adobergb)", err, clean.Log(filepath.Base(imageName)))
+			}
+		default:
+			log.Warnf("exif in %s has unknown interop index %s", clean.Log(filepath.Base(imageName)), ii)
```
lastzero: Based on my understanding, there could be values other than P98 and R03, in particular if P98 is a typo and it should be R98 instead. Should a warning really be triggered if that happens?

akdor1154: My intention was to include all known values (which i haven't done.. you picked up my typo, and I need to include THM), but then warn on something else - that way if our assumption is violated, users get something in their logs to point out this violation when their thumbs don't get a profile.

akdor1154:
- used your vars to put the icc var in Assets, thanks! also added NOTICE incl source.
  n.b. I snuck a panic() in a Must* in there as imo a bundled file being unavailable is a bug, right?
- subbed CC-BY-SA into test image exif, left image in testdata as that's where the other thumb tests go
- added some more comments to embedding code and fixed my typo, eep!
- have tightened the switch on InteropIndex tag to include all known cases as per the specs we've found, but kept the warning on unknown values - this ok or would you still like to veto it back to a Trace?

thanks again

internal/thumb/vips.go
lastzero: Does this switch execute for every image and/or thumbnail, or only if certain metadata is set and/or if an ICC profile is not already embedded? If an ICC profile is already embedded, would it be replaced by the one specified there or be removed if no profile is set as for sRGB? Since warnings are logged permanently in the database, I suggest changing the log level to debug; otherwise, if users have millions of affected images, millions of rows could be created in the related database table. Our development environment, where we perform unit and QA testing, has debug logs enabled by default. Therefore, the logs should not go unnoticed in case of general implementation issues.

lastzero:
Thanks a lot! For remaining issues/questions, see my comment above:
- https://github.com/photoprism/photoprism/pull/5178#discussion_r2309217304

I would also suggest using a lowercase name for the ICC profile file and shortening the file name to avoid issues with mixed case and typos. This makes the code more readable.

For example, you could use `adobe_rgb.icc`. To the best of my knowledge, no other version has been released, and the compatible profile should be as good as the original for practical purposes, so I don't think we need to add additional information like "1998" or "compatible" to the file name.

internal/thumb/vips.go
akdor1154: I will change to Debug, ta :)

For reference - the switch executes if the image has an InteroperabilityIndex tag. In that case, an ICC profile will be set if the tag = R03. If a profile already exists in the image, it will be (needlessly) transformed to Adobe RGB by Vips. I wouldn't expect images like this to exist in the wild though - the whole point of InteropIndex is to not need to embed a profile!

akdor1154:
> I would also suggest using a lowercase name for the ICC profile file and shortening the file name to avoid issues with mixed case and typos. This makes the code more readable. For example, you could use adobe_rgb.icc

Ta. I've gone with adobe_rgb_compat.icc because Adobe have this scaremongering copy on their official release:

> Legal note regarding color-space naming: Only the Adobe RGB (1998) ICC profile created by Adobe Systems Incorporated can accurately be referred to as "Adobe RGB (1998)." ICC profiles created by other vendors, even if they conform to the color image encoding described in the Adobe RGB (1998) color image encoding document, cannot be referred to as "Adobe RGB (1998)." If vendors choose to create their own profile according to this specification, and they want to indicate to their customers that this profile was written in accordance with Adobe's specification, then an alternate phrasing is required, such as "compatible with Adobe RGB (1998)."

internal/thumb/vips.go
lastzero: Is it possible to verify if an ICC profile already exists? In this case, we should skip this step and accept the existing profile (if it doesn't match the metadata, then that's a conflict we might not be able to properly resolve). Ideally, this should be done without impacting performance i.e. without first reading the entire existing profile.

akdor1154: i believe Vips already reads the profile to some extent, or at least its existence, as the lookup function just checks if a meta field exists. In any case this makes the check for us cheap. Have done so and pulled all this out to a new func as it's getting a bit long.

lastzero:
Thanks a lot! There are only two minor issues left:

- `MustGetAdobeRGB1998Path()` panics if the ICC profile file is missing. That’s risky in production. It would be better to replace it with `GetAdobeRGBCompatPath() (string, error)` and return early with an error if the file isn’t found.
- Consider adding a test that explicitly checks whether `vipsSetIccProfileForInteropIndex()` exits without overwriting an existing ICC profile.

lastzero:
@akdor1154 Since we haven't heard back from you, I have merged and refined your changes so that we can release them. Thanks again for your contribution!

PR #5188, opened 2025-09-01
Search: Add escaping capability for the characters "|" and "&"

keif888:
### Description

These changes implement the ability to escape the pipe, ampersand and backslash characters in the photo search for caption, filename, name or title fields.

The implementation is as backwards compatible as possible, with detection of the character following the backslash (escape character) to determine if it is valid for escaping or not.  
- Where not valid, no changes are made to the search.
- Where valid, the back slash is removed from the search, and the following character kept, and not treated as a separator.

### Acceptance Criteria

### DIFF
--- a/internal/entity/search/conditions.go
+++ b/internal/entity/search/conditions.go
@@ -33,7 +33,7 @@ func LikeAny(col, s string, keywords, exact bool) (wheres []string) {
 		wildcardThreshold = 2
 	}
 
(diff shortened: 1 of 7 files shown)


lastzero:
Thank you very much! This PR looks good. I might have left the split function in the `search` package since it's pretty specific to searching. However, you could also argue that it's generic enough to move it to the `txt` package.

Should I go ahead and merge it?

PR #5190, opened 2025-09-02
Search: Update Sql string escaping to match different DBMS

keif888:
### Description

These changes fix how PhotoPrism handles strings that are used in LIKE statements so they are escaped appropriately for each DBMS.

MariaDB LIKE strings will be escaped for single quotes and back slashes by duplicating that byte.
All other DBMS LIKE strings will be escaped for single quote by duplicating that byte.

### Acceptance Criteria


pkg/clean/sql.go
```diff
@@ -1,24 +1,40 @@
 package clean
 
+const (
+	MySQL    = "mysql"
```
lastzero: It would seem better to define dialect constants only once, since they are currently also defined in the `internal/entity` and `internal/config` packages. Suggestions?

keif888:
package constants has been renamed to enum, and code refactored.  All unit tests have passed.

lastzero:
Thanks a lot! I'll review and merge your changes as soon as possible. :ok_hand:

lastzero:
Since QA starts today, I applied your fixes and improvements manually yesterday so that we could merge them in time for the upcoming release. 🔧

PR #5191, opened 2025-09-02
Backend: improve Yes/No capabilities

keif888:
### Description

These changes improve the way PhotoPrism determines if a Yes or No has been provided in a query string.
It covers Yes and No for the following languages:
Czech, Danish, Dutch, English, French, German, Indonesian, Italian, Polish, Portuguese, Russian, Ukrainian.

### Acceptance Criteria


pkg/txt/bool.go
```diff
@@ -20,26 +21,32 @@ func Bool(s string) bool {
...
-	if s == "" {
+// Yes tests if a string represents "yes" in the following languages Czech, Danish, Dutch, English, French, German, Indonesian, Italian, Polish, Portuguese, Russian, Ukrainian.
+func Yes(s string) (result bool) {
+	t := strings.ToLower(strings.TrimSpace(s))
+	if t == "" {
 		return false
+	} else if strings.Contains(t, " ") {
+		result = false
+	} else {
+		noStrings := []string{"1", "yes", "include", "true", "positive", "please", "ano", "ja", "oui", "si", "tak", "sim", "Да", "ya", "так"}
```
lastzero: Should `noStrings` be `yesStrings`?

Since these functions are frequently called, we should consider defining this list of strings as a package variable. Then, it won't need to be created in memory each time. Additionally, we could use maps instead of slices for faster lookups:

```go
var yesSet = map[string]struct{}{
  "1": {}, "y": {}, "yes": {}, "true": {}, "include": {}, "positive": {},
  "ano": {}, "ja": {}, "oui": {}, "sí": {}, "si": {}, "tak": {}, "sim": {}, "да": {}, "ya": {}, "так": {},
}

var noSet = map[string]struct{}{
  "0": {}, "n": {}, "no": {}, "false": {}, "none": {}, "exclude": {}, "negative": {},
  "žádný": {}, "nej": {}, "nee": {}, "nein": {}, "non": {}, "nie": {}, "não": {}, "нет": {}, "tidak": {}, "ні": {},
}
```

lastzero:
How about we call the new package "pkg/enum"? I don't think it's a reserved word in Go and it's much less to type than "constants".

keif888:
> How about we call the new package "pkg/enum"? I don't think it's a reserved word in Go and it's much less to type than "constants".

I went with **constants** as that is what I already used for the SQL constants in #5190.
I can change both to **enums** if you would prefer that.
I prefer the plural, as the package will contain many enumerated values.

lastzero:
In Go, the best practice is to always use singular package names even if a package contains multiple items (most do). This way, it's better readable when you access them e.g. "enum.SQLite" or "enum.YesSet". We can make exceptions when there's a naming conflict, e.g. with a reserved word.

keif888:
OK, I will update both PR's to enum from constants.
I will push them when I have updated and retested.

keif888:
package constants has been renamed to enum, and code refactored.  All unit tests have passed.

lastzero:
Thanks a lot! In the meantime, I updated the [Best Practices](https://docs.photoprism.app/developer-guide/code-quality/) page in our **Developer Guide** to include Go naming conventions, as well as new sections on [Concurrency](https://docs.photoprism.app/developer-guide/code-quality/#use-safe-concurrency) and [Caching](https://docs.photoprism.app/developer-guide/code-quality/#be-careful-with-caching):
- https://docs.photoprism.app/developer-guide/code-quality/#naming-things-is-hard

I'll review and merge your changes as soon as possible! :ok_hand:

lastzero:
Thanks a lot! I'll go ahead and merge this for final testing :)

lastzero:
Our release notes have been updated to mention this improvement:

PR #5214, opened 2025-09-16
Backend: Ensure that all user input to SQL goes via parameterised queries

keif888:
These changes improve the way that user input is fed to the backend database to ensure that all queries that are derived from user input use parameterised queries.

### Acceptance Criteria


pkg/clean/sql.go
```diff
@@ -51,3 +51,34 @@ func SqlString(s string) string {
 
 	return string(b[:j])
 }
+
+// SqlClean removes bytes from a string that are determined as requiring omitting
+func SqlClean(s string) string {
+	var i int
+	for i = range len(s) {
```
lastzero, line 58: Is this a typo, or a new syntax introduced in Go 1.25? 🤔

lastzero:
Thanks a lot! I'll be happy to merge this when I have feedback on the range loop in `SqlClean()`, so I can complete the review :)

pkg/clean/sql.go
keif888: Go 1.22 added it.

A "for" statement with a "range" clause iterates through all entries of an array, slice, string or map, values received on a channel, **integer values from zero to an upper limit** [[Go 1.22](https://go.dev/ref/spec#Go_1.22)], or values passed to an iterator function's yield function [[Go 1.23](https://go.dev/ref/spec#Go_1.23)]. For each entry it assigns iteration values to corresponding iteration variables if present and then executes the block.

See https://go.dev/ref/spec#For_range

lastzero:
Since QA starts today, I applied your fixes and improvements manually yesterday so that we could merge them in time for the upcoming release. 🔧

Issue #5227, opened 2025-09-23
Labels: Add support for homophones and homophone-aware lookups


xiaobai6769:
<img width="793" height="178" alt="Image" src="https://github.com/user-attachments/assets/e7ca804c-8668-4a46-be4c-5d99ed61b99e" />

I can't add two labels like the one in the picture.

graciousgrey:
Thanks for the report! 

To process this, we need the information requested in our [bug report template](https://github.com/photoprism/photoprism/blob/develop/.github/ISSUE_TEMPLATE/bug_report.md) (e.g. steps to reproduce, expected behaviour, ....).

keif888:
# 1. What is not working as documented?
When attempting to add 2 labels that are homophones in Chinese, the 2nd label is not added.
https://docs.photoprism.app/user-guide/organize/labels/

# 2. How can we reproduce it

Examples of words that cause problems can be found here:  https://ltl-beijing.com/chinese-homophones/

To reproduce:
1. Go to https://demo.photoprism.app/library/browse
2. Click on any photo
3. Click on Edit
4. Click on the Labels Tab
5. Add a new label which uses non English characters, and click the + symbol
6. Add another new label which uses non English characters, but is a homophone, and click the + symbol
7. The 2nd label is not added.

Examples tested to not work:
问 and 吻
上海 and 伤害
是 and 十
老板 and 老伴
# 3. What behavior do you expect?
Each of these labels should be kept against the photo.

# 6. Which software versions do you use?
Demo site Build [250930-bd1be3346](https://docs.photoprism.app/release-notes/)
Chrome 140.0.7339.186

lastzero:
@xiaobai6769 I implemented the improvements for homophones based on @keif888's analysis and pull request:
- https://docs.photoprism.app/release-notes/#development-preview
- https://docs.photoprism.app/getting-started/updates/#development-preview

Please let us know if the preview build works for you! 🤗

PR #5238, opened 2025-10-03
Add capability for homophone labels to be detected and treated as separate labels

keif888:
### Description

These changes enable homophone handling for labels.
Homophones, for this purpose, are two "words" that are spelt differently in a character set other than the Latin character set which when transliterated to the Latin character set result in the same Latin character set word(s).
There are many of these in Chinese, and are possible in 23 (or more) other languages.
The following examples are from Chinese and Greek:
| Latin Homophone | Value 1 | Value 2 |
|-----------------|---------|---------|
| Shang Hai       |    上海     |   伤害      |
| Lao Shi            |    老师     |   牢实      |
| Bazo | Βάζο | Βάζω |

The PR has tests which cover the front end, add label api, search, geo search, entity and pkg level tests.
It adds a new column to the labels table, called max_homophone, which for homophones stores the maximum breaker character on all the records that make up the homophone set.
Homophone labels, when detected, will have their slug updated to be suffixed with -c- and then a character up to the max_homophone value, incremented for each new homophone detected.
eg.
| label_name | label_slug | max_homophone |
|--------------|------------|------------|
| 上海 | shang-hai | a |
| 伤害 | shang-hai-c-a | a |
| Βάζω | bazo | a |
| Βάζο | bazo-c-a | a |

#### Changes to behaviour of the application
- When a user adds a new label which is a homophone, the label will now be created and saved.
- When a user searches for a label in a query, or via a form field as a name, if the name's slug finds a homophone then the name will be compared to all the name values returned on the associated homophone records, and only the name based matching label will be returned

#### Changes to the code base
It adds a new file fixture, a new photo fixture and 2 new label fixtures to support testing.  Some non homophone tests have been updated as the new photo fixture increased the number of photos found in those tests.

It adds a new clean function LabelName and associated tests which cleans the label name to allow comparison in a similar way to slug generation.

It updates the FirstOrCreateLabel function to detect a new homophone, and add it as a new label maintaining the max_homophone column up to a value of z (allows a maximum of 27 words as a single homophone set).

It adds a new entity level function FindLabels and associated tests which will return the list of labels that match a set of names.  It understands homophones, so if you provide one it will only return that one, and not any of the other ones (unless they are included in the search criteria).  eg. If you search for 上海, then 伤害 will not be returned.  This is different from the current behaviour.
This function is used by the Labels, Photos and PhotosGeo search functions.

It updates the Labels, Photos and PhotosGeo search functions to utilise FindLabels for consistent handling of Labels (and reduce code duplication).  Associated tests are included.

#### Related Issues

This fixes issue #5227 

### Acceptance Criteria

- [X] New features or enhancements are fully implemented and do not break existing functionality, so that they can be released at any time without requiring additional work
- [X] Automated unit and/or acceptance tests are included to ensure that changes work as expected and to reduce repetitive manual work
- [X] The user interface has been tested on Chrome, and no changes have been made to it
- [X] Database-related changes have been successfully tested with SQLite 3 and MariaDB 10.5.12+


keif888:
I will return this to review status when the following are fixed and unit tests exist:

- [X] Renamed homophone labels unable to be found
- [X] Renamed labels cause next instance of non-renamed to generate another instance
- [X] and vs & results in homophone but needs to match existing functionality
- [X] + vs - results in homophone but needs to match existing functionality

lastzero:
@keif888 Thanks a lot! I implemented the improvements for homophones based on your analysis and pull request.

keif888:
Close, as implemented.

PR #5298, opened 2025-10-29
Speed acceptance testing up

keif888:
These changes fix an unreported issue with how notifications are timed out by the UI, and improves the performance of the acceptance testing suite that is using testcafe.

**Notifications**
The issue with notifications in the UI is that when there are multiple queued event messages, or when an event message is received just after one is closed, if the previous message's timer has not expired, then it will fire to move to the next message.
This results in the time that a notification is shown being reduced from the configured time, to the remaining time of the previous message's timer.  This can lead to messages being shown for a time that the eye can not see.

**Testing Execution Time**
The execution time of the acceptance testing suite is improved from around 1h 50m to less than 1 hour on my development setup. (Virtual Box VM which is running Debian 12, 8Gb RAM, 4 vCPU on a Windows 11 host with SSD backed storage and Ryzen 7 5700X CPU).

This is achieved by four distinct test changes, and a few order of execution changes in the tests.
1. Modifying the way that the navigation is opened so that it will be fast to detect that nothing is needed
2. Detecting the notification messages that test actions cause to detect the completion of an action, rather than using a static t.wait
3. Update getPhotoCount to use waitForPhotosToLoad to ensure that the photos are ready to count
    1. This assumes that it's called straight after an action that will trigger a notification about number of photos found which is used only for completion of load detection
5. Update search to close all notifications before searching, and the use waitForSearchToFinish
6. Change the order of execution to count photos before getting UID's to ensure that the notifications are still there

The navigation opening improved from on average 6.5s per call to 1.1s when needing to open, and 0.04s when nothing is needed, with ~250 calls in the suite.

There are ~250 waits in the suite, not all of which can be removed, but those that are were averaging 6.6s per call.  This has been reduced to less than 1s on average across all the types of notification calls.
The old wait time has been used as the parameter into the replacement function calls, so that the total elapsed time of a call should be no longer than preceding if an event message isn't received by the UI.

**Makefile**
The Makefile has been changed such that the acceptance tests can be run with ```make -i acceptance-run-chromium``` and it will run all tests even when the Multi-Window tests fail due to a testcafe issue.
The call to the Multi-Window tests has been changed so that each Multi-Window test is run in a separate instance of testcafe to prevent a testcafe hang.

### How it works

A new page-model has been added for the notifications.
This contains 2 main action functions.

1. closeAllEventPopups
    1. This will close all event popups that are shown on the screen at the time of it being called.
2. waitForSpecificEvent
    1. This will close all event popups that are NOT the one that is being waited for (via event parameter), and wait for the delay for it to appear.  
    2. It can optionally close the event message.

There are a set of specific wait functions to cover the scenarios that have been fixed in the code.  These calls are (hopefully) named in such a way that what they are for is self documenting:

- waitForFileDeleted
- waitForFoldersToLoad
- waitForImport
- waitForIndexing
- waitForPeopleToLoad
- waitForPersonCoverUpdate
- waitForPhotosToLoad
- waitForSearchToFinish
- waitForUnstack
- waitForUpload
- waitForUploadFailed

There is also support for an environment variable SHOW_LOGS=true which will turn on timing and debugging messages for these functions.

An example of a change in a test is as follows:
Old:
```
 await t
    .click(photoedit.toggleExpandFile.nth(1))
    .click(Selector(photoedit.deleteFile))
    .click(Selector(".action-confirm"))
    .wait(10000);
```
New:
```
  await t
    .click(photoedit.toggleExpandFile.nth(1))
    .click(Selector(photoedit.deleteFile))
    .click(Selector(".action-confirm"));
  await notifies.waitForFileDeleted();
```
This usually takes 1.5s when I was testing it, but with the default delay of 10000 in waitForFileDeleted it could still wait the same amount of time as before.

### Acceptance Criteria

- [X] New features or enhancements are fully implemented and do not break existing functionality, so that they can be released at any time without requiring additional work
- [X] Automated unit and/or acceptance tests are included to ensure that changes work as expected and to reduce repetitive manual work


lastzero:
Thanks a lot! I'm going to merge this for further testing in `develop`.

Issue #5145, opened 2025-08-04
People: disappearing confirmation dialog on mobile when adding new person


tilusnet:
#### 1. What is not working as documented?

Try adding a name to an unrecognised face. Recently Phoroprism will ask to confim the addition.
On mobile, this dialog often is not visible or slides out of view - it's basically unusable.

#### 2. How can we reproduce it?

See section 1.

#### 3. What behavior do you expect?

Honestly, just get rid of that silly dialog.
The old 2024 version worked just fine.

#### 6. Which software versions do you use?

Version 7 July 2025

graciousgrey:
Thank you for your report. 

In order for us to reproduce this can you please let us know on which device and with what browser your observe that the dialog is not visible or slides out of view?

tilusnet:
Thanks @graciousgrey 

It's an Android Pixel 8, on Chrome.

graciousgrey:
Unfortunately I could not reproduce this so far on our Android devices. 

Which Android and Chrome versions do you use? Do you have browser plugins installed? Can you reproduce this on our public demo?

lastzero:
@tilusnet @graciousgrey This might only happen if a certain number of faces are found, and it probably also depends on the specific person being added (top, middle, or bottom)? Would be good to see a screen recording in this case. Another possible cause could indeed be a plugin, like an ad blocker, that interferes with the user interface.

If it's working on our test devices, the issue might also be resolved after updating to the latest operating system and browser versions if the phone doesn't have all the updates installed...

tilusnet:
Hey guys thanks for the replies.

I can reproduce it easily.
I have dozens of unrecognised faces.

I uploaded a screen cast here:

https://filebin.net/y6zur9zbc99xed8v

My Chrome browser is vanilla, updating regularly.
I don't know how to add plugins so I guess that's not an issue.

<img width="1080" height="964" alt="Image" src="https://github.com/user-attachments/assets/460590c6-843b-4b5b-9fdd-5923d1ac60b7" />

tilusnet:
I should also add that as a worrying side effect is that while the user believes that the name has been added, this does NOT HAPPEN.
Assume because the confirmation dialog was not acknowledged?

Out of interest why did you guys complicate your lives with this really useless dialog?

lastzero:
@tilusnet Can you confirm this is still happening with our latest development preview build or on our [public demo](https://demo.photoprism.app/) (based on the preview)?

[...]

PR #5307, opened 2025-11-04
UX: Improve people name editing and confirm dialog focus management

omerdduran:
These changes improve the reliability and user experience when editing the name of a person/face in the photo editing dialog and on the new People page:
- A more robust confirmation process for renaming or assigning people is introduced, ensuring that only one menu is open at a time while also improving keyboard and menu interactions.
- Additionally, the view helper now automatically traps focus in dialog overlays.

#### Related Issues

- https://github.com/photoprism/photoprism/issues/5145
- https://github.com/photoprism/photoprism/issues/5313


lastzero:
@omerdduran Thanks for pushing this forward!

I added a focus trap to the confirmation dialog and adjusted the menu properties to use an invisible scrim layer in the background. Theresa said that with these changes, everything seems to be working, but more testing is needed after the changes are merged into the `develop` branch.

I don't think the class `.text-selectable` is necessary for `<v-combobox>`, as text in inputs should be selectable by default. Could you do some additional testing? If you find that the class makes no difference, simply remove it.

Issue #4951, opened 2025-04-18
SQLite: Cannot use Google OIDC due to dynamic typing


jbortkiewicz:
#### 1. What is not working as documented?

On PhotoPrism configured with SQLite 3 database and Google OIDC authentication, once an user account is created, then it cannot login second time.

#### 2. How can we reproduce it?

1. Use default [docker compose file](https://dl.photoprism.app/docker/compose.yaml)
2. Configure OIDC by setting environment variables:
```
      PHOTOPRISM_SITE_URL: "https://your.public.domain" 
      PHOTOPRISM_OIDC_URI: "https://accounts.google.com"
      PHOTOPRISM_OIDC_CLIENT: ********"
      PHOTOPRISM_OIDC_SECRET: "*********"
      PHOTOPRISM_OIDC_PROVIDER: "Google"
      PHOTOPRISM_OIDC_REGISTER: "true"
``` 
3. Configure SQLite database by setting environment variables:
```
      PHOTOPRISM_DATABASE_DRIVER: "sqlite"
``` 
4. Start the container: `docker compose up photoprism`
5. Open a browser and navigate to configured public PhotoPrism address
6. Login with Google account - login is successful and account is created
7. Log out 
8. Login with Google account - cannot login, error message "Invalid credentials" is shown

#### 3. What behavior do you expect?

User can login with the Google account. The same Google OIDC configuration works properly with MariaDB database.

#### 4. What could be the cause of your problem?

Login status shows an error related to the auth id:
```
> photoprism audit logins
|----------------|----------------------|-------|----------------------------------|---------------------|---------------------|
|   Client IP    |       Username       | Realm |              Status              |     Last Login      |      Failed At      |
|----------------|----------------------|-------|----------------------------------|---------------------|---------------------|
| 192.168.80.154 | john.doe             | oidc  | error: invalid auth id (3 times) | 2025-04-18 09:52:34 | 2025-04-18 10:04:57 |
|----------------|----------------------|-------|----------------------------------|---------------------|---------------------|
```

The login error is caused by the fact that `auth_id` is stored incorrectly on Sqlite3 database:
```
> photoprism users show
|---------------|--------------------------------------------------|
|     Name      |                      Value                       |
|---------------|--------------------------------------------------|
| AuthID        | "1.0047673928577674e+20"                         |
| AuthIssuer    | "https://accounts.google.com"                    |
| AuthMethod    | ""                                               |
| AuthProvider  | "oidc"                                           |
...
```

[...]

PR #5309, opened 2025-11-06
OIDC: Add handling to AuthID so that SQLite doesn't corrupt on save

keif888:
### Description

These changes alter the data type for the AuthID field outside of GORM's control for the Users struct so that when saving to SQLite it doesn't get corrupted when the value is just a number that exceeds 20 digits in length.

Please note that GORM v1 checks for column's existence but not data type.

The fields have to be annotated in GORM v2 so that they are ignored.
With pre-creation statements for MariaDB and Postgres added (see MariaDB commented out code).

#### Related Issues

- #4951 

### Acceptance Criteria


lastzero:
To ensure SQLite compatibility, how about we try this instead?

1. Make the column `TEXT` with binary collation
  ```sql
  CREATE TABLE auth_sessions (
  ...
  auth_id TEXT NOT NULL COLLATE BINARY,
  ...
  ) STRICT;  -- 
  ```
  - `TEXT` (or `STRICT` tables) prevents numeric coercion.
  - `COLLATE BINARY` ensures case-sensitive, byte-exact comparisons.
  - Requires SQLite 3.37+, so it should work since we use 3.50.4.
2. Bind the `auth_id` as a `TEXT` string when inserting/updating it:
  ```sql
  INSERT INTO auth_sessions (auth_id) VALUES (CAST(? AS TEXT));
  ```
3. Add a database migration, to update/fix existing SQLite schemas.

keif888:
> To ensure SQLite compatibility, how about we try this instead?
> 
> 1. Make the column `TEXT` with binary collation
> 
> ```sql
> CREATE TABLE auth_sessions (
> ...
> auth_id TEXT NOT NULL COLLATE BINARY,
> ...
> ) STRICT;  -- 
> ```
> 
> * `TEXT` (or `STRICT` tables) prevents numeric coercion.
> * `COLLATE BINARY` ensures case-sensitive, byte-exact comparisons.
> * Requires SQLite 3.37+, so it should work since we use 3.50.4.
> 
> 2. Bind the `auth_id` as a `TEXT` string when inserting/updating it:
> 
> ```sql
> INSERT INTO auth_sessions (auth_id) VALUES (CAST(? AS TEXT));
> ```
> 
> 3. Add a database migration, to update/fix existing SQLite schemas.

Gorm AutoMigrate will try and set the column back to the expected type in Gorm V1.
Gorm V2 can have a tag that tells it to not apply AutoMigrate to that column, so maybe possible there.
Can not use the STRICT setting as it breaks Gorm's ability to move the data into the struct (because STRICT doesn't support most of the required types).
I will experiment on the Postgres branch and see if Gorm will allow interaction with the column in the "wrong" type.

Will have to handle the 2 tables that use AuthID special in post phase of the migration process.

keif888:
The latest commit has the non text wrapping solution in it.
Both solutions are still in the PR, whilst I await feedback on the preferred direction.

I have tested both methods (wrap vs force table creation with data type change) against Gorm V2, and they function correctly.

keif888:
This no longer uses a wrap/unwrap of the data in the column.  It's forcing the column datatype in SQLite so that it doesn't do the auto conversion to a number.
It does it by pre-creating the 2 tables with AUTHID columns in a new database, and then letting the migrator come along and adjust anything else (indexes, new columns) that need to be done.
For an existing database, it migrates the data to a new table with the required AUTHID column definition, and then copies the data across, renaming the tables so that the correct structured table will be left with the right name.  The original tables are left behind as migrate_tablename.

internal/entity/auth_user.go
```diff
@@ -627,10 +627,10 @@ func (m *User) SetMethod(method authn.MethodType) *User {
 // SetAuthID sets a custom authentication identifier.
 func (m *User) SetAuthID(id, issuer string) *User {
 	// Update auth id if not empty.
-	if authId := clean.Auth(id); authId == "" {
+	if authID := clean.Auth(id); authID == "" {
```
lastzero: Since this is a local variable, the casing isn't critical, but I suggest keeping `authId` because most code follows that convention. This improves readability when there's a suffix, like `authIdMutex`, and avoids confusion with the struct/exported field, `AuthID`.

internal/entity/migrate/dbms_authid_migration.go
```diff
@@ -0,0 +1,169 @@
...
+			}
+			for _, index := range indexes {
+				dropStatement := fmt.Sprintf(`DROP INDEX IF EXISTS "%s"`, index.Name)
+				if err := db.Exec(dropStatement).Error; err != nil {
+					return fmt.Errorf("migrate: error dropping index %s was %w", index.Name, err)
+				}
+			}
+		}
+	case MySQL: // MySQL
+		// These create statements will get out of date, but that is ok, as the main migrate path will add any missing columns/indexes in later.
```
lastzero: What is the reasoning behind the MariaDB/MySQL migrations? How do the tables/schema created automatically by GORM differ? 🤔

keif888: Oops, I built this for GormV2, and retrofitted it.
It's not needed for GormV1.  I'll fix.

In GormV2 the AuthID columns will be annotated with a tag of ```-:migration``` in the gorm tags, which stops the migrator from dealing with this field.  It is needed because Gorm V2 compares the column types and alters the table to match the struct (which it wont in SQLite's case).  That is why there is a create statement for it in MySQL (and the commented out Postgres lines).

internal/entity/auth_user.go
keif888: I have revive as my linter and it was complaining with "var authId should be authID".

lastzero:
It would be great if you could implement the following improvements, so we can merge this:
1. It seems the temporary tables are not removed after successful copy and verification, i.e. add `DROP TABLE migrate_auth_sessions` and `DROP TABLE migrate_auth_users` (or rename back on failure).
2. Would it be possible to wrap this in a transaction so that the tables are not left in a mixed state if it fails?

keif888:
Transaction initiated at start of migration, uses defer to commit/rollback via err variable.
Drop statements added.  I had left them behind with the intention of removing them in a later release in an abundance of caution.

Manual testing with deliberate bad sql statement in code was done to confirm that the rollback works (with bad statement then removed).  There is no easy way to test a dbms level failure path.
All unit tests passed.

lastzero:
Thanks a lot! I'm testing it again and will merge it if it works.

internal/entity/migrate/dbms_authid_migration.go
```diff
@@ -0,0 +1,196 @@
...
+			DfltValue string
+			Pk        int
+		}
+
+		var err error
+		// Start a transaction
+		tx := db.Begin()
+
+		if tx.Error != nil {
+			return fmt.Errorf("migrate: error creating transaction %w", err)
```
lastzero: `err` should probably be `tx.Error` here?

internal/entity/migrate/dbms_authid_migration.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+
+		defer func() {
+			if err == nil {
```
lastzero, line 41: The following should prevent Commit()/Rollback() errors from being silently ignored:

```go
defer func() {
	if err != nil {
		log.Warningf("migrate: rolling back DBMS AuthID Data Types (%v)", err)
		if rbErr := tx.Rollback().Error; rbErr != nil {
			log.Warningf("migrate: rollback failed: %v", rbErr)
		}
		return
	}

	if commitErr := tx.Commit().Error; commitErr != nil {
		err = fmt.Errorf("migrate: error committing DBMS AuthID Data Types: %w", commitErr)
		log.Warning(err)
		return
	}
}()
```

internal/entity/migrate/dbms_authid_migration.go
```diff
@@ -0,0 +1,196 @@
...
+		type pragmaTable struct {
+			Cid       int
+			Name      string
+			Type      string
+			Notnull   int
+			DfltValue string
+			Pk        int
+		}
+
+		var err error
```
lastzero: `err` is already specified as return value on line 13:
```go
func ConvertDBMSAuthIDDataTypes(db *gorm.DB) (err error) {
```

internal/entity/migrate/dbms_authid_migration.go
keif888: Yes, the joy of not being able to test the error path.  And rushing the fixes.

internal/entity/migrate/dbms_authid_migration.go
keif888: I have ensured that there are no silent failures.

internal/entity/migrate/dbms_authid_migration.go
keif888: extra var err error removed.

PR #5431, opened 2026-01-27
Add ability to specify Albums when Importing via the UI

keif888:
### Description

These changes implement the ability to create albums and/or apply existing albums to the images that are imported via the UI at /library/import.
The changes are front end only changes as they use the existing list of albums capability of api/v1/import.

### Acceptance Criteria

- [x] The user interface has been tested on Chrome, and Firefox and is fully responsive for use on phones, tablets, and desktop computers (I don't have Safari to test on)


graciousgrey:
Thank you for working on this :)

The reason why we haven't done this yet is that we haven't had time to think about possible pitfalls. For example one difference between upload and import is that there is a separate upload folder for each user. The import folder, on the other hand, is shared by all users of an instance. So we need to make sure that there are no unwanted side effects when multiple users use the import on one library.

Users can also specify that files from the import folder are copied to the “Originals” folder during import instead of being moved. In this case, all photos in a folder in the import folder will be added to the album when a folder is selected during import, not just newly added files. This may not be clear to users, which is different from uploading, where users consciously select the files to be uploaded.

keif888:
To address your 2nd concern I'm working on making visible the number of potential images that could be added to the selected albums, which can inform the user of the impact of their choices.
Until I have that available for review I've moved this back into draft.

keif888:
### The following is my analysis of what happens for imports with albums by name or uid via the underlying api.
It was tested against:
PhotoPrism® Plus
Build 251130-b3068414c

**Preliminary testing for an instance with two users utilising album and import via api has the following results:**
eg.
```
curl --location 'https://redacted:8443/api/v1/import/' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer redacted' \
--data '{
    "albums": ["test album"],
    "move":false,
    "path":"AlbumTest"
}'

curl --location 'https://redacted:8443/api/v1/import/' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer redacted' \
--data '{
    "albums": ["ata14nq918pvoa3g"],
    "move":false,
    "path":"AlbumTest"
}'
```

- User1 (user role) imports against "named album" that doesn't exist.  Results in new album uid "ata14xtmj12wzd5l".
- User2 (super admin) imports against "named album" (same name as above) that does exist but with a creator of User1.  Results in a new album uid "ata14nq918pvoa3g".
- Remove different images from album ata14xtmj12wzd5l and ata14nq918pvoa3g.
- User1 imports against "named album".  Results in album "ata14xtmj12wzd5l" being updated.
- User2 imports against "named album". Results in album "ata14nq918pvoa3g" being updated.
- Remove different images from album ata14xtmj12wzd5l and ata14nq918pvoa3g.
- User1 imports against "ata14xtmj12wzd5l". Results in album "ata14xtmj12wzd5l" being updated.
- User2 imports against "ata14xtmj12wzd5l". Results in no changes to either album.
- Remove some images from album ata14xtmj12wzd5l.
- User2 imports against "ata14xtmj12wzd5l".  Results in album "ata14xtmj12wzd5l" being updated.
- User1 imports against "ata14nq918pvoa3g". Results in album "ata14nq918pvoa3g" being updated.
- Remove some images from album ata14nq918pvoa3g.
- User2 imports against "ata14nq918pvoa3g".  Results in album "ata14nq918pvoa3g" being updated.

**The following is my code analysis by walking through the import API and calls it makes to the backend.**

- When using the front end, if the user selects an album from the UI, then the UID will be passed to the API.  This ensures that existing album(s) will be updated, as the user role can update all albums.
  -  (frontend/src/page/library/import.vue line 303)
- If the user enters the name of a new album, then the backend treats this as a request to create a new album IF that user does not already have an album with that name/slug.  
  - (frontend/src/page/library/import.vue line 301)
- The test for an existing album by name also tests the creator of the album and it must match the user's id from the import. 
  - (internal/entity/album.go FindAlbum does the matching)

Please note that I can't test this in the test environment as it uses PhotoPrism Essentials that doesn't support multiple users.

keif888:
To allow the user to see the impact of the import command, if the user drops down to select the folder to import from, the number of images in each folder (including children) will be shown.
image

And additional help text has been added to describe the use of the albums selection (shown here with 2 albums selected).
image

PR #5439, opened 2026-02-04
Scripts: Update FFmpeg installation script

omerdduran:
Refactors the FFmpeg installation script to exclusively use johnvansickle.com for all builds, properly supporting specific versions via the old-releases directory. This update also enhances error handling during download and extraction, adds a symlink for ffprobe, and includes post-install verification to ensure binaries are correctly configured.

### DIFF
--- a/scripts/dist/Makefile
+++ b/scripts/dist/Makefile
@@ -16,10 +16,10 @@ install-https:
 	@/scripts/install-https.sh
 ffmpeg: install-ffmpeg
 install-ffmpeg:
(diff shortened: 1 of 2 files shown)


lastzero:
Since the builds on https://www.johnvansickle.com/ffmpeg/ don't seem to be maintained anymore, we can remove that.. version 7 is already shipped with our Ubuntu base image now, so there is no reason to install the static build instead.

lastzero:
Thanks a lot! I'll close this PR since I applied the changes locally, which is causing a conflict now.

PR #5496, opened 2026-03-21
fix: save sidecar YAML after vision worker updates metadata

### LINKED ISSUE #5493: CLI "vision run" does not save sidecar YAML files (inconsistency with REST API)
### Before You Continue

- [x] This is a new, confirmed bug that has not yet been reported or documented

### What Is Not Working as Documented?

When running caption generation via Ollama using the CLI command (docker compose exec photoprism photoprism vision run --models=caption), the database is updated immediately upon caption generation, but the sidecar YAML files are not.

Tracing the code reveals an inconsistency between how the REST API handles saves versus how the background vision worker handles them.

Given that my environment (MacBook Pro M1 Max, 64GB RAM) handles native Ollama inference where each caption takes several seconds, skipping the millisecond YAML write does not yield a performance benefit, and leaves the backup YAMLs out of sync with the database. Furthermore, hitting the REST API with local zsh scripts triggers immediate YAML updates without performance penalties.

### How Can We Reproduce It?

```markdown
1.Execute docker compose exec photoprism photoprism vision run --models=caption

2.Monitor the logs, database and the sidecar YAML files.

2.Observe that the logs do not mention sidecar YAML, database reflects the new caption, but the sidecar YAML does not update.
```

### What Behavior Do You Expect?

The sidecar YAML file should be updated simultaneously with the database, maintaining consistency with the behavior of the REST API. And providing a full backup.

### What Could Be the Cause?

The issue appears to be a missing call to SaveSidecarYaml() in the vision worker loop, which is otherwise consistently applied in the API routes.

In the REST API (internal/api/photos.go etc.):
The SaveSidecarYaml function is reliably called after database updates to keep backups in sync.

In the Vision Worker (internal/workers/vision.go):
Around line 247 inside the Start function, the new caption is saved to the database using m.SaveVision(), but it immediately loops to the next photo without writing the YAML:

```
// Current implementation in internal/workers/vision.go
if changed {
    if saveErr := m.SaveVision(); saveErr == nil {
        updated++
    }
}
```

Suggested fix (not 100% sure):
```
// Proposed addition in internal/workers/vision.go
if changed {
    if saveErr := m.SaveVision(); saveErr == nil {
        updated++
        
        // Update YAML sidecar to match API behavior
        if w.conf.SidecarYaml() {
            _ = m.SaveSidecarYaml(w.conf.OriginalsPath(), w.conf.SidecarPath())
        }
    }
}
```

### Logs, Sample Files, or Screenshots

_No response_

### Which Software Versions Do You Use?

```markdown
Build 251130-b3068414c Docker
```

### On What Device Is PhotoPrism Installed?

```markdown
M1 Max 64GB
All on solid state
```

### Do You Use a Reverse Proxy, Firewall, VPN, or CDN?

_No response_

majiayu000:
### Description

Add missing `SaveSidecarYaml()` call in the vision worker. When the vision worker updates photo metadata (labels, caption), the changes are saved to the database via `SaveVision()` but the sidecar YAML backup files are never written. This means users with `--sidecar-yaml` enabled don't get YAML backups of vision-detected metadata.

The fix adds a `SaveSidecarYaml()` call inside the `if changed` block, following the same pattern used by the indexer in `index_mediafile.go`. It checks the `SidecarYaml()` config flag before writing and logs errors with the existing `vision:` prefix.

#### Related Issues

- Fixes #5493

### Acceptance Criteria

- [ ] Automated unit and/or acceptance tests are included to ensure that changes work as expected and to reduce repetitive manual work
- [ ] Documentation has been / will be updated, especially as it relates to new configuration options or potentially disruptive changes
- [ ] Database-related changes have been successfully tested with SQLite 3 and MariaDB 10.5.12+

Note: This is a one-line logic fix (missing function call), not new functionality. The underlying `SaveSidecarYaml` method is already tested in `internal/entity/photo_yaml_test.go`. No UI, database schema, or config changes are involved.

### DIFF
--- a/internal/workers/vision.go
+++ b/internal/workers/vision.go
@@ -270,6 +270,13 @@ func (w *Vision) Start(filter string, count int, models []string, customSrc stri
 			if saveErr := m.SaveVision(); saveErr == nil {
 				updated++
 			}
+
+			// Save sidecar YAML backup if enabled.
+			if w.conf.SidecarYaml() {
+				if yamlErr := m.SaveSidecarYaml(w.conf.OriginalsPath(), w.conf.SidecarPath()); yamlErr != nil {
+					log.Errorf("vision: %s (save yaml sidecar)", yamlErr)
+				}
+			}
 		}
 
 		if mutex.VisionWorker.Canceled() {


lastzero (approved):
Reviewed the patch and validated the affected code paths locally. The sidecar YAML save now matches the existing indexer behavior, and focused tests passed for `internal/entity` and `internal/workers`.

PR #5505, opened 2026-03-31
Lightbox: Add inline metadata editing and face markers #4965 #4966

omerdduran:
## Description

Expanded lightbox sidebar: from a small "title/date/size" panel to a full metadata workbench — inline edits, date/time and camera dialogs, interactive face-marker overlay, label/album/person chips, and role-based gating for restricted viewers.

Closes #4965 #4966 — Refs #1307 #1548 #5081

### Sidebar metadata (frontend/src/component/sidebar/info.vue)
- Renders full metadata: title, caption, taken-at, camera/lens/EXIF, GPS + map preview, subject/artist/copyright/license, keywords, notes, technical type/size/codec.
- Inline editing (Enter/Escape/blur) with per-field `editingField`.
- Label/album/person chips deep-link to matching lists; private albums filtered client-side (commit `610e75d43`).
- Edits route through `Photo` actions and evict per-UID cache so next open refetches authoritative state.

### Date/time and camera dialogs
- `datetime-dialog.vue` (304 LOC): Y/M/D and H/M/S autocompletes, timezone select, `invalidDate` validation, reset to taken-at.
- `camera-dialog.vue` (204 LOC): camera/lens selectors driven by `$config.values.cameras/lenses`, ISO, exposure, f-number, focal-length with bounded validation.
- Both hydrate via `loadFromPhoto()` and emit a single `confirm` payload for atomic sidebar persists.

### Face-marker overlay (frontend/src/component/photo/face-marker-overlay.vue, 558 LOC)
- SVG overlay on PhotoSwipe viewport; supports `display` and `draw` modes.
- Drawing: drag-to-create (MIN_DRAW_SIZE=16), four resize handles (tl/tr/bl/br), on-canvas confirm/cancel; emits `create` / `cancel` events consumed by lightbox marker actions.
- Named markers render in accent color and are read-only by default; rename gated behind explicit "eject" affordance (commit `184bfe568`) to prevent accidental overwrite.
- Markers with `SubjUID` link to the person page; new chip iconography (commit `05e3365d7`) clarifies named vs unnamed markers.

### Lightbox integration (frontend/src/component/lightbox.vue, +329 LOC)
- Mounts face-marker overlay only when feature flag (`featPeople`), edit permission, and PhotoSwipe instance align.
- Adds `markersVisible` toggle, `addingMarker` mode, and event plumbing (`onCreateFaceMarker`, `cancelAddingMarker`).
[...]

PR #5547, opened 2026-04-29
Sidebar: Bind dialog photo prop via $view alias #5505

omerdduran:
The 7a611b6d6 refactor removed the legacy photo prop on PSidebarInfo; however three dialog tags in info.vue still referenced the now-undefined photo identifier. As a result the date/camera/location dialogs opened with empty inputs even when the photo had values.


lastzero:
Thank you! :)

Issue #4106, opened 2024-03-17
Metadata: Support reading GPS information from xmp


derchr:
#### 1. What is not working as documented?

Reading in GPS data from XMP sidecar files generally works. However, when updating (e.g., adding them for the first time) GPS information with an external program such as digikam, it will not get updated in Photoprism even after a complete re-index.

Often GPS files are added afterwards based on a GPX file because the camera itself does not have an embedded GPS tracker.

#### 2. How can we reproduce it?

1. Index an image without any GPS data. Photoprism will show Lat 0 and Long 0 in the details.
2. Now add GPS information with an external program such as digikam or darktable.
3. Do a complete re-index in Photoprism.
4. Photoprism will still show Lat 0 and Long 0 in the details and not display the images in the map.

#### 3. What behavior do you expect?

Photoprism should pick up the new GPS information and save it in its internal yml sidecar files and show the geoinformation in the UI.

#### 4. What could be the cause of your problem?

The GPS location gets read in when first importing, but it cannot be not updated afterwards by a re-index operation.

#### 6. Which software versions do you use?

Latest docker image with latest docker-compose example.

graciousgrey:
Thank you for your feedback!

We do not yet read GPS information from xmp files. 

If you let digikam write the GPS information into the EXIF fields supported by PhotoPrism, Lat/Lng should be updated during the next indexing.

https://www.photoprism.app/kb/metadata

derchr:
Thank you for triaging my issue.

I want to add some further findings:
- As I primarily talk about RAW files, the additional GPS data is *only* present in the XMP file as an exif field.
- When the image files are first imported, the additional GPS data is in fact read in from the XMP. Maybe has something to to with exiftool?
- After changing the GPS data in the XMP file, neither a simple re-index nor a full re-index will update the Photoprism GPS entry.
- Even after `touch`ing all files, *including* the RAW and sidecar files, a re-index will not pick the new changes up.
- The only way to update the GPS information is to temporaily delete the image files, do a cleaning re-index, add the files again and doing another re-index.

GlassedSilver:
Someone correct me if I'm wrong here, but wouldn't the same limitation of not reading XMP for geodata also apply to writing such data? Say I use Photoprism to assign a location to a picture that metadata would ONLY live in Photoprism right?

graciousgrey:
@GlassedSilver Except from the image orientation PhotoPrism does not yet write metadata changes back to originals files. 

[...]

Issue #4094, opened 2024-03-10
Stacks: Stack files by dc:identifier


EricMyhrer:
#### 1. What is not working as documented?

Images are not stacking at all. Both Unique ID and Sequential Name options are turned on. Seems similar to #576, but my files are closely named

These images do not get put into a stack

Filename examples:
111353204_p0.xmp
111353204_p1.xmp
111353204_p0.jpg
111353204_p1.jpg

111353204_p0.xmp dc:identifier="111353204"
111353204_p1.xmp dc:identifier="111353204"

In fact, the xmp sidecars are identical aside from filename.

#### 2. How can we reproduce it?

1. Move files via ftp or command line, etc
2. Start an index
3. Check Library->Stacks to see no images stacked

When reporting an import, indexing, or performance issue, please include the number and type of pictures in your library, as well as any configuration options you have changed, such as for thumbnail quality.

#### 3. What behavior do you expect?

For these images to be staacked

#### 4. What could be the cause of your problem?

No idea. 

#### 5. Can you provide us with example files for testing, error logs, or screenshots?

Filename examples:
111353204_p0.xmp
111353204_p1.xmp
111353204_p0.jpg
111353204_p1.jpg

111353204_p0.xmp dc:identifier="111353204"
111353204_p1.xmp dc:identifier="111353204"

Log file example:
```
2024-03-09 23:24:18 INFO indexing completed in 7 s

2024-03-09 23:24:18 INFO index: updated 4 files [7.329822166s]

2024-03-09 23:24:17 INFO index: stacked related xmp file 116712697_p0.xmp

2024-03-09 23:24:17 INFO index: added main png file 116712697_p0.png

2024-03-09 23:24:17 INFO index: stacked related xmp file 116712697_p1.xmp

2024-03-09 23:24:17 INFO index: added main png file 116712697_p1.png

2024-03-09 23:24:17 INFO index: matched 1 label with 116712697_p1.png [3.092411089s]

2024-03-09 23:24:14 INFO media: 116712697_p0.png was taken at 2024-03-10 04:23:33.85586975 +0000 UTC (file birth time)

2024-03-09 23:24:14 INFO media: 116712697_p1.png was taken at 2024-03-10 04:23:33.779870866 +0000 UTC (file birth time)

2024-03-09 23:24:14 INFO media: created 12 thumbnails for 116712697_p0.png [638.673368ms]

[...]

Issue #2828, opened 2022-10-31
Metadata: Read GPS information from Apple XMP


graciousgrey:
Acceptance Criteria:
- [ ] Read GPS information from apple xmp
- [ ] Write unit test
- [ ] Update documentation

Follow up for #636

gacolitti:
I'm considering migrating a large Apple Photos library to PhotoPrism. I want to retain GPS information. If I migrate now, GPS data will be lost and there is no easy way to add it in the future, correct? So it's best to wait for this issue to be closed first.

gacolitti:
I followed the migrating from Apple Photos [guide](https://docs.photoprism.app/user-guide/use-cases/apple/), which seems to suggest GPS data is not currently read. But I see location info (including Lat and Long) in PhotoPrism after importing. Is this information read from somewhere other than Apple XMP?

graciousgrey:
Yes, GPS data is read from your files EXIF data, see https://photoprism.app/kb/metadata & https://docs.photoprism.app/user-guide/library/metadata/

lastzero:
Improved support for reading metadata from XMP sidecar files has been merged into `develop`, so it will ship with the next stable release and can be tested using our preview builds:

- #5563
- https://docs.photoprism.app/release-notes/#development-preview

Issue #2260 has been turned into an epic to track XMP-related enhancements and feature requests:

- #2260

Issue #2075, opened 2022-02-27
Metadata: Support for darktable's `dc:subject`


maxammann:
**Is your feature request related to a problem? Please describe.**

I just tagged my pictures in darktable and noticed that photoprism is not aware of them. Darktable uses the `dc:subject` for keywords/tags/labels: https://docs.darktable.org/usermanual/3.8/en/module-reference/utility-modules/shared/tagging/#editrename-tag

**Describe the solution you'd like**

Entries from the  `dc:subject` could be added to the keywords or labels of photoprism.

**Describe alternatives you've considered**

An alternative would be to write a tool which extracts the darktable tags from the XMP files and converts them to a format photoprism already supports.

**Additional context**

I attached two xmp files.
[XMPs.zip](https://github.com/photoprism/photoprism/files/8148550/XMPs.zip)

maxammann:
It seems like exiftool already supports "Subject":
```
ExifTool Version Number         : 12.30
File Name                       : MAX_0297.NEF.xmp
Directory                       : .
File Size                       : 1393 bytes
File Modification Date/Time     : 2022:02:27 12:00:13+01:00
File Access Date/Time           : 2022:02:27 12:23:38+01:00
File Inode Change Date/Time     : 2022:02:27 12:00:15+01:00
File Permissions                : -r--------
File Type                       : XMP
File Type Extension             : xmp
MIME Type                       : application/rdf+xml
XMP Toolkit                     : XMP Core 4.4.0-Exiv2
Date/Time Original              : 2022:01:30 22:19:12
Rating                          : 0
Derived From                    : MAX_0297.NEF
Import timestamp                : 1645884581
Change timestamp                : -1
Export timestamp                : -1
Print timestamp                 : -1
Xmp version                     : 4
Raw params                      : 0
Auto presets applied            : 0
History end                     : 0
Iop order version               : 2
Masks history                   :
History                         :
Creator                         : Maximilian Ammann
Subject                         : funchal
Hierarchical Subject            : funchal
```

Faexa:
Hello devs,
First of all, thank you very much for your product and your time. I'm in testing and it seems very promising for my use case.

I'm very confused by the differences between keywords (which are translated to labels by photoprism if set one time manualy) and Subject.

Anyway, for the darktable case cited before :
From what I've understood you use exiftool to read EXIF and embedded XMP tags in JPG whereas you parse XML for XMP files ?
[...]

PR #5563, opened 2026-05-06
Index: Improve support for metadata in XMP sidecar files

### LINKED ISSUE #2260: Metadata: Improve XMP sidecar file support
**As a user with a lot of metadata in XMP sidecar files, I want PhotoPrism to read (and eventually write) more of that information, so that I can view, search, and organize my pictures based on the tags my photo software already created.**

This is an **epic** that tracks all related XMP read/write enhancements, bug reports, and pull requests. Please file specific problems and feature requests as separate issues and link them here rather than expanding this description.

Our goal is to make PhotoPrism a first-class citizen in XMP-based workflows by:

- reading a broad, well-tested set of XMP tags from both embedded and sidecar metadata;
- grouping related files via their XMP document identifier (stacking);
- importing face-region metadata written by other applications; and
- eventually writing PhotoPrism's own metadata back to XMP/EXIF.

The XMP sidecar reader was rewritten in #5563 on top of `antchfx/xmlquery` + XPath, replacing the original proof-of-concept `encoding/xml` reader and expanding sidecar tag coverage from 10 to 33 fields (title, caption, keywords/subject, GPS lat/lng/altitude, capture time with `OffsetTime` handling, camera/lens, exposure, copyright/license, document/instance IDs, and more).

### Place & Time

- [ ] #4106 — Support reading GPS information from XMP _(reader landed in #5563; values are applied on a complete rescan)_
- [ ] #2828 — Read GPS information from Apple XMP _(covered by #5563)_
- [ ] #3894 — Timestamps may be incorrect for non-standard XMP sidecars created by PhotoSync

### Labels

- [ ] #5714 — Match multi-word label names from title, caption, subject, and keyword text (links existing Labels only)
- [x] #2075 — Support for darktable's `dc:subject` _(read into the Subject field; searchable and label-generating)_
- [ ] #5710 — Import hierarchical XMP keyword tags (`lr:hierarchicalSubject`, `digiKam:TagsList`, `MicrosoftPhoto:LastKeywordXMP`) as Labels
- [ ] #1828 — Add support for digiKam pick/rating labels/tags (e.g. `digiKam:PickLabel`)
- [ ] #1143 — Automatically create labels for keywords from photo metadata (Exif/IPTC/XMP)
- [ ] #5513 — Option to disable automatic label generation from captions and titles (opt-out for the same text sources as #5714)

### Face Regions

- [ ] #554 — Import Adobe's XMP face-region metadata
- [ ] #747
- [ ] #1570 — Import Microsoft's XMP face-region metadata
- [ ] Related PR in progress: #5572 — Import MWG XMP face regions from sidecar files

### Stacking

- [ ] #4094 — Stack photos by `dc:identifier` from XMP sidecar files _(#5563 now stores the identifier as the photo `UUID`; grouping separate files by a non-canonical identifier from a sidecar is still open)_

### Writing & Embedding

- [ ] #402 — Write PhotoPrism metadata to XMP sidecars or EXIF
- [ ] #243 — Embed XMP metadata in JPEG files

### Completed

- [x] #5563 — Index: Improve support for metadata in XMP sidecar files (reader rewrite, 33-field coverage)
- [x] #1161 — Add support for XMP sidecar `CreateDate` and `Keywords`
- [x] #1873 — Add support for reading the `fstop:favorite` flag
- [x] #636 — Read metadata from Apple iPhoto / Photos app XMP

### Documentation

- Developer Guide › Metadata › XMP: https://docs.photoprism.app/developer-guide/metadata/xmp/
- Reader implementation & supported-tag coverage: [`internal/meta/README.md`](https://github.com/photoprism/photoprism/blob/develop/internal/meta/README.md)
- XMP specification (ISO 16684-1) and overview: https://www.adobe.com/products/xmp.html

omerdduran:
Closes #2260. Touches #4106, #4094, #2828, #2075

## Summary

- Replaces the proof-of-concept `encoding/xml` XMP sidecar reader with an XPath-based implementation on `antchfx/xmlquery` + `antchfx/xpath`.
- Expands sidecar tag coverage from 10 to 33 fields per the gap-analysis (4 preserved, 6 extended chains, 23 new accessors). Section-2 follow-ups (rating, hierarchical keywords, face regions, structured rights/location) are explicitly **not** persisted.
- Adds the namespace-priority fallback engine mandated by the brief, plus security guards (1 MiB size cap, 64-level depth cap, XXE regression test), GPS 2-component Adobe form support, migration tests, and updated `internal/meta/README.md`.

## Highlights

**Engine**
- `chainXPath`: ordered, pre-compiled XPath expressions; first non-empty match wins (analogous to the embedded path's `meta:"A,B,C"` convention).
- `elemOrAttr` helper: matches both element and attribute encodings of scalar properties (digiKam emits `xmpMM:*`/`exif:*` as attributes, Adobe writes child elements; both are valid RDF/XML).
- All chains compile once at package init via `xpath.CompileWithNS`.

**Composition (encapsulated in accessors, not the engine)**
- `Lat` / `Lng`: cardinal in the value (Adobe form `"52,30.4567N"`) wins; otherwise `*Ref` supplies the sign.
- `Altitude`: rational + `GPSAltitudeRef = "1"` inverts sign.
- `TakenAt`: joins `SubSecTimeOriginal` only when the parsed datetime lacks fractional seconds (preserves photoshop.xmp parity).
- `TakenGps`: canonical combined ISO 8601 → legacy split-form `GPSDateStamp + GPSTimeStamp` fallback.
- `Flash`: reads only `exif:Flash/Fired`.
- `Exposure`: `ExposureTime` (rational seconds) → `ShutterSpeedValue` (APEX, `t = 2^(-Tv)`).

**`GpsToDecimal` extension**
- 2-component Adobe form (`"52,30.4567N"`) added next to the existing 3-component DMS branch.
- `TestGpsToDecimal_RegressionAgainstExistingFixtures` asserts identical output for the named EXIF/JSON fixtures the embedded path consumes (`gopher-original`, `panorama360`, `berlin-landscape`, `date.mov`).

**Security**
- 1 MiB file-size cap (`ErrXmpFileTooLarge`) via `os.Stat` + `io.LimitReader` defence-in-depth.
- 64-level element-nesting cap (`ErrXmpTooDeep`) via post-parse traversal.
- XXE regression test (`encoding/xml` refuses external entities by default; this guards future loader changes).
- `apexToSeconds` clamped to `[-30, 30]` to avoid `±Inf` on pathological input.

**Migration**
- `TestXMP_OverridesPreExistingMetaValues`: stale embedded-path values give way to the sidecar; sidecar omissions leave embedded values untouched.
- `TestXMP_FillsPreviouslyEmptyFields`: partial-sidecar (old reader) → full-sidecar (new reader) fills in `License`, GPS, `xmpMM` IDs, `Copyright`, `ColorProfile`, `Keywords`.
- `TestXMP_GpsCompositionEndToEnd`: float `Lat`/`Lng`/`Altitude` populate; ExifTool-format `GPSLatitude`/`GPSLongitude` strings preserved.

Per-field `SrcMeta` → `SrcXmp` transitions in the database are enforced by the entity setters that consume `meta.Data` (no code change required at this layer; structural via `internal/entity/src.go`).

**Tests**
72 top-level functions, 288 sub-tests. Every gap-analysis section-1 in-scope tag has fixture-backed coverage. Section-2 follow-up tags verified absent (`TestXmpDocument_DarktableFixture/OutOfScopeTagsIgnored`).

**Fixture corpus**
Adobe Bridge, Darktable, digiKam, plus 12 hand-written synthetic fixtures under `internal/meta/testdata/xmp/{adobe,darktable,digikam,synthetic}/` with `.exiftool.txt` reference siblings (ExifTool 13.25). Layout documented in `internal/meta/testdata/xmp/README.md`.


go.sum
```diff
@@ -34,6 +34,10 @@ github.com/abema/go-mp4 v1.6.0/go.mod h1:vPl9t5ZK7K0x68jh12/+ECWBCXoWuIDtNgPtU2f
 github.com/alexbrainman/sspi v0.0.0-20250919150558-7d374ff0d59e h1:4dAU9FXIyQktpoUAgOJK3OTFc/xug0PCXYCqU0FgDKI=
 github.com/alexbrainman/sspi v0.0.0-20250919150558-7d374ff0d59e/go.mod h1:cEWa1LVoE5KvSD9ONXsZrj0z6KqySlCCNKHlLzbqAt4=
 github.com/andybalholm/cascadia v1.1.0/go.mod h1:GsXiBklL0woXo1j/WYWtSYYC4ouU9PqHO0sqidkEA4Y=
+Meta: Tidy XMP sidecar reader comments and reuse logName local #2260github.com/antchfx/xmlquery v1.5.1 h1:T9I4Ns1EXiWHy0IqKupGhnfTQtJwlGrpXtauYOoNv78=
```
lastzero: What happened here? xD

omerdduran: I accidentally wrote the commit message to the go.sum file 😄. I will delete now

lastzero:
@omerdduran Overall, this looks like it's headed in the right direction. Please take a look at the findings below and address them as needed. I will then do another review with possible refactoring suggestions. Thanks a lot! :hugs: 

### 1. `go.sum` line 37 is corrupted — `go build ./...` fails immediately

In the most recent commit (`93632e2db Meta: Tidy XMP sidecar reader comments and reuse logName local #2260`), the commit subject got pasted onto the start of an entry in `go.sum`:

```
Meta: Tidy XMP sidecar reader comments and reuse logName local #2260github.com/antchfx/xmlquery v1.5.1 h1:T9I4Ns1EXiWHy0IqKupGhnfTQtJwlGrpXtauYOoNv78=
```

This is a hard build break — `go build` exits with `malformed go.sum: /go/src/.../go.sum:37: wrong number of fields 13` and no further work is possible. The fix is to strip the leading `Meta: Tidy XMP sidecar reader comments and reuse logName local #2260` prefix so the line reads `github.com/antchfx/xmlquery v1.5.1 h1:T9I4...`. Running `go mod tidy` regenerates it correctly; alternatively, hand-edit the line. Worth re-running `make test-short` after fixing to make sure nothing else slipped in via the same paste.

CI will catch this at the first `go vet` / `go build` step, but it should not have shipped to a draft PR — recommend Ömer run `go build ./... && go test -count=1 ./internal/meta/... ./internal/photoprism/...` locally before pushing the next round.

### 2. `data.Subject` is never populated by the XMP reader

This might be under development and not finished yet? :thinking: 

### 3. `data.CreatedAt` overwriting `data.TakenAt` in the shared resolver

`internal/meta/resolver.go` opens with:

```go
if !data.CreatedAt.IsZero() {
    data.TakenAt = data.CreatedAt
}
```

This was inherited from the ExifTool path, where `CreatedAt` maps to `SubSecCreateDate` / `CreateDate` (typically the same instant as `DateTimeOriginal` for stills, but authoritative for videos). On the XMP path the chain is different:

- `data.TakenAt` ← `photoshop:DateCreated → exif:DateTimeOriginal → xmp:CreateDate`
- `data.CreatedAt` ← `xmp:CreateDate → xmpDM:CreationDate`

When a sidecar carries both `photoshop:DateCreated` (the capture instant Adobe treats as authoritative) and a different `xmp:CreateDate` (e.g. when the file was first written to disk later), the resolver overwrites the carefully-chosen `photoshop:DateCreated` with the less-specific `xmp:CreateDate`, contradicting the spec's stated priority (specs/media/xmp-sidecar.md, "TakenAt" row).

This **happens to be masked** in the current tests because:

- The `Photoshop` test fixture has GPS coordinates, so after the CreatedAt overwrite the GPS branch re-derives `TakenAt` from `TakenAtLocal` (which the XMP path preserved correctly).
- The synthetic `time-offsets-subsec` fixture has identical seconds on `photoshop:DateCreated` and `xmp:CreateDate`, so the overwrite is a no-op.

Add a regression test where `photoshop:DateCreated` and `xmp:CreateDate` disagree by more than a few seconds AND no GPS coordinates are present — that test will fail today. The minimal fix is to either (a) guard the resolver's `CreatedAt` overwrite behind a video-only check (the original intent — video containers genuinely use `CreateDate` as the capture time), or (b) only apply the overwrite when `TakenAt.IsZero()` (preserve a non-empty TakenAt). Option (b) is the safer change and matches the docstring expectation; option (a) requires a MimeType signal the XMP path doesn't currently set.

### 4. `data.TakenAt` ends up with wall-clock time tagged as UTC when `OffsetTime*` is present but `photoshop:DateCreated` has no inline offset

I ran a probe on `testdata/xmp/synthetic/time-offsets-subsec.xmp`:

```
TakenAt:      2026-05-06 15:42:18.123456 +0000 UTC (loc=UTC)
TakenAtLocal: 2026-05-06 15:42:18.123456 +0000 UTC
TimeZone:     UTC+2
```

The sidecar says `+02:00`, so the true UTC instant is 13:42:18, not 15:42:18. The resolver's `else if hasTimeOffset` branch only fires when `TakenAtLocal` carries a non-zero `Zone()` offset — and `txt.ParseTime("2026-05-06T15:42:18.123456", "")` returns a UTC value because the date string has no offset suffix. The resolver therefore never reaches the branch that would shift TakenAt to UTC using the OffsetTime cascade.

The current test `TestXMP/SyntheticTimeOffsetsSubsec` asserts the wall-clock value AS IF it were UTC (`"2026-05-06 15:42:18"`) and passes — which means the test enshrines the bug rather than catching it. This is a pre-existing bug in the ExifTool resolver (same code path), but the XMP rewrite now exercises it, and the spec promises EXIF 2.31 `OffsetTime*` cascade support. Either the resolver should also consume `data.TimeOffset` when `TakenAtLocal` has no inline offset, or the XMP reader should apply the offset string to TakenAt before handing off (`time.Parse` with the offset, then re-tag the location). I'd recommend the latter — fix in `xmp.go` so the resolver continues to assume "TakenAtLocal carries its own location".

### 5. `photo.UUID` (column `VARBINARY(64)`) can be truncated by long XMP DocumentIDs

The indexer assigns `photo.UUID = data.DocumentID` directly (`internal/photoprism/index_mediafile.go`, the `IsXMP` branch), bypassing the strict `rnd.IsUUID` check on purpose because real-world XMP IDs are non-canonical. The spec at `specs/media/xmp-sidecar.md` already flags this as an open follow-up: "Confirm there is no length constraint on `photos.uuid` that would reject long non-canonical DocumentID values."

Confirmed: `internal/entity/photo.go` defines `UUID string \`gorm:"type:VARBINARY(64);index;"\``. A value like `adobe:docid:photoshop:7d592d87-eb1e-1040-809a-e16c6b85b3fd` is 56 bytes — under the limit but only just. Some real-world Lightroom and Camera Raw sidecars emit longer compound identifiers (`adobe:docid:photoshop:` + `uuid:` + 36-char UUID = 65 bytes). The Adobe XMP spec doesn't bound DocumentID length, so this WILL truncate or fail to insert silently on some user files.

### 6. Indexer wiring duplicates work and lacks a "main file not yet indexed" guard

In `index_mediafile.go`, the `IsXMP` branch calls `photo.PrimaryFile()` to mirror `InstanceID` and `Software` to the primary file row. This implicitly assumes the primary file (JPEG/RAW/HEIC) was already indexed in the same `IndexRelated` pass — which is true for the documented order, where `IndexMain` runs before related sidecars. But:

- `PrimaryFile()` requires `photo.PhotoUID` to be non-empty. If the photo is brand-new and somehow the .xmp file is processed before its primary (e.g. when `RelatedFiles` orders files unexpectedly), this returns `(nil, gorm.ErrRecordNotFound)` and the mirror block silently skips. That's the failure mode the code handles, so functionally OK.
- The branch then calls `primary.Save()` — which is an extra DB write per XMP sidecar on every full re-index pass. The spec's "Performance" section flags this: "for large re-index passes consider batching if measured throughput regresses." 100k photos = 100k extra UPDATE statements.
 
These might not be hard blockers, but they are worth investigating/tracking. Maybe there is a better solution? :wrench: 

### 7. Two cosmetic violations of the "no blank line between consecutive `t.Run` subtests" rule

Per `go-testing.md`, top-level `t.Run(...)` calls inside the same `Test*` function should sit back-to-back without blank-line separators (subtests read as a table). Two files violate this:

- `internal/meta/xmp_document_test.go::TestXmpDocument_DarktableFixture` — four subtests each separated by a blank line.
- `internal/meta/resolver_test.go::TestData_ResolveTimeZone` — eleven subtests, all separated.

This is purely a style nit — no behavior impact. Run `make fmt-go` if you have a hook that catches it, otherwise just delete the blanks before merge.

### 8. `loadXmpString` / `writeFile` use the hard-coded `0o600` literal instead of `fs.ModeFile` / `fs.ModeSecretFile`

`internal/meta/xmp_document_test.go` lines 44 and 211 use `0o600` directly when writing temp XMP fixtures. Per `go-code-style.md`, file modes should use the shared `fs.Mode*` constants. For test temp files `fs.ModeFile` (0644) is most appropriate; `0o600` is reserved for secrets. Drive-by cleanup for the next round.

### 9. Doc-comment line references in spec point at line numbers

`internal/meta/xmp.go:155` says `// (exif.go:321-328, json_exiftool.go:282-289)`. Per "No Line Numbers in File References" rule (which applies to specs but the principle is identical for code comments — line numbers rot), prefer naming the symbol: `// (exif.go's AutoAddKeywords block, json_exiftool.go's panorama / HDR detection)`. Tiny nit; ignore if you don't agree.

lastzero:
Well done! I merged this to `develop` for final testing. :partying_face:

Issue #713, opened 2020-12-16
Star/label photo's in focused view


hampelman1:
I would love to be able to label and (most importantly) star (5-star system) my photo's in the focused view mode..

This way I can use photoprism to manage my large collection...

lastzero:
Are you actually using 1 to 5 stars in a way that makes a significant difference if an image has 4 or 5 stars?

I felt like this provides no value to me (actually huge waste of time) and most photographers, so we decided to implement a simple like button instead. It would be relatively easy to map 3-5, 4-5 or just 5 stars to this so that existing metadata can be used.

hampelman1:
yes, the numbers do make a difference. I have a huge collection, and by using the star system, i can filter them somewhat. I could maybe do with 3...but then compatability wth other tools (and metadata fields???) is broken..

Synology Photostation uses the 5 star system, as does digikam. (https://docs.kde.org/trunk5/en/extragear-graphics/digikam/using-dam.html > scroll to rating/ranking)

1 star: Meh. Photo could safely be deleted in the future (for example when running out of space)
2 stars: Photo should be retained, but nothing special
3 stars: good quality photo, cloud be printed in photobook (if you're into that*)
4 stars: highlights of a trip/year: use in a slideshow for others
5 stars: very good photo. Top of the top. Poster-like quality.

* my wife likes very large photobooks. Distinction between ratings may vary upon personal workflow...

graciousgrey:
I changed the title to keep this ticket to implement a star rating. 
We have another ticket to customise actions that can be done in fullscreen mode --> https://github.com/photoprism/photoprism/issues/508

bernsel:
I also use the Star Rating system, as a way to quickly categorize images. In DigiKam on MacOS, it is possible to use keyboard shortcuts (Command+1 to Command+5) to very quickly set a Star Rating on a photo when you preview or scroll through, which makes it a quick and effective way of organizing images. I do not use the stars as a quality metric, but rather as a classification metric. This is how I apply them.

1 star:  Picture to be deleted
2 stars: Work-related picture - should be moved to work folder
3 stars: Picture of me + daughter, or me+girlfriend
4 stars: Image good enough to include in Album
5 stars: Image I want to print as Poster

After classifying like this in DigiCam, I can filter on the exact number of stars in DigiCam, as a quick double-check to see I got each classification right. After that, I select "write metadata to files" in DigiKam, to preserve the rating in the files themselves (and delete the 1-star images in bulk).

[...]

PR #5569, opened 2026-05-11
Metadata: Add photo star ratings

Shlummie:
Adds the 5-star photo rating plumbing requested in #713.

This keeps ratings separate from the existing quality score, imports ratings from XMP / ExifTool metadata, tracks the rating source, and lets manual or batch ratings take priority over lower-priority metadata updates.

Also adds API/search result exposure, `rating:` search filtering, batch edit support, and frontend controls in the photo info/sidebar flows.

Checks run:
- frontend lint on changed files
- focused frontend model/component tests
- frontend build
- gofmt

I could not run the full Go test set locally because this Windows checkout is missing native TensorFlow/vips dependencies and hits an unrelated Windows syscall compile issue.


photoprismci:
Thanks for the work, but we can't merge this. The main reasons:

- `make test-go` fails on the branch. The PR description notes you weren't able to run the Go suite locally, which is understandable, but a change of this size (new column, migrations, search filter, batch edit, metadata import, two viewer result types) needs to be green before we can review it.
- `make test-js` also fails, even though the description lists the focused frontend tests as passing.
- The batch-edit "mixed rating" state is wired on the Go side but doesn't render in the UI, and the new row breaks the surrounding layout.
- The schema collapses "0 stars" and "never rated" into the same value; we want to keep those distinguishable.
- We're mid-rework on the XMP/ExifTool metadata parser, so the `internal/meta/` changes here would conflict with work in flight.

5-star ratings are on our roadmap and we'll take another look at this when we get to it. Closing for now.

PR #5571, opened 2026-05-11
Search: use NATURAL_SORT_KEY for name and title sort order on MariaDB

### LINKED ISSUE #4912: UX: Implement Natural Sorting for Numerically Named Files in Folder View
**Describe what problem this solves and why this would be valuable to many users**  
Current lexicographical file sorting in folder view (`/library/folders/`) causes numerically named files (e.g., `1.jpg`, `2.jpg`...`101.jpg`) to appear in non-natural order (`1 → 10 → 100 → 2`). This disrupts workflows for:

1. Photographers managing sequenced shots
2. Users organizing scanned documents
3. Scientific datasets with numeric naming

Natural sorting is an expected behavior in modern file management systems and media applications.

Implement File Name to **natural sort order** (numeric-aware sorting) specifically in folder view, following this pattern:
1.jpg → 2.jpg → ... → 9.jpg → 10.jpg → 11.jpg → ... → 100.jpg → 101.jpg

1. Manual zero-padding (e.g., `001.jpg`)  
   - Not feasible for existing libraries  
   - Requires batch renaming tools  
2. Using metadata-based sorting  
   - Doesn't solve filename-driven organization needs  
3. Third-party sorting tools  
   - Breaks integrated workflow  

**Additional context**  
1. Current behavior screenshot:  
Image

2. Technical justification:  
   - Natural sort is [ISO/IEC 14651](https://en.wikipedia.org/wiki/ISO/IEC_14651) compliant  
   - Supported natively in modern browsers via `Intl.Collator`

ChrisScott9456:
### Description

Implements natural sort order for file name and title sorting by using MariaDB's NATURAL_SORT_KEY() function, so that numerically named files sort in human-readable order (1, 2, 3... 10, 11 instead of 1, 10, 11, 2, 3).

#### Related Issues

- Fixes: https://github.com/photoprism/photoprism/issues/4912

### Acceptance Criteria

### DIFF
--- a/internal/entity/search/photos.go
+++ b/internal/entity/search/photos.go
@@ -204,9 +204,17 @@ func searchPhotos(frm form.SearchPhotos, sess *entity.Session, resultCols string
 		s = s.Where("files.file_diff > 0")
 		s = s.Order(OrderExpr("photos.photo_color ASC, photos.cell_id ASC, files.file_diff, files.photo_id, files.time_index", frm.Reverse))
 	case sortby.Name:
-		s = s.Order(OrderExpr("photos.photo_path ASC, photos.photo_name ASC, files.time_index", frm.Reverse))
+		if entity.DbDialect() == entity.MySQL {
+			s = s.Order(OrderExpr("NATURAL_SORT_KEY(photos.photo_path) ASC, NATURAL_SORT_KEY(photos.photo_name) ASC, files.time_index", frm.Reverse))
+		} else {
+			s = s.Order(OrderExpr("photos.photo_path ASC, photos.photo_name ASC, files.time_index", frm.Reverse))
+		}
 	case sortby.Title:
-		s = s.Order(OrderExpr("photos.photo_title ASC, photos.photo_name ASC, files.time_index", frm.Reverse))
+		if entity.DbDialect() == entity.MySQL {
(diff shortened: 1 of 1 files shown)


lastzero:
@ChrisScott9456 Thanks a lot for your contribution! After reviewing the changes in PR #5722, which follows the same approach, I found two potential issues that appear difficult to resolve:

### Compatibility with MariaDB <10.7

`NATURAL_SORT_KEY()` was introduced in MariaDB 10.7.0. However, our current minimum supported version is 10.5.12, and 10.6 is an LTS release with a large user base:
- https://mariadb.com/docs/server/reference/sql-functions/string-functions/natural_sort_key

On any supported MariaDB in the 10.5.12-10.6.x range, the query will fail with "ERROR 1305: FUNCTION ...NATURAL_SORT_KEY does not exist", so every name/title search would return error 500.

### `NATURAL_SORT_KEY` defeats the index

There is a composite index, `idx_photos_path_name` (`photo_path`, `photo_name`), (see lines 62-63 in `internal/entity/photo.go`). The previous `ORDER BY photos.photo_path ASC, photos.photo_name ASC` order could be served by that index. Since `NATURAL_SORT_KEY(col)` is a per-row function, it forces a filesort/temp table over the full result set on every page load. MariaDB docs also warn the generated key can be large.

As a result, sorting by name (the default for folders) with these changes results in significant overhead per request.

PR #5572, opened 2026-05-11
Import MWG XMP face regions from sidecar files

### LINKED ISSUE #554: support Adobe's XMP Face region metadata
Facial recognition could be done with a third-party software before it is implemented in photoprism.
I suggest supporting Adobe's XMP Face region metadata, see [pigallery2](https://github.com/bpatrik/pigallery2) for a reference implementation.

hard2forgetme:
# Import MWG XMP face regions from sidecar files

## Summary

Adds support for pulling face regions out of MWG XMP sidecars, including the `mwg-rs:Regions` / `mwg-rs:RegionList` metadata written by Adobe and similar tools.

I kept this separate from the existing face-detection path. XMP regions are converted into crop areas and stored as `SrcXmp` markers while the sidecar is indexed; the normal face merge logic is left alone.

## Changes

- Parse `mwg-rs:RegionList` face entries from XMP sidecars
- Convert MWG's center-based rectangles to PhotoPrism's top-left crop format
- Handle normalized coordinates and pixel coordinates when `AppliedToDimensions` is present
- Skip regions that are not type `Face`
- Store imported regions as `SrcXmp` face markers through a small helper, without modifying `File.AddFace`

## Tests

```text
go test ./internal/meta -run TestXMP/MwgRegions -count=1
go test ./internal/meta -count=1
go test ./internal/entity -run TestFile_AddFaceRegions -count=1
go test ./internal/photoprism -run 'TestApplyFaceRegions|TestIndex' -count=1
git diff --check
```

On macOS, the entity and photoprism package tests needed Homebrew `libtensorflow` headers plus the usual `CPATH` / `LIBRARY_PATH` setup before the TensorFlow bindings would compile. Once that was in place, they passed cleanly.

## Notes

This is deliberately small so it does not interfere with the normal face detection flow. Existing behavior is unchanged unless an XMP sidecar contains MWG face regions.

Fixes #554.

### DIFF
--- a/internal/entity/file.go
+++ b/internal/entity/file.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/photoprism/photoprism/internal/ai/face"
 	"github.com/photoprism/photoprism/internal/config/customize"
+	"github.com/photoprism/photoprism/internal/thumb/crop"
 	"github.com/photoprism/photoprism/pkg/clean"
 	"github.com/photoprism/photoprism/pkg/fs"
 	"github.com/photoprism/photoprism/pkg/media"
@@ -809,6 +810,41 @@ func (m *File) AddFace(f face.Face, subjUid string) {
 	// Append marker if it doesn't conflict with existing marker.
 	if markers := m.Markers(); !markers.Contains(*marker) {
 		markers.AppendWithEmbedding(*marker)
+	}
+}
+
+// AddFaceRegions adds face markers from image metadata.
+func (m *File) AddFaceRegions(regions crop.Areas) {
+	if m.FileUID == "" || m.FileHash == "" {
+		return
+	}
+
+	for _, area := range regions {
+		if area.Empty() {
+			continue
+		}
+
+		marker := Marker{
+			FileUID:       m.FileUID,
+			MarkerType:    MarkerFace,
+			MarkerSrc:     SrcXmp,
+			MarkerName:    area.Name,
+			MarkerReview:  area.Name == "",
+			MarkerInvalid: false,
+			SubjSrc:       SrcXmp,
+			FaceDist:      -1,
+			X:             area.X,
+			Y:             area.Y,
+			W:             area.W,
+			H:             area.H,
+			Size:          int(float32(m.FileWidth) * area.W),
+			Thumb:         area.Thumb(m.FileHash),
+			MatchedAt:     nil,
+		}
+
+		if markers := m.Markers(); !markers.Contains(marker) {
+			markers.Append(marker)
+		}
 	}
 }
 
--- a/internal/entity/file_test.go
+++ b/internal/entity/file_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/photoprism/photoprism/internal/ai/face"
 	"github.com/photoprism/photoprism/internal/config/customize"
+	"github.com/photoprism/photoprism/internal/thumb/crop"
 	"github.com/photoprism/photoprism/pkg/clean"
 	"github.com/photoprism/photoprism/pkg/fs"
 	"github.com/photoprism/photoprism/pkg/http/header"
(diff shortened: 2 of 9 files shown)


photoprismci:
Thanks for picking this up. Four things before this is mergeable:

**1. Conflict with #5563.** We're replacing `internal/meta/xmp_document.go` end-to-end with an XPath-based reader (antchfx/xmlquery) in #5563, and that PR explicitly defers face regions as a follow-up. Your `xmpRegions` struct won't survive the rebase — please wait for #5563 to land and re-implement the parsing side as a `chainXPath` accessor on the new engine.

**2. Route through `Marker.SetName` instead of constructing markers by struct literal.** Our `SrcPriority` system in `internal/entity/src.go` is the right gate for the name-override decision (XMP at 32 sits above AI sources and embedded metadata, below `manual`/`vision`/`admin`), and `SetName` → `SyncSubject` → `Subject()` already lazy-creates or reuses a `Subject` row via `FirstOrCreateSubject` (which does a `FindSubjectByName` first). None of that runs in the current PR because `AddFaceRegions` slams `MarkerName` straight into a struct literal — `SubjUID` stays empty and the imported name never becomes a recognized Person, which is the second acceptance criterion in #554. Going through `SetName` is the prerequisite; the open questions in #4 then determine what fields to set or clear around that call.

**3. Resolve coordinate overlap with existing markers.** Right now if AI already detected a face at those coordinates, `Markers.Contains` returns true and the XMP marker is silently discarded — so the XMP name never reaches the existing marker, regardless of priority. That's the bug. The fix needs a "find overlapping" helper next to `Contains` in `internal/entity/markers.go` so the importer can locate the existing marker; what it does with that marker (whose geometry, whose `MarkerSrc`, how to reassign vs. rename the Subject) depends on the design questions in #4.

**4. Conflict cases the implementation needs to consider:**

1. **AI-assigned name vs. XMP.** Existing marker has `MarkerSrc=SrcImage`, `SubjSrc=SrcAuto` (face detector drew the box, face clustering matched it to a known cluster). XMP arrives with a different name.
2. **Manual name vs. XMP.** Existing marker has `MarkerSrc=SrcImage`, `SubjSrc=SrcManual` (face detector drew the box, user typed the name in PhotoPrism). XMP arrives with a different name.
3. **Untouched marker, name matches an existing Person.** Existing marker has `MarkerSrc=SrcImage`, `SubjSrc=""` (face detector drew the box, never named). XMP says "Jane Doe" and a Subject with that name already exists in PhotoPrism.
4. **Untouched marker, name is new.** Same starting state as above but no matching Subject exists.
5. **Rejected face vs. XMP.** Existing marker has `MarkerSrc=SrcImage`, `SubjSrc=SrcManual`, `MarkerInvalid=true` (user explicitly rejected the region as not-a-face). XMP says the same coordinates are a face.

Two cross-cutting questions apply across most of the cases above:

- **Geometry on overlap.** Whenever XMP overlaps an existing marker, the boxes don't always line up exactly (Lightroom/digiKam boxes are usually tighter than face-detector boxes). Whose `X/Y/W/H` wins, and does `MarkerSrc` flip from `SrcImage` to `SrcXmp` if XMP's box is taken?
- **Reassign vs. global rename.** Whenever XMP wins on name and the existing marker is already linked to a Subject (e.g. "John Doe") and the new name differs (e.g. "Jane Doe"), the current `SetName` → `SyncSubject` path calls `subj.UpdateName("Jane Doe")` on the John Doe Subject — renaming John globally across every photo, rather than reassigning just this marker to a different Person.

One more note: the embedded-XMP path (exiftool JSON) also doesn't read MWG regions — `RegionList` / `RegionInfoRegions` arrive empty in the JSON fixtures and are ignored. Most users have face metadata in the JPG itself rather than in standalone sidecars, so the feature only really lands once that path is wired up too.

Happy to review again once #5563 is in and the above is addressed.

photoprismci:
The review above listed four items that needed resolving before this could be merged — rebasing the parsing onto the XPath reader from #5563, routing names through `Marker.SetName` so they resolve to actual people, handling overlap with existing markers instead of discarding the region, and the conflict cases in point 4. The branch hasn't been updated since, and the embedded-XMP path was never covered.

XMP face region import has since been implemented in #5713 (merged 2026-07-22, tracked in #5712) and shipped in `260728-bbde8f452`, which closed #554. It covers MWG-RS *and* Microsoft `MP:RegionInfo` regions, standalone sidecars as well as embedded XMP packets, name assignment through the `SrcXmp` priority path, and reconciliation against existing markers. It is opt-in via `--xmp-faces` (`PHOTOPRISM_XMP_FACES`), which our bundled `compose.yaml` sets for new installations.

Closing as superseded. Thanks for the contribution.

Issue #639, opened 2020-11-23
[Suggestion] Add a count for Move/Index?


nowjon:
Hey, 

I wanted to suggest a count for how many pictures are done or how many are left when Moving or Indexing. Currently, it's very hard to discern how long it will take to import the items. 

Thanks!

lastzero:
Would require a two pass scan. Also processing RAWs for example takes much longer than indexing small JPEGs. Making a predicting here is more art than science, so we left this todo open for later.

erahhal:
Perhaps a parallel dynamic progress would make it more responsive, e.g. two threads, one crawling the directory tree and updating the model for how long it should take, while another is doing the indexing, so no need to hold up indexing with two serial passes.  People are used to seeing progress and estimated time change significantly towards the beginning of a job.

lastzero:
You could use the average time per file for estimates. Just need to query the files table. Takes time until we can eventually implement this in the UI. Feature requests growing faster than our budget.

JohannesRudolph:
Would a minimal version that just shows the number of files/MiBs left to process in the import dir (as opposed to a time estimate) be a reasonable way to cut the scope? Alternatively show some aggregate stats about the "current import batch" like how long has the importer been running for, how many images/MiBs has it processed already as this gives some user feedback on progress as well.

lastzero:
It would still be a two-pass scan, checking all folders. Depending on the file system and disk performance, this could cause unwanted side effects for some users, so it probably won't be done and tested in 10 minutes. We will be happy to revisit this once the multi-user and in-app upgrades are ready.

javixuwi:
maybe we can start showing the number of indexed files slash total files. The time is important, but more important some feedback of what is going on at the server.

lastzero:
That would require a two pass scan, meaning it will be slower. Also it's possible that files or directories are added after indexing has started, meaning the number of files would be wrong in this case.

RedShiftLight:
This sounds like a perfect opportunity for a switch to turn on the proposed feature. If you don't mind a 2 pass scan, then feel free to enable it. Else, no progress bar. I thought about the suggestion to count in 1 thread while indexing with other threads and was in favor of it until I thought about the read/(and possibly write) operations happening on the disk. I agree with @lastzero in that I think it would be simpler and faster overall to do counting before processing. In the space for the progress bar, you can put a cynical message saying "Counting files to create your damn progress bar!".

[...]

PR #5586, opened 2026-05-16
Metrics: Add import and index progress counters

gshaowei6:
Related to #639.

## Summary
- track the active import/index worker run state, processed file count, and processed bytes
- expose those values through the existing Prometheus-compatible /api/v1/metrics endpoint
- keep default import/index metric series available before the first run
- avoid a two-pass scan, total file count, or ETA calculation so the feature has minimal filesystem overhead

## Tests
- git diff --check
- go test ./internal/photoprism -run TestWorkerProgress -count=1 (not run: Go toolchain is not available in this Windows environment, Docker Desktop is not running, and downloading the official Go toolchain was blocked by local TLS/network policy)

### DIFF
--- a/internal/api/metrics.go
+++ b/internal/api/metrics.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/photoprism/photoprism/internal/auth/acl"
 	"github.com/photoprism/photoprism/internal/config"
+	"github.com/photoprism/photoprism/internal/photoprism"
 	"github.com/photoprism/photoprism/internal/photoprism/get"
 	reg "github.com/photoprism/photoprism/internal/service/cluster/registry"
 	"github.com/photoprism/photoprism/pkg/http/header"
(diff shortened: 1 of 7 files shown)


photoprismci:
Thank you for your interest in PhotoPrism. However, this pull request appears to be fully AI-generated, and the description states that `go test ./internal/photoprism -run TestWorkerProgress` cannot be run in your environment. Contributions whose author is unable to build and test the project locally are not acceptable — we need every change to be verified end-to-end by a human contributor before it is submitted.

The change also doesn't address #639 as requested — the issue asks for a *progress count* visible during import/indexing so the user can tell how many items are left. This PR explicitly skips total-file counting and ETA calculation, and surfaces the running tally only via Prometheus on `/api/v1/metrics` — not in the web UI where users actually see imports run.

Per [CONTRIBUTING.md](https://github.com/photoprism/photoprism/blob/develop/CONTRIBUTING.md#submitting-pull-requests):

> You are welcome to use AI tools for research, drafting, or other supporting tasks. However, **we do not accept fully AI-generated pull requests** at this time. Every contribution must be carefully reviewed, tested, and understood by a human contributor who can discuss the implementation details and maintain the change afterward.

We appreciate the effort, but we ask that contributions reflect genuine human understanding and ownership of the code being submitted.

Issue #5591, opened 2026-05-19
Feature: build-time disable of tensorflow


lapo-luchini:
### Before You Continue

- [x] This is not a support request, and I verified that no similar issues exist

### User Story

As a FreeBSD user, I'd like to be able to disable tensorflow at build time, as currently the recent py-tensorflow port does not include a .so anymore, and the old one fails like this:

ld-elf.so.1: Shared object "libgrpc.so.49" not found, required by "libtensorflow.so.2"

I tried to RTFS but failed to find a way to disable it at build time. Is it possible?

### What problem does this solve, and why would it be valuable to many users?

Allow FreeBSD users to access their photo library, although without face recognition and similar features, it is better than nothing, until tensorflow library can be built again.

### What solution do you propose?

Add a build-time tag to avoid the dependency. (if it isn't already there)

### Which alternatives or workarounds have you considered?

I tried pkg-locking an old version of the libtensorflow library, but it recently broke.

### Open Questions

```markdown

```

### Acceptance Criteria

```markdown
- [ ] photoprism should build, with tensorflow features disabled
- [ ] photoprism should run without any tensorflow library installed
```

PR #5609, opened 2026-05-24
Build: Add notf tag to build without TensorFlow dependency

lapo-luchini:
### Description

Adds a "notf" Go build tag to optionally compile PhotoPrism without libtensorflow.so runtime requirement for platforms like FreeBSD (which currently lacks it).

These changes implement/fix/improve...

#### Related Issues

- #5591

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -410,6 +410,8 @@ build-js:
 	(cd frontend &&	env BUILD_ENV=production NODE_ENV=production npm run build)
 	(cd frontend && node scripts/precompress.js)
 build-go: build-develop
+build-notf:
+	go build -tags notf -o $(BINARY_NAME) ./cmd/photoprism
 build-develop:
 	rm -f $(BINARY_NAME)
 	scripts/build.sh develop $(BINARY_NAME)
--- a/internal/ai/classify/model.go
+++ b/internal/ai/classify/model.go
@@ -1,3 +1,5 @@
+//go:build !notf
+
 package classify
 
 import (
--- a/internal/ai/classify/model_external_test.go
+++ b/internal/ai/classify/model_external_test.go
@@ -1,3 +1,5 @@
+//go:build !notf
+
 package classify
 
 import (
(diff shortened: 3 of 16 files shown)


photoprismci:
Thanks for this, and apologies for the slow response.

We've now built and tested the branch on our side, and the `notf` tag does what it sets out to do: `go build -tags notf ./cmd/photoprism` succeeds, the resulting binary does not link `libtensorflow`, and `github.com/wamuir/graft/tensorflow` is correctly absent from the dependency graph under the tag while remaining present in a default build.

One thing we checked and want to note as *not* a problem, in case it looks like one to a future reader: `internal/ai/tensorflow/info.go` carries no build tag and imports graft, which appears at first glance to defeat the tag. It doesn't — that file imports only graft's generated `*_go_proto` packages, which are pure Go. Leaving it untagged is correct.

We're reviewing the rest now and will follow up with specific comments.

### Where this fits

Some context on our side, since it affects how we'd like to take this forward.

We're planning to move PhotoPrism's remaining TensorFlow models to ONNX Runtime, which we already ship for face detection. That covers three consumers — image classification (labels), NSFW detection, and face embedding — and the last of those is already underway via the configurable face model work in #5758.

That matters for #5591 specifically: ONNX Runtime is available on FreeBSD, so once the migration lands, affected platforms get the **full** feature set rather than a build with classification, NSFW detection, and face recognition stubbed out. It removes a ~386 MB shared library from our images as a side effect.

So we see the two as complementary rather than competing:

- `notf` makes TensorFlow **optional at build time**, and unblocks affected platforms now.
- The ONNX migration **replaces** TensorFlow, so the features keep working everywhere.

The trade-off with `notf` as a bridge is that it introduces a build-tag variant for each TensorFlow-touching file, and those stubs would be removed again once the migration completes. That's a real but modest cost against unblocking users who currently can't build at all, and the migration is not close to done.

If you'd rather put your time toward the ONNX side instead of maintaining the stubs in the meantime, we'd genuinely welcome that — but there's no expectation either way, and we're not asking you to redo work you've already finished.

We'll come back to you shortly on the branch itself.

PR #5610, opened 2026-05-24
Fix old repo source

ghost:
darktable repo is set to Ubuntu 25.10 while container has shifted to 26.04. It tries to install failing dependencies as a result.

This makes the repo source correctly 26.04 again.

### DIFF
--- a/scripts/dist/Makefile
+++ b/scripts/dist/Makefile
@@ -66,8 +66,8 @@ install-clitools:
 	@apt-get -qq install zsh nano >/dev/null 2>&1 && echo "init: successfully installed zsh and nano" || echo "init: packages zsh and nano not available for installation"
 	@apt-get -qq install exa duf >/dev/null 2>&1 && echo "init: successfully installed exa and duf" || echo "init: packages exa and duf not available for installation"
 darktable:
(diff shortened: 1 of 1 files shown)


lastzero:
Our Docker image is based on Ubuntu 26.04 LTS, which was recently released and thus already includes the recent dependencies. 

This Makefile is for advanced users only. The specific target was intended for updating Darktable on old base images. It should not be used by default. If it turns out to be a footgun, we'll remove it.

ghost:
@lastzero this is the latest docker container with the "darktable' command in init

If you do `cat /etc/apt/sources.list.d/graphics\:darktable.list` inside the latest docker container you will see that it is still using the 25.10 repo.

image

ghost:
@lastzero see problem?

image

lastzero:
"darktable" must not be in `PHOTOPRISM_INIT`. We once added it as a workaround for an old base image. It is not intended/suitable  for the current release.

ghost:
Alright, then perhaps it will be better to update the code to ignore the now redundant flag? Because there will be a lot of old configs around running that in the init.

lastzero:
Yes, I agree. It would be best to remove it completely. 🔧

ghost:
Not needed as author removing deprecated option

PR #5635, opened 2026-06-02
Allow manual country override when coordinates exist

### LINKED ISSUE #678: Unable to change country for photos
Some of my photos get incorrectly located/tagged with a wrong country and I cannot change the country for these photos through the web interface

The photos in question were shot in Crimea, Ukraine before 2014, and while most of them are correctly identified as Ukraine, some have 'russian federation' location and label, which is wrong (even if one chooses to ignore the internationally recognized borders, there was no any 'russian federation' there before the occupation/annexation in 2014)

Though the main issue here is that photoprism doesn't allow to change the country in this case

While I'm gonna try to change it directly in the DB as a workaround, I believe there should be (at the very least) a possibility to change the country manually through edit menu

Thank you very much

DoView1:
## Summary
- Allows explicit manual/batch Country overrides while existing coordinates remain set.
- Preserves an explicitly submitted Country value when location resolution runs because coordinates changed.
- Leaves coordinate and Time Zone behavior unchanged; this only removes the Country-only lock and keeps that manual correction from being overwritten on save.

## Validation
- npm run test -w frontend -- tests/vitest/component/photo/edit/details.test.js tests/vitest/component/photo/batch-edit.test.js
- npx eslint --cache frontend/src/component/photo/edit/details.vue frontend/src/component/photo/batch-edit.vue frontend/tests/vitest/component/photo/edit/details.test.js frontend/tests/vitest/component/photo/batch-edit.test.js
- node --check frontend/tests/acceptance/acceptance-public/batch.js
- git diff --check
- go test ./internal/entity ./internal/photoprism/batch -run 'TestSavePhotoForm|TestSavePhoto' -count=1 (not run in this PowerShell session: go.exe is not on PATH; WSL attempt is blocked by missing libvips pkg-config metadata and TensorFlow C API headers)

Closes #678

### DIFF
--- a/frontend/src/component/photo/batch-edit.vue
+++ b/frontend/src/component/photo/batch-edit.vue
@@ -721,10 +721,8 @@ export default {
       return latData?.mixed || lngData?.mixed;
     },
     isCountryReadOnly() {
-      if (!this.formData || !this.formData.Lat || !this.formData.Lng) {
-        return false;
-      }
-      return !!(this.formData.Lat.value || this.formData.Lng.value);
+      // Country corrections should not require clearing existing coordinates.
+      return false;
     },
     // canViewLabels gates the Labels section on the deployment's
     // `labels` feature flag and the session's `labels:search` grant —
--- a/frontend/src/component/photo/edit/details.vue
+++ b/frontend/src/component/photo/edit/details.vue
@@ -155,7 +155,6 @@
                 v-model="view.model.Country"
                 :append-inner-icon="view.model.PlaceSrc === 'manual' ? 'mdi-check' : ''"
                 :disabled="disabled"
-                :readonly="!!(view.model.Lat || view.model.Lng)"
                 :label="$gettext('Country')"
                 hide-no-data
                 autocomplete="off"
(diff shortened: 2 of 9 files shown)


lastzero:
So now you can edit the metadata of pictures of the Empire State Building and claim they were taken in Canada? Can you also set a custom time zone for New York, like Europe/Berlin?

DoView1:
Thanks, that's a fair concern.

With this branch, the coordinates are not changed; the change is an explicit manual/batch metadata override of `PhotoCountry`. The provenance is still marked as manual/batch (`PlaceSrc` is set by the existing single-photo model update path when Country changes, and by the batch save path for batch edits). `UpdateLocation()` still owns the derived place/cell data when coordinates change; if the same save explicitly sets Country, the branch restores that explicit Country value after resolution.

I read #678 as requesting this kind of manual correction for cases where reverse-geocoded country data is undesirable for otherwise valid coordinates. It can produce inconsistent metadata, so your concern is valid. The current Time Zone field is already manually editable independently of coordinates; this PR does not add a new time-zone override, it only removes the Country-only lock and preserves explicit Country edits on save.

If you prefer coordinate-derived place data to remain authoritative, I can narrow this down, e.g. keep the UI lock and only preserve Country when an API/batch request explicitly sends it, or switch to a more explicit manual override affordance instead of making the Country autocomplete generally editable.

PR #5651, opened 2026-06-10
Search: correct logic for photo searching to handle public and private=true

keif888:
### Description

These changes fix:
- searching for both Public and Private photos (setting public=true and private=true)
- searching for Private photos with specific Quality

Previously searching via the UI with a search of **private:true** resulted in all Public and **Review** photos being returned.
Searching via the UI from the private page for **quality:4** would return all private photos, ignoring the quality search request.

Unit tests are included to cover the changed conditions, with some improvements to ensure that other tests did not regress.

### Acceptance Criteria


photoprismci:
Thanks for digging into the search filters, and for including tests! Before we change the behavior here, I'd like to make sure we're fixing the right layer, because both of these conditions look intentional and removing them may regress other cases.

### 1. The `&& frm.Private == false` quality guard

This guard interacts with defaulting that happens one layer up, in `searchPhotosForm` (`internal/api/photos_search.go`):

```go
if frm.Scope == "" && settings.Features.Review &&
   acl.Rules.Deny(acl.ResourcePhotos, s.GetUserRole(), acl.ActionManage) {
    frm.Quality = 3
}
```

For **non-manager** users, the API *forces* `Quality = 3`. The guard exists so that forced floor is **not** applied on the private view — i.e. restricted users can see all of their private photos, including low-quality / unreviewed ones. If we drop the guard, `quality >= 3` starts applying on the private view for those users, and their low-quality private photos silently disappear.

For an **admin** (who has `ActionManage`), `Quality` is *not* force-defaulted, so an explicitly typed `quality:4` on the private page genuinely is ignored today — which I think is the case you hit. That's a real limitation, but the fix probably needs to distinguish "user explicitly typed a quality" from "role/frontend forced `quality=3`" (e.g. a separate signal), rather than keying off `Private`, so we don't regress the restricted-user case.

### 2. `public=true` + `private=true`

The Private page route already sends private only — `frontend/src/app/routes.js`: `staticFilter: { private: "true", public: "" }` — so in the normal UI flow `Public=false, Private=true` and the current `else if frm.Private` branch returns only private photos. The both-`true` combination only arises from a contradictory query (e.g. typing `public:true private:true` together), which the current code resolves deterministically (public wins).

### A note on the tests

The new tests call `Photos(f)` directly at the entity-search layer with a nil session, which bypasses both `searchPhotosForm` (the `Quality=3` / `Public=false` defaulting) and the role-based forcing in `photos.go` (`Deny(AccessPrivate)` → `Public=true, Private=false`). So they validate the SQL builder in isolation rather than the UI/role behavior being reported. `SearchForPrivateQuality` sets `Quality=3` explicitly and asserts it filters, but in the running app `Quality=3` on a private view is exactly the forced default the guard is meant to ignore — so that assertion encodes the opposite of the intended behavior for restricted users.

### Could you share the exact request params?

To pin the root cause, could you grab the actual query parameters from the browser's Network tab for both scenarios?
1. The `private:true` search that returned public/review photos — I suspect `public` is also being sent as `true` here.
2. The `quality:4` search on the private page.

That'll tell us whether the right fix is in the frontend (not sending contradictory `public`+`private`) or in the API-layer defaulting, rather than in the entity-layer filter. Thanks again!

keif888:
1. public:true, private:true from demo.photoprism.app
Did this send me down a rabbit hole?
https://github.com/photoprism/photoprism/blob/7c6546ce19db5ead369636bf5464357584bac810/internal/form/search_photos.go#L48
```
	Public      bool      `form:"public" notes:"Excludes private content"`
	Private     bool      `form:"private" notes:"Finds private content only (except when public:true)"`
```

See the UI call copied as a curl request below.  The query is populated with private:true, with a parameter of public=true.  Parameter quality=3 is being provided from the UI to the API, but quality < 3 photos are being returned.  This indicates that there is a logic problem in the backend, rather than the front end.
```
curl 'https://demo.photoprism.app/api/v1/photos?count=156&offset=0&merged=true&country=&camera=0&lens=0&label=&latlng=&year=0&month=0&color=&order=added&reverse=false&q=private:true&public=true&quality=3' \
  -H 'X-Auth-Token: 234200000000000000000000000000000000000000000000' \
  -H 'sec-ch-ua-platform: "Windows"' \
  -H 'Referer: https://demo.photoprism.app/library/browse?view=cards&order=added&q=private:true&public=true&quality=3' \
  -H 'Accept-Language: ru' \
  -H 'sec-ch-ua: "Google Chrome";v="149", "Chromium";v="149", "Not)A;Brand";v="24"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'X-Client-Uri: https://demo-cdn.photoprism.app/static/build/app.20cf7362e71267753971.js' \
  -H 'User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/149.0.0.0 Safari/537.36' \
  -H 'Accept: application/json, text/plain, */*' \
  -H 'X-Client-Version: 260614-b24e84f5f-Linux-AMD64-Plus' \
  -H 'DNT: 1'
```
2. quality:5 (I used 5 because it shouldn't return anything in demo.photoprism.app)
See the UI call copied as a curl request below.
The UI is sending both the query with quality:5 and the parameter quality=3.  Parameter private=true is sent through.
This one could be a combination UI and Backend issue, as the quality=3 parameter could interfere with the quality:5 in the query string.  I haven't traced through the code to see which wins in a query string and parameter value clash.
```
curl 'https://demo.photoprism.app/api/v1/photos?count=156&offset=0&merged=true&country=&camera=0&lens=0&label=&latlng=&year=0&month=0&color=&order=added&reverse=false&q=quality:5&public=&quality=3&private=true' \
  -H 'X-Auth-Token: 234200000000000000000000000000000000000000000000' \
  -H 'sec-ch-ua-platform: "Windows"' \
  -H 'Referer: https://demo.photoprism.app/library/private?view=cards&order=added&q=quality:5&public=true&quality=3' \
  -H 'Accept-Language: en' \
  -H 'sec-ch-ua: "Google Chrome";v="149", "Chromium";v="149", "Not)A;Brand";v="24"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'X-Client-Uri: https://demo-cdn.photoprism.app/static/build/app.20cf7362e71267753971.js' \
  -H 'User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/149.0.0.0 Safari/537.36' \
  -H 'Accept: application/json, text/plain, */*' \
  -H 'X-Client-Version: 260614-b24e84f5f-Linux-AMD64-Plus' \
  -H 'DNT: 1'
```
Because I hadn't looked at the request before I hadn't noticed that the UI is also sending a quality through on both the calls above.

internal/api/photos_search.go
```diff
@@ -39,11 +39,12 @@ func searchPhotosForm(c *gin.Context) (frm form.SearchPhotos, s *entity.Session,
 		frm.Public = false
```
keif888, line 39: Not sure that this should be frm.Public = false, but it does work.
Logically it should be frm.Private = false, which also works.
I didn't fix it as it would extend the scope of this PR.

keif888:
@photoprismci 
re: A note on tests.
Non Manager roles.  The community edition doesn't support the additional roles, so it is not possible to test these cases via the API.

I have added tests to cover the questions above (although directly via the UserSearch), and fixed the handling so that non manager private queries will perform the same as before.
I did this by adding a new non published (via swag) form field NonManager, which is populated in  searchPhotosForm and used in the query generation.

photoprismci:
@keif888 Thanks for the follow-up and the added tests, and sorry this sat for so long! We've re-reviewed the changes against the current `develop` branch:

### Blocking

**Low-quality private pictures would no longer be shown to admins.** The *Private* page does not define a `quality` value in its static filter (see `frontend/src/app/routes.js`), so `frontend/src/page/photos.vue` adds `quality=3` by default. Previously, the backend ignored the quality filter when `private=true` was set. With this change, it is applied to managers and admins, so private pictures with a quality below 3 are no longer returned on the *Private* page. They don't appear under *Review* either, since that page sends `public=true`. As a result, these pictures would only be found through a manual search.

The unit tests don't catch this, because they call the search functions directly rather than using the parameters the frontend sends. A possible fix would be to add `quality: "0"` to the static filter of the `private` route (as the `all` route already does), along with a test covering these request parameters.

**Merge conflicts:** `develop` has since changed `frm.Private == false` to `!frm.Private` in `internal/entity/search/photos.go` and `internal/entity/search/photos_geo.go`, which conflicts with the lines this PR changes. Could you please rebase or merge the latest `develop` branch?

### Suggestions

- **`NonManager` form field:** Instead of setting a hidden form field in the API handler, `search.UserPhotos` could derive this from the session it already receives, e.g. `sess.Denies(acl.ResourcePhotos, acl.ActionManage)`. That keeps the logic in one place and doesn't rely on every caller setting the flag.
- **Consistency with geo search:** In `photos_geo.go`, the `!frm.Private` condition is removed entirely, while `photos.go` retains it for non-managers. As a result, the same user could get different results in *Places* and *Search*. It would be good if both behaved the same.
- **Fixture change:** `PhotoQuality` of a shared photo fixture changes from 3 to 4. Since fixtures are used throughout the test suite, please also run the full test suite (e.g. `make test-go`) after the rebase, not just the search tests.

Thanks again for looking into this!

PR #5652, opened 2026-06-11
Index: Serialize indexing of byte-identical files to avoid duplicates #5607

### LINKED ISSUE #5607: Index: Duplicate photos created when identical files are indexed in parallel
As a PhotoPrism user indexing photos, I want byte-identical files to be reliably detected as duplicates regardless of how many indexer workers are running, so that my library does not contain redundant entries for the same image.

**Description**

The indexer's duplicate check (internal/photoprism/index_mediafile.go:98-118) performs a hash lookup followed by an insert with no mutex, transaction, or unique constraint on files.file_hash. With PHOTOPRISM_INDEX_WORKERS > 1 (the default is cores/2), two workers processing copies of the same file can both miss the lookup before either commits, and both proceed to create a Photo. Reproducer: place three byte-identical copies of the same image in three subfolders and run photoprism index. Expected: 1 Photo + 2 duplicate-table rows. Actual (non-deterministic): 1–2 Photos + 1–2 duplicate-table rows per image.

**Acceptance criteria**
- Indexing N byte-identical files (same SHA-1) produces exactly one Photo and N−1 rows in the duplicates table, independent of PHOTOPRISM_INDEX_WORKERS.
- Result is deterministic across repeated runs on the same input.
- A regression test covers the parallel-import case (≥ 2 workers, ≥ 2 identical files imported in the same indexing pass).
- Indexing behavior for non-duplicate files is unchanged.
- Setting PHOTOPRISM_INDEX_WORKERS=1 is no longer required as a workaround.

knowald:
Fixes #5607

### Problem

The indexer's duplicate check performs a hash lookup in `UserMediaFile()` and inserts the resulting Photo and File rows roughly 950 lines later, with no synchronization and no unique index on `files.file_hash` (a unique index is not an option because existing databases contain duplicate and empty-string hashes). With `PHOTOPRISM_INDEX_WORKERS > 1`, workers processing byte-identical files in different folders can all miss the lookup before any of them commits, creating one Photo per copy instead of one Photo plus duplicate records.

### Fix

A process-wide, reference-counted per-hash mutex (`lockFileHash` in `internal/photoprism/index_filehash.go`) is held from the hash lookup until `UserMediaFile()` returns, so byte-identical files are indexed strictly one at a time. The waiting worker then finds the committed File row by hash and takes the existing `AddDuplicate` path, returning `IndexDuplicate` exactly as sequential indexing does. Map entries are removed when the last holder releases, so memory use stays bounded on large libraries. Files with an empty hash (unreadable) skip the lock. The import flow is covered as well, since `ImportWorker` indexes through the same `UserMediaFile` call.

Locking is always outermost relative to the entity-layer mutexes, and each goroutine holds at most one hash lock, so no new lock-ordering cycles are introduced. For libraries without duplicates the overhead is one map operation and mutex pair per file. Indexing remains parallel across distinct files; only identical content serializes. Runs that previously required `PHOTOPRISM_INDEX_WORKERS=1` as a workaround can use the default worker count again.

Not covered (unchanged behavior): two separate processes sharing one database can still race, and near-duplicate stacking continues to rely on `photo.Merge`.

### Tests

- `TestIndex_UserMediaFile_ParallelDuplicates` indexes three byte-identical copies concurrently and asserts one `IndexAdded`, two `IndexDuplicate`, and exact row deltas (+1 photo, +1 file, +2 duplicates). It fails deterministically without the lock (3 photos, 0 duplicates) and passes with it.
- `TestLockFileHash` covers serialization under contention and map cleanup after release.
- Verified with `go test -race ./internal/photoprism -run 'TestIndex|TestImport|TestUserImport|TestLockFileHash'`, `golangci-lint` (0 issues), and `gofmt`.

### DIFF
--- a/internal/photoprism/index_filehash.go
+++ b/internal/photoprism/index_filehash.go
@@ -1,0 +1,40 @@
+package photoprism
+
+import "sync"
+
+// fileHashLock serializes indexing of files that share the same content hash.
+type fileHashLock struct {
+	sync.Mutex
+	refs int
+}
+
+// fileHashLocksMutex guards fileHashLocks.
+var fileHashLocksMutex sync.Mutex
+
+// fileHashLocks tracks the per-hash locks currently held or awaited by indexing workers.
+var fileHashLocks = make(map[string]*fileHashLock)
+
+// lockFileHash acquires the indexing lock for the given file hash and returns the function that releases it.
+// Entries are reference counted and removed on release, so the map only holds in-flight hashes.
+func lockFileHash(fileHash string) (unlock func()) {
+	fileHashLocksMutex.Lock()
+	l, found := fileHashLocks[fileHash]
+	if !found {
+		l = &fileHashLock{}
+		fileHashLocks[fileHash] = l
+	}
+	l.refs++
+	fileHashLocksMutex.Unlock()
+
+	l.Lock()
+
+	return func() {
+		l.Unlock()
+		fileHashLocksMutex.Lock()
+		l.refs--
+		if l.refs == 0 {
+			delete(fileHashLocks, fileHash)
+		}
+		fileHashLocksMutex.Unlock()
+	}
+}
--- a/internal/photoprism/index_filehash_test.go
+++ b/internal/photoprism/index_filehash_test.go
@@ -1,0 +1,62 @@
+package photoprism
+
+import (
+	"sync"
+	"sync/atomic"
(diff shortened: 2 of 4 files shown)


lastzero:
Thanks a lot! I'll go ahead and merge this :)

lastzero:
You can now test this using our updated preview build:

Issue #5488, opened 2026-03-17
Documents: Add support for viewing multi-page PDFs


brawlins:
### Confirmation

- [x] I checked this request against the roadmap and existing issues

### What Problem Does This Solve and Why Is It Valuable?

As a user with multi-page PDF documents in my library, I would like PhotoPrism to support viewing all pages of those documents. Currently, you can only view the first page of a PDF. 

Since PhotoPrism already supports PDF files and has a "document" file type, it makes sense to enhance that feature by implementing a PDF viewer that supports multi-page documents. 

### What Solution Would You Like?

When you click on a PDF thumbnail, it should display the document in a PDF viewer that supports:

- Page navigation (prev/next and/or jump to page)
- Zoom in/out controls to adjust zoom level for reading
- Scrolling to subsequent pages
- Viewing page thumbnails for quick preview or navigation

The PDF viewer could be loaded in an iframe on the page.

### What Alternatives Have You Considered?

A PDF viewer could be implemented in several different ways. Modern browsers can display PDFs natively, or there are several JavaScript libraries that render PDF viewers:

- Browser's native PDF viewer
- [PDF.js](https://mozilla.github.io/pdf.js/)
- [PDFObject.js](https://pdfobject.com/)
- [EmbedPDF](https://www.embedpdf.com/)
- [WebPDF.js](https://github.com/foxitsoftware/WebPDF.js/)

Each approach will have its own tradeoffs. My guess is that PDF.js would be a good choice. It's maintained by Mozilla and allows control over UI and behavior so you could match existing PhotoPrism styles. 

### Additional Context

Probably the quickest and easiest solution would be to use PDFObject.js. It detects browser PDF support and adds a fallback link if not supported. It simply manages the iframe and uses the browser's native PDF viewer. The drawback is that you cannot control the styling or behavior, and some older browsers may not support PDF viewing. 

Attached is a quick demo file.
[pdftest.html](https://github.com/user-attachments/files/26062078/pdftest.html)

I'm sure the PhotoPrism team would want more control over the styling, so this is probably not a viable option, but could be a quick win to just implement a viewer (MVP) that could later be enhanced.  

I'm happy to help where I can, but I'm not well versed in Go.

brawlins:
I used Claude Code to create a working proof of concept which adds a PDF.js viewer to support multi-page PDFs. 

Here is a draft PR:
https://github.com/photoprism/photoprism/pull/5500

brawlins:
[...]

PR #5655, opened 2026-06-12
PDF: Add inline multi-page document viewer for the lightbox #5488

omerdduran:
Adds an inline, multi-page PDF viewer to the lightbox so documents can be read directly in PhotoPrism instead of only showing their cover page:
- #5488

### Overview

- Opening a PDF in the lightbox now shows an interactive viewer with page navigation (previous/next and jump-to-page), zoom in/out, scrolling through all pages, and a page-thumbnail strip for quick navigation.
- Pages render as they scroll into view, so large documents stay responsive, and the layout adapts for smaller and mobile screens.
- Documents are streamed on demand through the standard, session-scoped Files API, so access follows the same sharing and visibility rules as the rest of a user's library — including shared albums and published links.
- Rendering uses Mozilla's `pdf.js` (`pdfjs-dist`), loaded only when a PDF is opened so it doesn't affect the initial app load.

### Testing & QA

- Open a multi-page PDF from the library and from a shared album/link: page navigation, zoom, scrolling, and the thumbnail strip should all work, on both desktop and mobile.
- Confirm regular photos and videos in the lightbox are unaffected.
- Backend and frontend unit tests plus acceptance tests are included and passing.


lastzero:
**Thanks a lot! It's a solid piece of work.** 👏

The engine/component split (`common/pdf.js` + `pdf-viewer.vue`), lazy `IntersectionObserver` rendering, dynamic import chunking so the base bundle remains untouched, `markRaw` on the `pdf.js` document, cancelable render handles, and the lazy-render layout-shift re-alignment are all thoughtful and correct. It follows the Options-API / translation / per-function-test conventions, and I verified the backend tests and the 8 component tests pass locally.

There's one change required before we can merge (the new `/pdf` endpoint), plus a set of suggestions below.

---

### 🔴 Required before merge: replace the `/pdf/{hash}/{token}` endpoint

The new endpoint serves the full original PDF gated only by the **preview token**, which is the coarse token shared for thumbnails and video streams. That token exists because ``/`<video>` can't send auth headers — but **pdf.js is fetch/XHR-based and *can* send headers**, so we don't need it here. We should instead reuse the same authorization the rest of the Files API already uses, so PDF access is scoped to the session (shared-link scope, private/archived state) consistent with how `GetFile` and photo search already filter results. As a side effect this also closes the `?download=1` path, which currently serves a forced download behind the preview token instead of the download token.

**Backend**

1. **Delete `internal/api/pdf.go`** and remove `api.GetPDF(APIv1)` (and its comment) from `internal/server/routes.go`. Keep `query.DocumentByPhotoUID` + its test — they're good and reused below.
2. **Serve the bytes from the existing `/files/:hash` route** in `internal/api/files.go` via content negotiation — no new route. `/files/<hash>.pdf` already matches `/files/:hash` (the extension is inside the single path segment), so it doesn't conflict with the `/files/:hash/...` sub-routes.
3. **Parse the extension off the *raw* param before cleaning** — `clean.Token` strips `.`, so clean *after* stripping:
   - `ext := fs.FileType(param)` → `fs.DocumentPDF` for `.pdf`, `fs.TypeUnknown` (`""`) for no/unknown extension.
   - `hash := clean.Token(fs.StripExt(param))`.
4. **Reuse `GetFile`'s authorization** for both branches: `s := Auth(c, acl.ResourceFiles, acl.ActionView)` → `s.Abort(c)` → `search.FileVisibleToSession(hash, s)` (404 when out of scope) → `query.FileByHash(hash)`.
5. **Detect "is this already a PDF?" via `file_type`, not `media_type`.** Use `f.Type() == fs.DocumentPDF` (`File.Type()` is just `fs.Type(m.FileType)`). Avoid `media_type`: at the *file* level PhotoPrism files a PDF under `image` (both `BSc-Thesis.pdf` and its `.pdf.jpg` cover carry `media_type = "image"`); the "document" classification lives on the **photo** (`PhotoType`). `file_mime == header.ContentTypePDF` is an acceptable fallback, but `file_type` is canonical and indexed.
6. **Bytes branch** (when `ext == fs.DocumentPDF`): if `f.Type() != fs.DocumentPDF`, resolve the sidecar via `DocumentByPhotoUID(f.PhotoUID)` (415 if none) — this is what lets a request carry *either* the PDF hash or the cover hash. Then stream with `c.File` (keeps Range support), `header.ContentTypePDF`, `Content-Disposition: inline`, and `Cache-Control: private, immutable` (not the shared-CDN immutable header — the response is session-scoped). Keep `?download` → `c.FileAttachment`, now behind the same auth.
   - Worth a one-line comment: serving a PDF for a `.pdf` request whose hash is a *jpg* cover is deliberate (the document *is* the sidecar). If this generalizes later to `.jpg`/`.mp4` originals, those should serve the matching type directly and **not** cross-resolve.
7. **No/unknown extension** → unchanged JSON response (fully back-compatible).
8. Regenerate Swagger (`make fmt-go swag-fmt swag`); the old `/pdf/...` block drops out and the `.pdf` variant is documented.
9. Optional: `rnd.IsSHA(hash)` early-reject for consistency with `GetVideo`.

**Frontend**

1. `frontend/src/common/util.js` — `$util.pdfUrl(hash)`: drop the preview token from the path, return the tokenless `${$config.apiUri}/files/${hash}.pdf`.
2. `frontend/src/common/pdf.js` — `loadPdfDocument(src)`: attach the session token so pdf.js authenticates the request (same mechanism `common/api.js` uses, `X-Auth-Token: session.token`):
   ```js
   import { getAppStorage } from "common/storage";
   // ...
   const params = typeof src === "string" ? { url: src } : { ...src };
   const token = getAppStorage().getItem("session.token");
   if (token) {
     params.httpHeaders = { ...(params.httpHeaders || {}), "X-Auth-Token": token };
   }
   ```
   This keeps `pdf-viewer.vue` unaware of auth. (Public mode still works — `Auth` resolves a public session when no token is present.)

**Tests**

- Move `internal/api/pdf_test.go` to cover the `files.go` bytes branch: Success (`.pdf`), ResolvesFromCoverHash, RangeRequest, DownloadAttachment, NotPDF (415), NotFound (404), the unchanged JSON branch (no extension), and a **new visibility case** — a session that can't see the photo gets `404` (the point of the change). The old InvalidToken case becomes an auth/no-session case. Keep `TestDocumentByPhotoUID`.
- Update `frontend/tests/vitest/common/pdf.test.js`'s `loadPdfDocument` case to assert the `X-Auth-Token` header is passed.

**One decision for us:** `ActionView` vs `ActionDownload` on the bytes branch — `ActionView` lets anyone who can see the document read it inline (like thumbnails); `ActionDownload` honors "downloads disabled". The visibility scope is fixed either way; this is just the "does inline PDF count as a download" call. (Optional `Accept: application/pdf` fallback when no extension is present is fine to add, but let's make the extension the canonical contract.)

---

### 💡 Suggestions

**Should-fix**

- **Drop the `NOTICE` / `frontend/NOTICE` changes from the branch.** Feature branches shouldn't modify them — they're regenerated after merge and only cause needless merge conflicts.
- **Trim the unreachable viewer lifecycle.** The lightbox mounts the viewer with `v-if=… :key="model.Hash"` and never passes `:active`, so the `active` prop + its watcher + `teardown()`-on-deactivate and the `src()`→`reopen()` watcher can't fire in production (the keyed `v-if` remounts instead). Either wire `:active` to something real or drop those paths to cut ~40 lines of untested-in-prod surface.

**Minor nits**

- The control-bar icon buttons use `:aria-label`; `:title` would also give a hover tooltip (Vuetify forwards it as the aria-label too). The `variant="text"` override is fine for the bespoke dark control bar.
- `submitPageInput` fires on both `@keyup.enter` and `@blur`, so Enter-then-blur double-calls `goToPage` (harmless/idempotent, just noting).
- The shared pdf.js worker is module-scoped and never terminated — fine for a single viewer; a one-line comment would help the next reader.
- Small flicker: the lightbox seeds `pdfThumbsVisible: true` before the viewer's `mounted` emits its real value (which can be `false` on small screens), briefly offsetting the prev arrow.

lastzero:
### Round 2 — endpoint refinement + remaining items

Thanks for turning the endpoint refactor around so quickly — the session-scoped `/files` approach is exactly right. One refinement to the **shape**, then a couple of small things.

**Move the inline bytes to a dedicated sub-route instead of a `.pdf` suffix on the hash:**

- **`GET /api/v1/files/:hash/:name`** (the viewer requests `…/file.pdf`). Dispatch on `fs.FileType(c.Param("name"))`; an unknown/unsupported type → `415`.
- **Keep the auth preamble identical:** `Auth(acl.ResourceFiles, acl.ActionView)` → `s.Abort(c)` → `search.FileVisibleToSession(hash, s)` → `404` if out of scope.
- **`:hash` becomes a pure `clean.Token` again** — the extension lives in its own segment, so you can drop the "parse the ext off the raw param before cleaning" handling, and **`GetFile` (`/files/:hash`) goes back to pure JSON** (remove the `ext == fs.DocumentPDF` branch).
- **Inline-only.** Serve with `Content-Disposition: inline` + Range via `c.File`, and **drop the `?download=` branch entirely** — real downloads stay on the existing `/dl` flow. (Verified: the lightbox's download action → `Photo.downloadAll()` already serves the PDF original via `/dl`, so nothing is lost.)
- **Keep** the cover-hash → PDF cross-resolution (`query.DocumentByPhotoUID`) and the missing-file flagging.
- **Frontend:** `$util.pdfUrl(hash)` → `${$config.apiUri}/files/${hash}/file.pdf`.
- **Tests/Swagger:** regenerate Swagger; rename the `TestGetPDF` cases; drop `DownloadAttachment` / `DownloadDisabled` (that path now lives on `/dl`, already covered there); `NotPDF` should now assert `415`; keep `SharedOnlySessionDenied`, `RangeRequest`, `ResolvesFromCoverHash`, `NotFound`.

**Should-fix (still open from round 1):**

- Drop the `NOTICE` / `frontend/NOTICE` additions — feature branches shouldn't touch them.

**New (from the media-nav arrows):**

- The viewer's scroll area binds `@keydown.left.exact.prevent="$emit('media-prev')"` (and `.right`) without `.stop`, so the event bubbles to the lightbox's `<v-dialog @keydown.left.exact="onKeyDown">`, which *also* calls `pswp().prev()/next()` → **two slides per arrow press**. Add `.stop` (also required by best-practices **B3** — every keydown inside the dialog must stop propagation).

**Minor nits:**

- `prevPage` / `nextPage` / `zoomIn` / `zoomOut` are missing the one-line doc comment (F1).
- `minScale` / `maxScale` / `scaleStep` never change — they could be module consts alongside `ThumbScale` (optional).

omerdduran:
Thanks moving it onto its own sub-route did clean things up nicely. All in last commit:

Endpoint:
- Inline bytes now come from `GET /api/v1/files/:hash/:name`. The handler switches on `fs.FileType(c.Param("name"))` and 415s on anything it doesn't recognize.
- Auth preamble is untouched: `Auth(...)` → `s.Abort(c)` → `FileVisibleToSession(hash, s)` → 404 when it's out of scope.
- `:hash` is a plain `clean.Token` again (dropped the parse-ext-before-clean handling), and `GetFile` on `/files/:hash` is back to pure JSON — the `ext == fs.DocumentPDF` branch is gone.
- Inline only now: `Content-Disposition: inline` + Range via `c.File`. Ripped `?download=` out entirely; downloads stay on `/dl`.
- Kept the cover-hash → PDF resolve via `query.DocumentByPhotoUID`, plus the missing-file flagging.

Frontend:
- `$util.pdfUrl(hash)` returns `${$config.apiUri}/files/${hash}/file.pdf`.

Tests / Swagger:
- Regenerated Swagger. Renamed the `TestGetPDF` cases, dropped `DownloadAttachment`/`DownloadDisabled` (lives on `/dl` now), `NotPDF` asserts 415, and kept `SharedOnlySessionDenied` / `RangeRequest` / `ResolvesFromCoverHash` / `NotFound`.

Also:
- Reverted the `NOTICE` / `frontend/NOTICE` additions from round 1.
- Added `.stop` to the viewer's `@keydown.left/.right` so the arrows stop bubbling up to the dialog and double-advancing.
- Added the doc comments on `prevPage` / `nextPage` / `zoomIn` / `zoomOut`.

lastzero:
Thanks a lot! This looks almost ready to merge. Looking at the code, I suggest generalizing the endpoint further by adding SVG as a second type to prove the shape and updating the visitor/guest permissions in all editions to close the ACL gap.

### 1. Make `GetFileBytes` type-generic (data-driven, not one handler per type)

Right now everything except two things is identical across types. Pull those two into a small descriptor keyed by `fs.Type`, and keep a single streaming path — adding a type becomes a map entry, not a new `getFileX`:

```go
// ServableType describes how GetFileBytes streams a file type inline. Exported
// (with exported fields) so an API subpackage can construct and register its own
// served types.
type ServableType struct {
    ContentType string
    Disposition string                 // defaults to "inline"
    SetHeaders  func(c *gin.Context)   // optional per-type response headers (see SVG)
    // Resolve maps the request to the file actually served. It MUST return a file
    // of the SAME photo that passed the visibility check — see the guard below.
    Resolve func(f *entity.File) (*entity.File, bool)
}

func serveSelf(want fs.Type) func(*entity.File) (*entity.File, bool) {
    return func(f *entity.File) (*entity.File, bool) {
        if f.Type() == want { return f, true }
        return nil, false
    }
}

// PDF cross-resolves the cover hash to the document original (same photo).
func resolvePDFDocument(f *entity.File) (*entity.File, bool) {
    if f.Type() == fs.DocumentPDF { return f, true }
    if doc, err := query.DocumentByPhotoUID(f.PhotoUID); err == nil && doc.Type() == fs.DocumentPDF {
        return doc, true
    }
    return nil, false
}

var servableTypes = map[fs.Type]ServableType{
    fs.DocumentPDF: {ContentType: header.ContentTypePDF, Resolve: resolvePDFDocument},
    fs.VectorSVG:   {ContentType: header.ContentTypeSVG, SetHeaders: svgSafeHeaders, Resolve: serveSelf(fs.VectorSVG)},
}
```

`GetFileBytes` then looks up the spec, keeps the auth/visibility preamble unchanged, resolves the target, and hands off to one shared streamer. **Two things this must get right — they're the reason it's a struct, not an inline `switch`:**

- **Enforce the same-photo scope invariant, don't just document it.** `FileVisibleToSession` authorized the *request* hash, but a resolver returns a possibly *different* file. `resolvePDFDocument` stays within `f.PhotoUID`, so it's safe today — but make it safe-by-construction so future resolvers can't leak past the visibility gate:
  ```go
  target, ok := spec.Resolve(f)
  if !ok || target.PhotoUID != f.PhotoUID {
      AbortEntityNotFound(c); return
  }
  ```
- **SVG is an XSS vector; PDF isn't.** `image/svg+xml` served inline same-origin can execute embedded `<script>`. That's exactly what the per-type `SetHeaders` hook is for — SVG opts into a neutralizing policy, PDF needs nothing:
  ```go
  func svgSafeHeaders(c *gin.Context) {
      c.Header(header.ContentSecurityPolicy, "default-src 'none'; style-src 'unsafe-inline'; sandbox")
      c.Header(header.ContentTypeOptions, header.PolicyNoSniff)
  }
  ```
  (`header.ContentTypeSVG`, `ContentSecurityPolicy`, `ContentTypeOptions`, `PolicyNoSniff` all already exist.)

Adding SVG as the second `servableTypes` entry is the validation that the abstraction holds. Tests: add an SVG `Success` case plus one asserting the CSP + `nosniff` headers are present on the SVG response; the PDF cases stay as-is.

### 2. The visitor 403 is an ACL gap — close it instead of hiding the affordance

In my earlier note I suggested gating the frontend so share-link visitors don't hit a 403'ing viewer. Having looked closer, the cleaner fix is the backend: **grant `RoleGuest` and `RoleVisitor` a view-only grant on `ResourceFiles`.** It's safe, and here's why I checked before recommending it:

- `ResourcePhotos`/`Videos`/`Albums` already use `GrantDefaults`, which gives `RoleGuest: GrantReactShared` and `RoleVisitor: GrantViewShared`. `ResourceFiles` was hardcoded to Admin/Client-only — it predates `search.FileVisibleToSession`, so the blanket lockout is now stricter than the sibling resources without reason.
- `ActionView` on `ResourceFiles` reaches **only** `GetFile` (JSON) and `GetFileBytes` (bytes), and **both already enforce `FileVisibleToSession`** — which scopes a visitor to their shared albums + published links, excluding private/archived. So the grant exposes nothing a visitor can't already see as thumbnails.
- `GrantViewShared` = `{AccessShared, ActionView, ActionDownload}` — no `AccessLibrary`/`ActionManage`/`ActionUpload`/`AccessAll`. I traced every non-handler branch on `ResourceFiles`: `Features.Files` needs `AccessLibrary`, `Features.Library`/`Import` need `ActionManage`, and the `photos.go` search predicate needs `AccessAll` — none flip. And `Auth` still ANDs the session scope check on top, so nothing is bypassed.

So:
```go
ResourceFiles: Roles{
    RoleAdmin:   GrantFullAccess,
    RoleGuest:   GrantViewShared,
    RoleVisitor: GrantViewShared,
    RoleClient:  GrantFullAccess,
},
```
Apply the same two lines to the `ResourceFiles` blocks in the `internal/auth/rules.go` file of each edition too — the editions do `acl.Rules = Rules` (whole-map replace), so a CE-only edit won't reach them. (The editions already grant `RoleViewer` file access; guest/visitor are the missing ones.)

Tests: flip the `ResourceFilesRoleVisitorActionDefault` case in `acl_test.go` (it currently asserts `DenyAll(...ActionView)` is true), and add a handler test that a shared-album visitor **can** fetch a PDF in scope (the positive counterpart to `SharedOnlySessionDenied`, which should now use a session that genuinely lacks scope).

With that grant in place, the frontend needs no capability gate — the viewer just works in shared albums. Keep the graceful error state for the genuinely-out-of-scope 404.

Happy to review once these land; the endpoint generalization + the ACL alignment are both small, and the SVG type is a good regression guard that the abstraction stays generic.

omerdduran:
Thanks! Both items are in the last commit.

**1. GetFileBytes is now data-driven, one streaming path**
- Added a `ServableType` descriptor (ContentType, Disposition, optional SetHeaders, Resolve) and a `servableTypes` map keyed by `fs.Type`. Adding a type is now a map entry, not a new handler branch — the old switch/getFilePDF split is gone.
- The same-photo scope invariant is enforced, not just documented: after `spec.Resolve(f)` the handler rejects with 404 unless `target.PhotoUID == f.PhotoUID`, so a resolver can never return a file that slipped past `FileVisibleToSession`.
- `resolvePDFDocument` keeps the cover-hash → PDF cross-resolution (same photo); `serveSelf(want)` covers types whose original is its own primary file. One shared `serveInlineFile` streamer handles missing-file flagging, Content-Type, `Cache-Control: private, immutable`, per-type headers, and inline `c.File` (Range).

**2. SVG added as the second type (the abstraction's regression guard)**
- `fs.VectorSVG` → `serveSelf`, plus `svgSafeHeaders` via the SetHeaders hook: `Content-Security-Policy: default-src 'none'; style-src 'unsafe-inline'; sandbox` and `X-Content-Type-Options: nosniff` to neutralize the inline-SVG XSS surface. PDF sets no extra headers.
- Swagger regenerated (added `@Produce image/svg+xml`).

**3. Visitor/guest ACL gap closed**
- `ResourceFiles` now grants `RoleGuest` and `RoleVisitor` `GrantViewShared` in `internal/auth/acl/rules.go`. Since both handlers already enforce `FileVisibleToSession`, this exposes nothing a share-link session can't already see as thumbnails, and the frontend needs no capability gate.

**Tests**
- Added `SVGSuccess` and `SVGSafeHeaders` (asserts CSP + nosniff), plus unit tests for `serveSelf` and `resolvePDFDocument` (PDF / cover / unrelated).
- Added `SharedAlbumVisitorSuccess` (positive counterpart to the old denied case: a visitor whose shared album holds the document gets the PDF).
- `acl_test.go`: flipped the visitor case from `DenyAll(...ActionView)` and added the guest case, asserting the grant reaches view/shared/download but not FullAccess/AccessAll/AccessLibrary/ActionManage/ActionUpload.

One decision to flag: with the resolver design, a `.pdf` request for a photo with no PDF now returns 404 (`GuestOutOfScopeNotFound/NotPDF`) rather than 415 — the type is supported, the representation just doesn't exist; genuinely unknown extensions still return 415 (`UnsupportedType`). Happy to switch back if you'd rather keep 415 there.

lastzero:
Reviewed the latest state on a merge with current `develop` — ready to merge from my side. 🎉

- **Endpoint** — `GET /api/v1/files/:hash/:name` streams inline bytes through the same session-scoped path as the rest of the Files API (`Auth(ResourceFiles, ActionView)` + `FileVisibleToSession`), with the same-photo resolver invariant enforced, the SVG CSP/`nosniff` headers in place, and `private, immutable` caching. The old `/pdf` route is gone and Swagger is regenerated.
- **ACL** — the guest/visitor `ResourceFiles` grant now matches the sibling resources; the companion edition PRs carry the same change so shared-album access stays consistent across editions.
- **Tests** — the Go `files` / `acl` / `search` suites and all 233 frontend unit tests pass on the merged tree; `go vet` and ESLint are clean.
- **Behavior** — confirmed a `.pdf` request for a document that has no PDF returns `404`, while a genuinely unknown extension returns `415`.

Really nice work on the endpoint iterations — the data-driven `servableTypes` map came out clean, and adding SVG as a second type is a good guard that the abstraction stays generic. 👍

lastzero (approved):

PR #5661, opened 2026-06-14
Helm: Allow referencing external secret for database password

kurczynski:
### Description

Allow the database password to be sourced from an existing Kubernetes secret
via a new `database.passwordSecretName` value, instead of requiring the
auto-generated `<release>-secrets` secret. This lets users manage credentials
externally (e.g. via Sealed Secrets, External Secrets Operator, or manual
pre-provisioned secrets).

When `database.password` is set directly, it is injected as a plain `value`.
When `database.passwordSecretName` is set instead, the password is read from
that secret via `secretKeyRef`. If neither is set, the env var is omitted
entirely.

### Acceptance Criteria

### DIFF
--- a/setup/charts/plus/README.md
+++ b/setup/charts/plus/README.md
@@ -72,6 +72,27 @@ ingress:
 
 Because TLS typically terminates at the ingress or proxy layer, the chart keeps `PHOTOPRISM_DISABLE_TLS` set to `true`. Only enable PhotoPrism’s internal TLS if your cluster design requires end-to-end encryption and you manage the certificates yourself.
 
+## Database Password
+
+You can provide the database password in two ways:
+
+1. **External secret** – set `database.passwordSecretName` to reference a pre-existing Kubernetes Secret. By default the
+   chart looks for a `PHOTOPRISM_DATABASE_PASSWORD` key; set `database.passwordSecretKey` to use a different key name:
+   ```yaml
+   database:
+     passwordSecretName: my-db-credentials
+     # passwordSecretKey: DB_PASSWORD  # optional, defaults to PHOTOPRISM_DATABASE_PASSWORD
+   ```
+2. **Inline** – set `database.password` directly (the chart stores it in its managed Secret and references it via
+   `secretKeyRef`):
+   ```yaml
+   database:
+     password: "changeme"
+   ```
+
+If both are set, `database.passwordSecretName` takes precedence. If neither is set, no password env var is injected (
+suitable for SQLite).
+
 ## Security Tips
 
 - When `adminPassword` is left blank, a random password is generated and stored in `secret/<release>-photoprism-secrets`.
--- a/setup/charts/plus/templates/statefulset.yaml
+++ b/setup/charts/plus/templates/statefulset.yaml
@@ -97,14 +97,19 @@ spec:
             - name: PHOTOPRISM_DATABASE_USER
               value: {{ $dbUser | quote }}
             {{- end }}
-            {{- $dbPassword := trim (default "" .Values.database.password) -}}
-            {{- if ne $dbPassword "" }}
+            {{- if .Values.database.passwordSecretName }}
+            - name: PHOTOPRISM_DATABASE_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Values.database.passwordSecretName | quote }}
+                  key: {{ (default "PHOTOPRISM_DATABASE_PASSWORD" .Values.database.passwordSecretKey) | quote }}
+            {{- else if ne (trim (default "" .Values.database.password)) "" }}
             - name: PHOTOPRISM_DATABASE_PASSWORD
               valueFrom:
                 secretKeyRef:
                   name: {{ printf "%s-secrets" (include "photoprism-plus.fullname" .) }}
                   key: PHOTOPRISM_DATABASE_PASSWORD
-            {{- end }}
+            {{- end}}
             - name: PHOTOPRISM_DATABASE_TIMEOUT
               value: {{ printf "%v" .Values.database.timeout | quote }}
             - name: PHOTOPRISM_DATABASE_CONNS
--- a/setup/charts/plus/values.yaml
+++ b/setup/charts/plus/values.yaml
@@ -156,6 +156,8 @@ database:
   name: "photoprism"
   user: "photoprism"
   password: ""
+  passwordSecretName: ""
+  passwordSecretKey: ""
   timeout: 15
   conns: 0
   connsIdle: 0


photoprismci:
Thanks for the contribution! The external-secret use case is genuinely useful, and the `passwordSecretName` branch + docs render correctly. One blocker before this can merge, though.

### 🔴 Security regression in the inline `database.password` path

The refactor changes the inline case from a Secret reference to a **plaintext env value** embedded directly in the StatefulSet. Rendering the branch with `database.password=s3cret`:

```yaml
# StatefulSet — was a secretKeyRef before this PR:
- name: PHOTOPRISM_DATABASE_PASSWORD
  value: "s3cret"
```

`templates/secret.yaml` still generates `<release>-secrets` containing the same `PHOTOPRISM_DATABASE_PASSWORD`, so the credential now lives in two places — one of them plaintext. This:

- broadens exposure across RBAC boundaries (`get statefulset` is typically granted far more widely than `get secret`), and also surfaces the password in Helm release history, GitOps-rendered manifests, ArgoCD diffs, etc.;
- makes the still-generated Secret redundant for the DB password;
- moves in the opposite direction of the PR's stated security intent.

Suggested fix — keep the inline case on `secretKeyRef` against `<release>-secrets`, and only add the new external branch:

```gotemplate
{{- if .Values.database.passwordSecretName }}
- name: PHOTOPRISM_DATABASE_PASSWORD
  valueFrom:
    secretKeyRef:
      name: {{ .Values.database.passwordSecretName | quote }}
      key: PHOTOPRISM_DATABASE_PASSWORD
{{- else if ne (trim (default "" .Values.database.password)) "" }}
- name: PHOTOPRISM_DATABASE_PASSWORD
  valueFrom:
    secretKeyRef:
      name: {{ printf "%s-secrets" (include "photoprism-plus.fullname" .) }}
      key: PHOTOPRISM_DATABASE_PASSWORD
{{- end }}
```

### 🟡 Minor

- **Lost `trim`/`default`.** The old guard was `trim (default "" .Values.database.password)`; the new `ne .Values.database.password ""` drops both, so a whitespace-only value injects a literal whitespace password and an explicit `null` compares unequal to `""`. The snippet above restores it.
- **Hardcoded secret key.** The external branch assumes the key is `PHOTOPRISM_DATABASE_PASSWORD` with no override. Fine for now; a `passwordSecretKey` value would be a nice follow-up.
- **Precedence.** The README says inline wins over `passwordSecretName` — just confirming that's the intended ordering (an explicitly-named external secret arguably has the stronger claim).

Otherwise the external-secret and SQLite/no-password paths both render correctly. Happy to re-review once the inline path is back on a `secretKeyRef`.

kurczynski:
> 🔴 Security regression in the inline database.password path

Fixed!

> Lost trim/default

I would argue that **all** variable logic should be in the application, but this is how it was previously.

> Hardcoded secret key

Fixed!

> Precedence

I left this the same because either way, the password is in the values file. Happy to change it though if you disagree.

photoprismci:
Thanks for the quick turnaround! Unfortunately the inline-vs-external branches got **crossed** in the fix — both `secretKeyRef` bodies now point at the wrong secret, so neither path works as written.

### 🔴 The two `secretKeyRef` branches are swapped

```gotemplate
{{- if ne (trim (default "" .Values.database.password)) "" }}   # inline password set
- name: PHOTOPRISM_DATABASE_PASSWORD
  valueFrom:
    secretKeyRef:
      name: {{ .Values.database.passwordSecretName | quote }}   # <-- external name (empty here!)
      key: {{ (default "PHOTOPRISM_DATABASE_PASSWORD" .Values.database.passwordSecretKey) }}
{{- else if .Values.database.passwordSecretName }}              # external secret set
- name: PHOTOPRISM_DATABASE_PASSWORD
  valueFrom:
    secretKeyRef:
      name: {{ printf "%s-secrets" (include "photoprism-plus.fullname" .) }}  # <-- internal secret
      key: PHOTOPRISM_DATABASE_PASSWORD
{{- end }}
```

Tracing each real use case:

- **Inline** (`database.password=s3cret`, `passwordSecretName` unset): renders `secretKeyRef.name: ""`. The password that `secret.yaml` writes into `<release>-secrets` is never referenced, and an empty `secretKeyRef.name` is rejected by the API server — the pod won't start.
- **External** (`passwordSecretName=my-db-credentials`, `database.password=""`): renders `name: <release>-secrets`, so the user's `my-db-credentials` is ignored — the whole point of the PR. Worse, `secret.yaml` only writes `PHOTOPRISM_DATABASE_PASSWORD` into `<release>-secrets` `{{- if ne $dbPassword "" }}`, so with an empty inline password that key doesn't exist and the keyRef dangles.

So the `name:` values need to trade places. The condition I'd suggested earlier keeps them paired correctly (and lets an explicitly-named external secret win — see the precedence note):

```gotemplate
{{- if .Values.database.passwordSecretName }}
- name: PHOTOPRISM_DATABASE_PASSWORD
  valueFrom:
    secretKeyRef:
      name: {{ .Values.database.passwordSecretName | quote }}
      key: {{ default "PHOTOPRISM_DATABASE_PASSWORD" .Values.database.passwordSecretKey | quote }}
{{- else if ne (trim (default "" .Values.database.password)) "" }}
- name: PHOTOPRISM_DATABASE_PASSWORD
  valueFrom:
    secretKeyRef:
      name: {{ printf "%s-secrets" (include "photoprism-plus.fullname" .) | quote }}
      key: PHOTOPRISM_DATABASE_PASSWORD
{{- end }}
```

### 🟡 Minor

- **`passwordSecretKey` is undocumented.** The template now reads `.Values.database.passwordSecretKey`, but it isn't in `values.yaml` (only `passwordSecretName` was added) or the README. Please add it to both (default `PHOTOPRISM_DATABASE_PASSWORD`).
- **Precedence vs. README.** The README still says "If both are set, `database.password` takes precedence." The snippet above flips that so the external secret wins, which I think is the more defensible ordering for an explicitly-named secret — but either way, the template and the README need to agree. Pick one and align both.
- **Quote the `key:` output** (`| quote`) for consistency with the rest of the env block.

Once the two `name:` references are paired with the right branch, please re-render both cases (`helm template` with each value set) to confirm — happy to re-review after that.

kurczynski:
> The two secretKeyRef branches are swapped

Whoops :facepalm: fixed!

> passwordSecretKey is undocumented

Fixed!

> Precedence vs. README

Fixed!

> Quote the key

Fixed!

### `helm template` results

With values:
```
database:
  passwordSecretKey: some-key
  passwordSecretName: photoprism-secrets
```
Result:
```                                                                                                              
            - name: PHOTOPRISM_DATABASE_PASSWORD                                                                                                                                                                    
              valueFrom:                                                                                                                                                                                            
                secretKeyRef:                                                                                                                                                                                       
                  name: "photoprism-secrets"                                                                                                                                                                        
                  key: "some-key"                      
```

With values:
```
database:
  password: some-password
```
Result:
```
           - name: PHOTOPRISM_DATABASE_PASSWORD                                                                                                                                                                    
              valueFrom:                                                                                                                                                                                            
                secretKeyRef:                                                                                                                                                                                       
                  name: photoprism-secrets                                                                                                                                                                          
                  key: PHOTOPRISM_DATABASE_PASSWORD
```

With values:
```
database:
  password: some-password
  passwordSecretKey: some-key
  passwordSecretName: photoprism-secrets
```
Result:
```
           - name: PHOTOPRISM_DATABASE_PASSWORD                                                                                                                                                                    
              valueFrom:                                                                                                                                                                                            
                secretKeyRef:                                                                                                                                                                                       
                  name: "photoprism-secrets"                                                                                                                                                                        
                  key: "some-key"         
```

photoprismci:
Thanks for the contribution and the thorough follow-ups, @kurczynski! 🌈

I rendered the chart locally with `helm v4.2.0` and `helm lint` passes with all four branches behaving as expected:

- `passwordSecretName` + custom `passwordSecretKey` → `secretKeyRef` to the external secret/key ✓
- `passwordSecretName` only → external secret with the default `PHOTOPRISM_DATABASE_PASSWORD` key ✓
- inline `password` only → managed `<release>-secrets` (existing behavior preserved) ✓
- neither set → no password env injected (SQLite default) ✓
- both set → external secret takes precedence ✓

The change is additive, defaults to `""`, and the README + `values.yaml` docs are clear.

We'll take care of a few small follow-ups on our side **after this is merged**, so no action needed from you:

- Bump the chart `version` in `Chart.yaml` (the publishing workflow needs it).
- When `passwordSecretName` is set, skip storing the now-unused inline password in the managed `<release>-secrets`.
- Minor cosmetic `{{- end}}` → `{{- end }}` and, optionally, surface the new values in `questions.yaml` (Rancher catalog form).

Thanks again!

Issue #5672, opened 2026-06-18
UX: Refine face-marker overlay performance and interaction


lastzero:
**As a user tagging faces in the lightbox, I want the face-marker overlay to stay smooth and predictable while I draw, resize, and remove markers, so that tagging is fast and reliable on both desktop and touch.**

A review of the lightbox face-marker overlay (`frontend/src/component/meta/face/markers.vue`), following the sidebar refinement in #4966, surfaced a set of small performance, correctness, UX, and documentation improvements. None are user-facing bugs today, but together they reduce per-interaction overhead and tighten edge cases in the drag-to-create / resize / remove flows. This issue groups them so they can be implemented and QA-retested as a unit.

### Performance

- Hit-testing (`findMarkerAt`) recomputes every marker's pixel rectangle on every hover move and pointer-down (O(n) per event). Marker pixel rects should be precomputed in a computed keyed on the overlay bounds + markers and reused for both hover and hit-testing.
- The draft / pending rectangle is reassigned to a new object on every `pointermove` during draw / move / resize, re-rendering the SVG each tick. These updates should be coalesced through the existing `requestAnimationFrame` batching (`scheduleUpdate`) or applied as non-reactive attribute updates.
- `updateBounds` performs two `getBoundingClientRect()` layout reads per animation frame during pinch-zoom; the parent rectangle can be cached when only the zoom (not the layout) changed.

### Correctness & UX

- The marker `:key` (`m.UID || m.CropID`) can be `undefined` for a brand-new, unsaved marker, which can collapse keys and cause Vue to reuse or misorder DOM nodes. New markers need a stable, unique client-side key.
- The on-canvas confirm (✓) can emit `create` more than once on a fast double-click before the `busy` lock propagates from the parent. A synchronous in-flight flag should make confirmation idempotent.
- When a slide image is already cached (`img.complete`) at the time the overlay attaches its `load` listener, that listener never fires, so initial marker bounds rely solely on PhotoSwipe zoom / change events. The overlay should trigger a bounds update immediately when the image is already complete.
- The resize-handle hit radius (14px) is larger than the visible handle (r=6), which helps mouse users but is still hard to grab on touch. Handles should be more comfortably grabbable on coarse pointers.

### Documentation

- Several methods lack the required one-line doc comment (e.g. `imageElement`, `updateBounds`, `onPointerDown` / `Move` / `Up`, `toLocal`, `insideBounds`, `clamp01`, `scheduleUpdate`).

### Constraints

[...]

Issue #4966, opened 2025-04-25
Viewer: Make information on the sidebar editable


graciousgrey:
**As a user, I want to edit photo details directly in the lightbox sidebar, so that I don't need to switch to the full edit dialog for routine edits.**

### Background

The lightbox info sidebar previously surfaced metadata in read-only form. Routine edits required closing the lightbox and opening the edit dialog, which interrupts browsing flow. This issue tracks bringing inline editing to the sidebar with the same correctness, keyboard handling, and accessibility as the full edit dialog.

### Scope

Fields and controls that should become editable from the sidebar:

- Inline text fields: Title, Caption, Subject, Artist, Copyright, License, Keywords, Notes.
- Dialog-launched editors: Date & Time, Camera, Location.
- Chip-based collections: People (face markers), Labels, Albums.

### Acceptance Criteria

- [x] **MUST** allow inline editing of Title, Caption, Subject, Artist, Copyright, License, Keywords, and Notes when the user has edit permission and the photo is not in a restricted view.
- [x] **MUST** allow editing the photo's date/time, camera details, and location via dedicated dialogs reachable from the sidebar.
- [x] **MUST** allow adding and removing People (face markers), Labels, and Albums from the sidebar.
- [x] **MUST** keep the lightbox keyboard shortcuts working when the sidebar is open and unfocused, and stay out of the way when the user is typing in a sidebar input.
- [x] **MUST** prompt the user to confirm before discarding unsaved changes when navigating away (closing the lightbox, swiping, or using arrow-key navigation).
- [x] **MUST** make all interactive controls reachable by keyboard (Tab, Enter, Esc, Delete) and provide labels/tooltips that work with screen readers.
- [x] **MUST** persist successful edits and surface a clear error notification when a save fails.
- [x] **SHOULD** match the terminology, case conventions, and visual treatment of the full edit dialog and batch edit dialog.
- [x] **SHOULD** keep network round-trips low when working with labels and albums across multiple photos in the same session.
- [x] **MAY** introduce additional UX refinements (e.g. instant-save semantics, dropdown sort/filter behavior, tooltip wording polish) as follow-ups identified during refinement.

PR #5677, opened 2026-06-22
Lightbox: Precompute marker rects and coalesce overlay updates #5672

omerdduran:
## Summary

Refines the lightbox face-marker overlay (`frontend/src/component/meta/face/markers.vue`) following the sidebar work in #4966. These are internal performance, correctness, UX, and documentation improvements to the drag-to-create / resize / move / remove flows — no visual or behavioral regression to the existing interactions. Implements #5672.

### Performance
- **Precomputed hit-testing** — marker pixel rects are computed once in a `markerRects` computed (keyed on bounds + markers) and reused by hover and pointer-down, instead of recomputing every marker's rect on each pointer event.
- **rAF-coalesced drag updates** — draft/pending writes during draw/move/resize are batched through `requestAnimationFrame`, collapsing multiple pointermoves per frame into a single SVG re-render; pointer-up flushes synchronously so the committed rect reflects the final position.
- **Cached parent rect** — the overlay-root rect is cached across pinch-zoom frames (only the image transform changes), removing one `getBoundingClientRect()` read per `zoomPanUpdate` tick.

### Correctness & UX
- **Stable marker keys** — brand-new unsaved markers get a stable, unique `:key` derived from rect geometry instead of collapsing to an empty UID.
- **Idempotent confirm** — a synchronous in-flight guard prevents a second `create` emit on a fast double-click before the parent's async `busy` lock propagates; it re-enables after the busy cycle so a failed save can be retried.
- **Cached-image bounds** — bounds initialize immediately for already-complete images instead of relying on a `load` event that never fires.
- **Touch-friendly handles** — resize handles use a larger grab radius on touch/pen pointers; the visible handles are unchanged (no visual regression).

### Documentation
- Added the required one-line doc comments to the previously undocumented overlay methods, computed properties, and watcher.

## Testing
- `make test-js` — full Vitest suite green (**1673 passed, 13 skipped**).
- New focused coverage in `markers.test.js`: precomputed hit-testing, rAF coalescing (deferred-frame flush + escape cancel), parent-rect caching, stable keys, idempotent confirm, cached-image bounds, and touch hit radius.
- `make fmt-js`, `make lint-js`, and `eslint` on the test file all clean.

## Acceptance criteria (#5672)
- [x] Hover/hit-testing must not recompute every marker's rect per pointer move
- [x] Stable, unique key for a brand-new unsaved marker
- [x] Confirm emits `create` at most once per pending marker
- [x] Bounds initialize correctly for already-cached images
- [x] Pointer-move updates rAF-coalesced
- [x] Resize handles grabbable on touch / coarse pointers
- [x] One-line doc comments on the overlay methods
- [x] Redundant `getBoundingClientRect()` reads reduced during pinch-zoom
- [x] Existing Vitest passes; new behavior covered by tests

### DIFF
--- a/frontend/src/component/meta/face/markers.vue
+++ b/frontend/src/component/meta/face/markers.vue
@@ -10,7 +10,7 @@
     @wheel="onWheel"
   >
     <svg v-if="bounds" class="p-meta-face-markers__svg" :style="svgStyle" :viewBox="`0 0 ${bounds.width} ${bounds.height}`">
-      <g v-for="m in markers" :key="m.UID || m.CropID">
+      <g v-for="m in markers" :key="markerKey(m)">
         <rect
           class="p-meta-face-markers__rect"
           :class="{
@@ -173,6 +173,7 @@ export default {
     isEditMode() {
       return this.mode === FaceMarkerEdit;
     },
+    // svgStyle positions the absolute SVG overlay over the letterboxed image area.
     svgStyle() {
       if (!this.bounds) {
         return { display: "none" };
@@ -191,6 +192,7 @@ export default {
     rootStyle() {
       return this.hoverCursor ? { cursor: this.hoverCursor } : {};
     },
+    // confirmStyle positions the confirm pill centered below the pending rect.
     confirmStyle() {
       if (!this.pending || !this.bounds) {
         return { display: "none" };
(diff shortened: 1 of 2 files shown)


lastzero:
Nice work on this. One small note on the documentation pass: a few of the new doc comments grew past the one-liner the style guide calls for — `markerKey`, `markerRects`, `scheduleUpdate`, and `queueDrag` run ~4–6 lines each. Per `.claude/rules/code-comments.md`, these should be a one-line "what" plus at most 1–2 follow-up lines for non-obvious "why," with any deeper rationale living in the issue (#5672). The explanations themselves are helpful — just worth trimming the inline versions when you get a chance. Non-blocking.

lastzero:
Well done! I merged your improvements to `develop` for final testing :)

PR #5685, opened 2026-06-23
Index: Re-read updated XMP sidecars during indexing #2260

omerdduran:
This pull request enhances the way XMP sidecar files are handled during incremental indexing in PhotoPrism. The main improvement is that updated XMP sidecars are now automatically detected and re-read during normal incremental passes, allowing external metadata edits (e.g., from darktable or digiKam) to be merged without requiring a full rescan. Manual edits within PhotoPrism remain preserved. Additionally, if a sidecar fails to parse, its modification time is not updated, ensuring that a fixed file will be retried on subsequent passes. Comprehensive tests have been added to verify these behaviors.

**Incremental XMP Sidecar Handling Improvements:**

* XMP sidecars are now re-read during normal incremental indexing if their on-disk modification time is newer than the last parsed value, enabling seamless merging of external metadata changes without a forced rescan. Manual edits (`SrcManual`) are preserved over external changes. If a sidecar fails to parse, its previous modification time is retained so that a corrected file is retried in the future. 

**New Functionality and Tests:**

* Introduced the `xmpSidecarChanged` method to detect when an XMP sidecar file has changed and should be re-read during indexing. (`internal/photoprism/index_sidecar.go`
* Added comprehensive tests to ensure the correct detection and handling of XMP sidecar changes, including scenarios for new, changed, malformed, and ignored sidecar files.


lastzero (commented):
Thanks for this, @omerdduran — genuinely useful feature, and the test coverage (new / changed / unchanged / malformed / manual-preserved / sidecar-path-ignored) is excellent. It applies cleanly on current `develop`, and preserving manual edits via the `Src` priority model is exactly right.

A few things worth addressing before we do a final review + test pass — the first matters most, since this path has to scale to libraries with **millions of files**:

**1. Per-file filesystem cost in the skip path (main concern).**
`xmpSidecarChanged` runs for every already-indexed file on every incremental pass, and `fs.SidecarXMP.FindAll` does ~4 `os.Stat` probes per media file **even when the photo has no sidecar at all** (plus `NewMediaFileSkipResolve` + `ModTime()` when one exists). The previous fast-skip was cache-only with no filesystem access, so at millions of files this turns a cheap cache check into millions of extra stat syscalls per incremental pass — noticeable on cold-cache or network storage. Could the probe be gated on the `Files` cache, so we only stat photos already known to carry a sidecar and the common no-sidecar case stays cache-only?

**2. Persistently-malformed sidecar → perpetual reprocessing.**
Since `mod_time` isn't advanced on parse failure (intentional, so a fixed file is retried), a broken sidecar makes its photo un-skippable on every pass — re-running the full `UserMediaFile` each time until it's fixed or removed. Fine in small numbers, but worth bounding/logging so a few bad files can't quietly inflate every pass.

**3. Sidecar deletion isn't detected.**
Adding or editing a sidecar re-triggers, but *removing* one leaves the stale sidecar-derived metadata in place until a full rescan. Worth a note in the README bullet.

**Context on recent changes** (these landed after you opened the PR; no conflict with your diff, but good to know):
- `dc:subject` now maps to the **Subject** field (`Details.Subject`), not the Keywords field — it's Adobe's "Keywords" panel, but PhotoPrism's Keywords / Labels / Subject serve distinct purposes. Your tests use `dc:title` / `dc:description` / GPS, so they're unaffected.
- We renamed the internal XMP spec from `media/xmp-sidecar.md` to `media/xmp-metadata.md` (it now covers sidecar **and** embedded XMP, plus an advanced-parsing roadmap for labels and faces). Worth syncing any references when you next touch it — and your change-detection feature is now the spec's "companion change-detection" follow-up.

Happy to run the full suite and do the final review once these are addressed. Thanks again!

omerdduran:
Thanks Michael, you're right that the per-file probe is the thing to fix here. At million-file scale, those stats really add up.

Before I wire up the cache gate, one thing worth flagging, because I think it turns point 1 into a behavior tradeoff rather than just a perf fix.

Gating the probe on the Files cache would drop the new-sidecar case. Right now `xmpSidecarChanged` stat-probes next to the original precisely so that a sidecar created after the photo was indexed (someone's first darktable/digiKam edit) still gets picked up. That's the `NewSidecar` path in the tests, the "adding a sidecar re-triggers" behavior from your point 3. A brand-new sidecar isn't in the cache yet, so "only stat photos we already know carry a sidecar" skips exactly that case, and the new sidecar ends up waiting for a full rescan (same bucket as the deletion gap).

So the real question is just whether new-sidecar-on-incremental is worth keeping:

* Fine to drop it: cache-gating is the simplest fix. Gate the probe on the cache and add a README line noting that a newly added (or removed) sidecar needs a full rescan, right next to the point 3 deletion note.
* Want to keep it: lean on the walk instead. It already reads every directory, so it already "sees" the `.xmp`, so react there rather than re-probing per photo. Photos with no sidecar cost nothing (the walk emits nothing for them), a photo that already has one costs the single stat the walk was doing anyway, and new sidecars still get caught. Roughly:

```go
found[fileName] = fs.Found
if fs.FileType(fileName) == fs.SidecarXMP {
    primary := ind.primaryForSidecar(relName) // photo.xmp -> photo.jpg / .heic / ...
    if primary != "" && !ind.files.Indexed(primary, entity.RootOriginals, fs.ModTime(fileName), o.Rescan) {
        // queue the primary for reindex so the sidecar is merged
    }
    return nil
}
```

* The catch: this lives in the walk/dispatch loop, and godirwalk hits `photo.jpg` before `photo.xmp`, so "queue the primary" can't be a clean inline check. You'd have to collect the changed sidecars first and reindex their primaries afterward, which is a bigger, messier diff.

Either way works for me; it's your call whether the new-sidecar case earns the larger change. Points 2 (bounding malformed-sidecar retries) and 3 (README note) I'll do regardless.

lastzero:
Thanks — your instinct to flag this instead of quietly wiring up the cheap gate was the right call, and your analysis of the tradeoff is correct. Let's go with the walk-based approach (your second option). The reason it wins isn't just "keep the feature": there's no cheap way to learn a **new** sidecar exists except to stat for it per-photo (the cost we're removing) or to see it in the directory listing — and the walk already enumerates every `.xmp` for free (it falls through to `!media.MainFile()` and returns). So the walk is the only design that fixes the per-file stat cost **and** keeps the new-sidecar case. Cache-gating the per-photo probe would win back stats the walk never spends, at the price of the feature's main use case.

One correctness issue to fix before you build it, though — the comparison in the snippet is keyed on the wrong file:

```go
if primary != "" && !ind.files.Indexed(primary, entity.RootOriginals, fs.ModTime(fileName), o.Rescan) {
```

`Indexed(primary, …, fs.ModTime(fileName))` looks up the **primary's** cache entry (whose stored value is the main media file's own mtime) and compares it to the **`.xmp`'s** mtime. Two different files, so they essentially never match → this queues the primary on every pass. And it never settles: reindexing the primary writes the *primary's* mtime back to its cache entry, still ≠ the sidecar's mtime, so every photo that has a sidecar reindexes forever. It's also conceptually off — the primary's mtime doesn't move when the sidecar is edited, so the primary's entry can't be what detects a sidecar change.

The change-check has to key on the sidecar's **own** entry (its relname + its mtime) — which is exactly what `xmpSidecarChanged` already gets right. Carry that comparison over verbatim; use the primary resolution only to decide what to enqueue:

```go
if fs.FileType(fileName) == fs.SidecarXMP {
    // relName/fileName are the .xmp here: compare its OWN cache entry to its disk mtime.
    if !ind.files.Indexed(relName, entity.RootOriginals, fs.ModTime(fileName), o.Rescan) {
        if primary := ind.primaryForSidecar(relName); primary != "" {
            // queue primary for reindex (deferred; see ordering note)
        }
    }
    return nil
}
```

This self-clears the way the current code does: reindexing the primary re-parses the related sidecar and advances the sidecar row's `mod_time`, so the next pass sees it unchanged.

A few more things to get right while you're in there:

- **Resolve the primary from the cache only** — no per-sidecar stat storm. `photo.jpg.xmp` gives you the primary name directly (strip `.xmp`); a bare `photo.xmp` needs a prefix lookup across the candidate media extensions in `ind.files`. Don't call back into `FindAll`.
- **Two-phase, as you noted:** godirwalk hits `photo.jpg` before `photo.xmp`, so collect the changed sidecars during the walk and reindex their primaries after it completes.
- **Dedupe the reindex set** so a photo whose main file *also* changed this pass isn't processed twice.
- **Keep detection scoped to the originals location** next to the file (as `xmpSidecarChanged` does) — don't react to `.xmp` in the sidecar/hidden paths, which indexing never reads.
- **Deletion (your point 3):** the walk already knows every sidecar present this pass, so a removed sidecar is detectable as "a `.xmp` in the cache that wasn't found" — the same machinery. Nice-to-have if it's cheap here; otherwise keep the README note.
- Points 2 (bound/log malformed-sidecar retries) and 3 (README) as you already planned.

Net: right placement, keep the `xmpSidecarChanged` comparison, and the snippet's `primary`-keyed check is the one thing that would bite. Happy to do the final review + full test pass once it's updated.

omerdduran:
Thanks, Michael, for the detailed review and design guidance. I addressed the feedback in the latest commit (`60fcdff`).

  ### What Changed

  - Moved XMP change detection into the existing filesystem walk. The common incremental skip path no longer performs sidecar discovery or additional `stat`
  calls for every media file.
  - Kept the comparison keyed to the XMP sidecar's own Files cache entry and on-disk modification time.
  - Added cache-only primary resolution for both `photo.jpg.xmp` and `photo.xmp` naming conventions, without calling `FindAll` or probing candidate files on
  disk.
  - Scoped detection to sidecars next to originals and excluded disabled media formats.
  - Implemented two-phase processing: changed sidecars are collected during the walk, and their primaries are queued only after the walk completes
  successfully.
  - Deduplicated queued primaries and skipped those already processed because their main file also changed.
  - Preserved incremental detection of both newly added and updated sidecars.
  - Changed malformed-sidecar handling so the parse error and current modification time are recorded. This prevents an unchanged malformed sidecar from
  triggering reprocessing on every incremental pass. Editing the sidecar makes it eligible for another attempt, and a successful parse clears the stored
  error.
  - Preserved manually edited metadata through the existing `SrcManual` priority rules.
  - Documented that incremental sidecar deletion is not supported and that a forced rescan cannot guarantee complete removal of stale XMP-derived metadata
  because some fields do not retain sufficient source information.

  I also expanded the focused tests to cover primary resolution, naming and case variants, disabled formats, deferred processing, new and updated sidecars,
  deduplication, malformed-sidecar retry behavior, manual metadata preservation, and originals-only scope.

lastzero:
Sounds great! I'll do another review when I have time, so we can (hopefully) merge this. 🚀

lastzero:
### Post-merge review & refinements (on `develop`)

Reviewed the merged change end-to-end and applied four small refinements plus doc updates, committed to `develop` (`#2260 #5685`) for QA:

**Refinements**
- Skip XMP sidecar change-detection on **forced rescans** — the main file is reindexed and re-reads its sidecar anyway, so this avoids redundant per-sidecar work at library scale.
- Centralized the main-media extension list as `media.MainExtensions()` (keeps file-type data in `pkg/media`).
- Flattened the sidecar→main-file resolver (dropped the base-name case permutation and its dedup map).
- Renamed the sidecar helpers/vars from `primary` to `main` — the reindex target is the **main media file** (which may be RAW/HEIC), not the JPEG/PNG **primary preview**.

**Validation**
- Unit: full `internal/photoprism` + `pkg/media` packages, `-race` on the indexing walk, `golangci-lint` clean.
- E2E on the dev workspace (MariaDB): full + partial rescan of all originals (no errors); a **new and an edited** `.xmp` next to an existing photo were both picked up on a plain incremental pass and merged (`SrcXmp`), for both the `photo.xmp` and `photo.jpg.xmp` conventions; and the import pipeline stacks an uploaded `.xmp` and applies it over embedded EXIF.

**Follow-up (out of scope here)**
On instances that set a restrictive `PHOTOPRISM_UPLOAD_ALLOW`, web-uploaded `.xmp` sidecars are rejected before import (the upload allowlist filter drops them). Worth either exempting metadata sidecars from that filter or documenting that `xmp` must be added to the allowlist.

PR #5690, opened 2026-06-25
Portal: Add app.vue alias to vitest config for hooks unit tests

omerdduran:
This pull request adds a new alias to the Vite configuration in `frontend/vitest.config.portal.js`, allowing the import of `app.vue` using a simplified path.

Configuration update:

* Added a path alias so that `"app.vue"` now resolves directly to `./src/app.vue`, making imports cleaner and more consistent.


lastzero:
Sorry, I should have merged this PR first. It's fixed now. Thanks anyway! :)

PR #5716, opened 2026-07-08
Media: Render original-format 360° files via FFmpeg v360 #5711

### LINKED ISSUE #5711: Media: Render original-format 360° files (Insta360 .insv/.insp, fisheye DNG) via FFmpeg
As a 360° camera owner, I want PhotoPrism to import my original Insta360 (.insv/.insp) and fisheye RAW (.dng) files, dewarp them to equirectangular with FFmpeg, and view them in the interactive 360° sphere viewer, so that I can use my unprocessed camera originals without first stitching them in proprietary desktop software.

**Summary**

PhotoPrism already detects and renders equirectangular 360° photos and videos in the Photo Sphere Viewer (#5623), keying off the Projection (file\_projection) and Panorama (photo\_panorama) fields. However, this only works for media that has already been stitched to equirectangular by another tool.

This issue extends that support to original, unstitched camera formats, using FFmpeg's built-in v360 filter for the conversion:

- Insta360 .insv — 360° video, stored as dual-fisheye (typically HEVC, plus a gyro/metadata track in an MP4-style container).
- Insta360 .insp — 360° photo, a JPEG carrying side-by-side dual-fisheye image data.
- Fisheye .dng — RAW captures from fisheye/360° rigs whose projection is fisheye, not rectilinear.

The end goal is that a user drops these originals into their import/originals folder, PhotoPrism indexes them, produces an equirectangular derivative with FFmpeg, and opens them in the existing sphere viewer — no manual pre-stitching required.

**Expected Behavior**

1. Recognition. .insv and .insp are registered as media types and are no longer skipped on import; fisheye .dng files are recognized as fisheye-projected rather than ordinary rectilinear RAW.
2. Metadata extraction. Insta360 / fisheye projection metadata (camera make Insta360, projection geometry) is read during indexing and recorded as a projection type.
3. Dewarp to equirectangular (FFmpeg v360). A conversion step transforms dual-fisheye (Insta360) and fisheye (DNG) sources into an equirectangular derivative used for thumbnails and viewing. The original file is preserved untouched.
4. Viewing. The generated equirectangular derivative routes through the existing is360Equirectangular logic into the Photo Sphere Viewer for both photo and video.
5. Search. These files satisfy the existing panorama:yes filter once indexed.

**Rationale**

360° camera owners overwhelmingly have original exports on disk (.insv/.insp/fisheye .dng), not pre-stitched equirectangular files — stitching today requires Insta360 Studio or similar desktop tools. FFmpeg already ships with PhotoPrism and its v360 filter performs the fisheye→equirectangular conversion with no new dependencies, so supporting originals directly removes a manual step at zero extra runtime cost.

**Technical Considerations**

- File-type registry. Add .insv / .insp to pkg/fs/file\_ext.go + pkg/fs/file\_exts.go (video and image respectively). .dng is already registered as ImageDng (pkg/fs/file\_exts.go:34); fisheye detection must be metadata-driven, not extension-driven.
- Projection vocabulary. Extend pkg/media/projection/types.go with at least fisheye and dual-fisheye; ensure projection.Find() maps the new names. These must NOT route to the sphere viewer as-is — only the generated equirectangular derivative does (frontend/src/common/sphere.js).
- Metadata. Insta360 embeds maker-specific tags (and .insv keeps gyro/stitch data in a side stream); ExifTool surfaces some Insta360 fields. Detection should combine maker (Insta360) + extension + projection geometry.
- Conversion engine — FFmpeg v360 (single back-end, three intake paths). One filter performs the geometry for all sources; the difference is only how decoded pixels reach it:
    - .insp — FFmpeg decodes the JPEG directly → v360=input=dfisheye:output=e.
    - .insv — v360=input=dfisheye:output=e after demux. Note the lens layout varies: some models store both fisheye circles side-by-side in one frame (works directly), while newer models (X3/X4) store the two lenses as separate video streams that must be demuxed and combined (hstack)  before v360. Verify the bundled FFmpeg covers the HEVC streams.
    - .dng — FFmpeg cannot develop camera RAW. Develop the DNG via the existing RAW pipeline (libraw / Darktable / RawTherapee) first, then  v360=input=fisheye:output=e (single fisheye) on the developed image.
- FOV tuning. v360 stitch quality depends on ih\_fov/iv\_fov; sensible per-camera defaults should be chosen, with a reasonable fallback for unknown cameras. Note that the correct angle is the one the stored fisheye disc spans, not the lens specification: the disc is inscribed in its half of the frame, so it is smaller than the optics suggest. Measuring seam discontinuity across candidate angles put the optimum at 186-194 on four scenes from two cameras, so the shipped value is 190 rather than the ~204 originally assumed here.
- Quality limitation (accepted). v360 is a geometric remap, not optical-flow blending — the seam where the two fisheye halves meet is navigable but not perfectly blended. This is acceptable for viewing; higher-quality stitching is explicitly out of scope.
- Cost & async. Conversion is CPU/GPU-intensive; it must run as a background conversion job (like existing transcode/convert workers), never inline on request.
- Derivative storage. Decide where the equirectangular derivative lives (sidecar/cache) and how re-indexing/cleanup treats it, so originals remain the source of truth.

**Constraints**

- Original files MUST be preserved unmodified.
- The conversion MUST use the bundled FFmpeg v360 filter — no new heavyweight or proprietary stitching dependency is introduced.
- The feature MUST degrade gracefully: an unrecognized or unconvertible file is indexed as a normal photo/video rather than failing the import.

**Implementation Status**

Implemented in #5716 (squash-merged as `b5dfd8f7`), with follow-up refinements in `91aed7de`, `a0ef7a0c`, and `d4e95e0a`.

Shipped and verified end to end on real captures (three Ricoh Theta Z1 DNGs, an Insta360 ONE RS `.insp`, a ONE RS fisheye DNG in the vertically stacked layout, and a full ONE RS separate-lens video capture):

- `.insp` stills, single-file ~2:1 `.insv`, separate-lens `.insv` pairs (`_00`/`_10`, reconciled into one photo on a forced rescan), and fisheye DNGs developed through the RAW pipeline.
- Raw `fisheye` / `dual-fisheye` values are recorded on the original but never reported as a viewable projection, so only the dewarped derivative reaches the sphere viewer.
- New operator option `PHOTOPRISM_FFMPEG_FISHEYE_FOV` / `--ffmpeg-fisheye-fov` (default 190), used only as a fallback for cameras without a measured angle.

Not covered by this iteration, tracked in #5771:

- **Insta360 X3 / X4 / X5 single-file captures.** That layout stores each lens as a separate video track in one file; such files fall back to a flat video built from the front lens. `ffprobe` reports the stream count and per-stream dimensions, and `[0:v:0][0:v:1]hstack=inputs=2:shortest=1,v360=...` was confirmed to produce a correct 2:1 output from a two-track file — only lens order and roll need a real capture to verify, and no public sample exists.
- **Hardware-accelerated `v360`.** The filter is software-only, so the transcode deliberately falls back to the software AVC encoder whenever a dewarp runs.
- **The lens seam stays visible.** `v360` is a geometric remap without blending, and it assumes an ideal equidistant fisheye while real lenses ship per-unit calibration that neither vendor exposes. This is the accepted quality limitation noted above.

**Acceptance Criteria**

- [x] .insv and .insp files MUST be recognized as media types and indexed instead of skipped.
- [x] Fisheye .dng files MUST be detected as fisheye-projected and assigned a projection type distinct from rectilinear RAW.
- [x] The projection vocabulary MUST include fisheye and dual-fisheye (or equivalent) values, with extraction populating file_projection.
- [x] Indexing MUST produce an equirectangular derivative for supported Insta360 and fisheye-DNG sources via FFmpeg v360, used for thumbnails and viewing.
- [x] The equirectangular derivative MUST open in the Photo Sphere Viewer for both 360° photo and 360° video originals.
- [x] Original files MUST remain byte-for-byte unmodified after import and conversion.
- [x] The conversion MUST run as an asynchronous background job, not inline in the request path.
- [x] Indexed originals MUST satisfy the panorama:yes search filter.
- [x] The conversion MUST rely only on the bundled FFmpeg (no proprietary or non-redistributable dependency).
- [x] Files that cannot be converted MUST fall back to normal import without erroring the indexer.
- [x] Unit/integration tests SHOULD cover type recognition, projection extraction, and the FFmpeg dewarp step with representative fixtures.
- [x] The 16K (fit_15360) thumbnail ceiling SHOULD be reused for the generated equirectangular derivatives.
- [x] A dedicated projection facet/filter (e.g. distinguishing fisheye originals) MAY be added to search.

omerdduran:
Adds support for original, unstitched 360° camera files: Insta360 `.insv` / `.insp` and fisheye DNG. They are indexed, dewarped to equirectangular with the bundled FFmpeg `v360` filter, and opened in the existing Photo Sphere Viewer. Originals are never modified.

## What it does

`.insp` (image) and `.insv` (video) become first-class media types, `fisheye` and `dual-fisheye` join the projection vocabulary, and fisheye DNGs are detected from the camera make plus the frame geometry — conservatively enough not to catch rectilinear DNGs.

Three intake paths, one filter:

| Source | Handling |
|---|---|
| `.insp` still | side-by-side dual fisheye → `v360` → equirectangular JPEG |
| `.insv`, separate lens files (`_00` / `_10`) | `hstack` + `v360` into one equirectangular AVC; the separate records reconcile into a single panorama on reindex |
| `.insv`, single file, ~2:1 frame | `v360` folded into the AVC transcode (hardware encoders bypassed, `v360` is software only) |
| fisheye DNG | developed through the existing RAW pipeline, then dewarped — side-by-side or vertically stacked |

The derivative carries GPano XMP plus an ExifTool JSON sidecar, so the indexer reads `equirectangular` through the normal metadata path. Raw `fisheye` / `dual-fisheye` values are never reported as a viewable projection, so the viewer only receives a compatible file. `panorama:yes` matches, and a `fisheye:yes` filter finds the originals even though the primary file is the derivative. `ffmpeg-fisheye-fov` sets the dewarp angle, with per-camera defaults and a fallback chain.

## Verification

Checked end to end on real captures rather than fixtures alone: three Ricoh Theta Z1 DNGs, an Insta360 ONE RS `.insp`, a ONE RS fisheye DNG in the vertically stacked layout, and a full ONE RS separate-lens video capture. All produce a correct 2:1 equirectangular derivative, flag the photo as a panorama, and open in the sphere viewer.

That work corrected the dewarp angle. It had been taken from lens specifications, but `v360` needs the angle the stored fisheye disc spans, and the disc is inscribed in its half of the frame. At the previous 204 and 200 the remap discarded a wedge of the scene at each seam — a storefront lost its door frames, a subject lost part of a sleeve. Scoring seam discontinuity across candidate angles put the optimum at 186–194 on four scenes from two cameras, so the values are now 190.

## Known limitations

- **Insta360 X3 / X4 / X5 single-file captures are not dewarped.** That layout stores each lens as a separate video track in one file; such files fall back to a flat video built from the front lens. No public sample of a real capture exists, so lens order and roll cannot be verified.
- **The lens seam stays visible.** `v360` is a geometric remap without blending, and it assumes an ideal equidistant fisheye while real lenses ship per-unit calibration that neither vendor exposes. Correcting the angle removes the lost scene content, not the residual misalignment or the exposure step between lenses.
- **GPU `v360` is not implemented**, which the issue lists as optional.
- **Angles for the X-series and ONE X are extrapolated**, since no sample was available to measure them.

## Possible follow-ups

- **Dewarp the X-series stream layout.** `ffprobe` reports the stream count and per-stream dimensions and is already resolved through `conf.FFprobeBin()`; `[0:v:0][0:v:1]hstack=inputs=2:shortest=1,v360=…` was confirmed to produce a correct 2:1 output from a two-track file, and the decision point in the AVC converter takes a third case cleanly. Only lens order and roll need a real capture.
- **Calibrate the angle per file instead of per model.** Rendering six candidate angles at 1024×512 took 942 ms plus 69 ms to score and picked the same angle as a full-resolution sweep — about a second per file, cacheable per camera. This would retire the hand-maintained table and handle the fact that Ricoh calibrates every Theta Z1 individually, so no per-model constant can be exactly right for it. The angle cannot simply be read from the file: both vendors keep their calibration in undocumented or out-of-band form, there is no fisheye equivalent of GPano, and ExifTool's `FieldOfView` tag is a rectilinear computation that is wrong for these files.
- **Derive roll from the gravity vector** exposed by both vendors, instead of a per-model constant, which would also level captures taken with the camera tilted.
- **Hardware-accelerated `v360`**, which needs a `hwdownload` / `hwupload` round trip.

Better stitching stays out of scope, as agreed in the issue: closing the seam properly needs optical-flow blending and the per-lens calibration neither vendor publishes.

Closes #5711 apart from the X-series stream layout noted above.


photoprismci:
Hi @omerdduran — I reviewed this PR on behalf of the team. Here's a summary of my findings.

## Review: Media: Render original-format 360° files via FFmpeg v360

**Overall verdict: ✅ Complete, correct, and safe to merge — pending draft resolution (see below).**

---

### What the PR does

Adds end-to-end support for Insta360 (`.insp`/`.insv`) and fisheye DNG originals:

- Registers `.insp` (ImageInsp) and `.insv` (VideoInsv) as first-class media types
- Adds `Fisheye` and `DualFisheye` projection types to the projection package
- Dewarps dual-fisheye sources to equirectangular via FFmpeg `v360` filter during conversion (in-place for images, via transcode for video)
- Writes GPano XMP metadata + ExifTool JSON sidecar on dewarped derivatives so the indexer reads the equirectangular projection through the normal pipeline
- Adds a `fisheye:yes` search filter and a "Fisheye" nav item (routes to `browse?q=fisheye`)
- New `ffmpeg-fisheye-fov` CLI flag / config option with per-camera defaults (Insta360 X-series: 204°, other Insta360/Ricoh Theta: 200°)
- Correctly redacts raw fisheye-family projections from viewer DTOs so the sphere viewer only receives compatible equirectangular output

### Code quality

- Architecture is clean: fisheye detection is extension-authoritative for `.insp`/`.insv`, metadata-heuristic for DNGs (conservative — won't misdetect rectilinear DNGs)
- `DualFisheyeLayout()` tolerance of ±0.2 from 2:1 is sensible; unknown aspect (0) is treated as compatible so proprietary formats without EXIF still dewarp
- Hardware encoders are correctly bypassed for dual-fisheye video (`v360` is a software filter)
- The dewarp-in-place uses a temp file + atomic rename; deferred cleanup ensures no stray `.dewarp.jpg` files left behind
- `sphereProjection()` correctly redacts fisheye values in both regular photo results and the geo viewer path
- The `fisheyePhotoFilter` subquery mirrors the base query's `media_id IS NOT NULL` scope — correct
- All `#nosec G204` annotations are sound: paths come from validated config/app logic, not user input
- Navigation item matches the exact markup pattern of surrounding items (`nav-panoramas`, `nav-stacks`, etc.)
- CLA check passes, no CI failures

### Test coverage

Comprehensive: unit tests for every new function, edge cases for `DualFisheyeLayout` (known/unknown aspect, square rejection), `FisheyeDng` detection (Insta360, Ricoh Theta, correctly rejects plain "360" in model name and non-DNG files), `MediaProjection` with fisheye redaction, `sphereProjection`, the `v360` filter string, command construction, `fisheyeFov` fallback chain, and a full index integration test for `.insp` → dual-fisheye projection → panorama flag.

### Minor follow-ups (non-blocking, routine post-merge)

- `config-options.md` is auto-generated from `flags.go` — the new `ffmpeg-fisheye-fov` flag will appear there after a `make` docs regeneration pass.
- `swagger.json` should be rebuilt with `make swag` to pick up any search-filter annotations not manually added.

Neither item blocks the merge.

---

The PR is currently marked as a **draft**. Are there any planned changes still in progress, or is it ready to be marked as ready for review? Asking so we don't merge it prematurely. 🙂

lastzero:
@omerdduran Could this adapter resolve some of our issues?
- https://github.com/mistic100/Photo-Sphere-Viewer/pull/1782

lastzero (commented):
Reviewed locally on the branch. **Looks good — I'm happy to merge this.** The feature work is well-factored and conservatively gated, and I verified it end to end here:

- `go build ./...`, `make fmt-go` (no diff), `golangci-lint` (no new findings in touched files)
- `make fmt-swag swag` reproduces byte-for-byte
- Full `internal/photoprism`, `internal/entity/...`, `internal/api`, plus `pkg/fs`, `pkg/media/...`, `internal/ffmpeg/...`, `internal/config` — all pass
- Vitest 1790 passed / 13 skipped, `make lint-js` clean
- The new dewarp tests genuinely execute here rather than skipping (ffmpeg/exiftool present)

Every `MUST` in #5711 is met — I checked the write paths rather than taking it on trust: only sidecar-path derivatives are ever written, conversion stays in the index/convert worker with the video endpoint aborting instead of converting inline, and failures warn and fall back. The X-series stream layout gap is documented honestly, which I appreciate.

Two things to be aware of, neither of which should hold you up — I'll take care of both:

**1. The develop merge in `778065ecb` dropped two tests.** `internal/photoprism/index_mediafile_test.go` loses `TestIndex_MediaFile_ImportFaceTags` and `TestIndex_MediaFile_FacesOnlyRecountsAfterDelete` (117 lines) — the conflict was resolved by discarding develop's side, so the XMP face-import coverage from the earlier face-marker work disappears. Nothing to do on your end; I'll restore them in a follow-up commit. Worth keeping in mind for the next long-running branch, though: it's the second place in this one where develop-side changes were dropped rather than merged.

**2. `$gettext("Fisheye")` in `navigation.vue` hasn't reached `translations.pot`,** so `make gettext-lint` is red and the label would render English in every locale. That's expected for a feature branch — catalog regeneration belongs in its own commit, so I'll run `make gettext-extract` on develop after merging.

Minor items I'll fold into a cleanup pass:

- `Config.FFmpegFisheyeFov` still documents the default as "204 (typical Insta360 X-series lens overlap)", and the `encode` constants describe it as lens overlap — both predate your FOV correction to 190.
- `entity.CameraFisheyeFov` has three branches that all return 190, so the table collapses to "recognized 360° camera → 190". Also note `strings.Contains(model, "one x")` matches `"iPhone X"` — inert today since every caller is gated on fisheye detection, but a trap if the helper gets reused.

And a few follow-ups I'll file rather than block on:

- `TestPhotosQueryFisheye` passes whether the filter matches everything or nothing, and no fixture carries a fisheye-family `file_projection`, so `fisheyePhotoFilter` isn't really covered.
- `MediaProjection`/`MediaInfo` scan all files for `equirectangular` unconditionally, so a stack mixing a 360° file with a ~2:1 non-360° primary could misroute to the sphere viewer. Narrow, and the frontend aspect guard absorbs most of it.
- `reconcileInsta360Photos` clears `deleted_at` on the canonical row when any merged member is un-archived, which silently un-archives a photo the user archived.

Thanks — nice work on tracking the dewarp angle down to a measured value instead of the lens spec, and on writing that up.

lastzero (approved):

PR #5722, opened 2026-07-10
Search: use natural sort for name and title sort order

### LINKED ISSUE #4912: UX: Implement Natural Sorting for Numerically Named Files in Folder View

keif888:
### Description

This PR implements natural sort order for file name and title sorting by using natural sort, so that numerically named files sort in human-readable order (1, 2, 3... 10, 11 instead of 1, 10, 11, 2, 3).
It supersedes #5571, which was merged in as part of this PR.
It is implemented in a DBMS specific way, using MariaDB's NATURAL_SORT_KEY, and a custom Go function in SQLite.  It will need to be implemented in Postgres via an icu collation and order by, with the codebase setup to easily add that capability into the switch statements.

An additional test has been implemented to confirm that natural sorting is happening for Names, and the Titles test updated to check for natural sort results.  
To achieve this 3 photo fixtures were changed to ensure that non natural sorting would result in a different result to natural sorting.  ie. remove the 0 that prefixed a 09 so that it will sort in the "wrong" order.  This did not break any existing unit tests.

There has been no impact or changes to the acceptance tests.

These changes implement/fix/improve...

#### Related Issues

- Fixes #4912 
- This PR supersedes #5571  

### Acceptance Criteria


lastzero:
Thanks a lot! After reviewing the changes, I found two potential issues, as detailed below. I'm not sure if these can be easily resolved, and if it would be worth the time vs. working on the GORM v2 upgrade and other features.

### Compatibility with MariaDB <10.7

`NATURAL_SORT_KEY()` was introduced in MariaDB 10.7.0. However, our current minimum supported version is 10.5.12, and 10.6 is an LTS release with a large user base:
- https://mariadb.com/docs/server/reference/sql-functions/string-functions/natural_sort_key

On any supported MariaDB in the 10.5.12-10.6.x range, the query will fail with "ERROR 1305: FUNCTION ...NATURAL_SORT_KEY does not exist", so every name/title search would return error 500.

### `NATURAL_SORT_KEY` defeats the index

There is a composite index, `idx_photos_path_name` (`photo_path`, `photo_name`), (see lines 62-63 in `internal/entity/photo.go`). The previous `ORDER BY photos.photo_path ASC, photos.photo_name ASC` order could be served by that index.

Both new expressions make the sort non-sargable:
- MariaDB: `NATURAL_SORT_KEY(col)` is a per-row function, it forces a filesort/temp table over the full result set on every page load. MariaDB docs also warn the generated key can be large.
- SQLite: A custom collation invokes a CGo callback per comparison during the sort, which is expensive for large libraries.

As a result, sorting by name (the default for folders) with these changes might result in significant overhead per request. Benchmarking on a set of at least 100,000 photos would be good before moving forward.

Issue #5712, opened 2026-07-01
People: Import Face Markers & Names from XMP


graciousgrey:
**As a user whose photos are already face-tagged in another application, I want PhotoPrism to import those face regions and names from XMP, so that I don't have to tag the same faces again.**

PhotoPrism should read face regions and their assigned names from XMP in the MWG-RS (`mwg-rs:Regions` / `RegionList`) and Microsoft (`MP:RegionInfo`) schemas from both embedded XMP and `.xmp` sidecars, and create face markers for them during indexing. Names link to (or create) the corresponding Person so the face is recognizable as that named person. This is a backend/indexing feature; the only UI surface is an "Import Faces from XMP" option in Settings › Advanced.

**Status:** Implemented in #5713 (merged to `develop`). Import is controlled by the global `PHOTOPRISM_XMP_FACES` option, which is **enabled by default in the bundled config examples** (`compose.yaml` / `setup/`) and defaults to off in code so re-indexing an existing library does not create People without an explicit opt-in.

  ### Design Notes

  - AI face detection runs first so embeddings are captured; each XMP region is then reconciled onto the overlapping AI marker via `Marker.SetName(name, SrcXmp)`. A region with no AI overlap becomes a name-plus-box marker with `MarkerSrc = SrcXmp`.
  - `MarkerSrc` (box origin) and `SubjSrc` (name origin) are separate fields. Reconciled markers keep the AI box and `MarkerSrc = SrcImage`, with `SubjSrc =SrcXmp` carrying the name; only no-overlap regions get `MarkerSrc = SrcXmp`.
  - Source priority: `SrcXmp` = 32 — overrides AI-assigned names (auto / image / onnx / ollama), never manual (64) or admin (128).
  - No clustering and no embeddings in v1: XMP markers carry no embedding, so the recognition worker skips them and an XMP-named person will not auto-match other photos yet.

  ## Conflict Resolution Matrix

  Each case MUST have a dedicated unit test. 

  |#|Existing marker state|XMP supplies|Expected outcome|
  |-|-|-|-|
  |1|No existing marker at those coordinates|Region + name|Create a SrcXmp marker and link or create the Person|
  |2|AI marker, AI-assigned name (SubjSrc = SrcAuto), differs from XMP|Different name|XMP name wins (priority 32 over auto); apply to this marker only via SetName — keep the AI box and MarkerSrc = SrcImage, set SubjSrc = SrcXmp|
[...]

PR #5751, opened 2026-07-24
Faces: Import XMP regions from HEIC and RAW sources

omerdduran:
This PR fixes XMP face-tag import for still-image formats whose metadata source is not the primary indexed JPEG or PNG file.

Previously, XMP face reconciliation only inspected the primary media file. When PhotoPrism used a generated JPEG/PNG preview as the primary file, embedded XMP regions inside HEIC files and XMP sidecars belonging to RAW files were never found, so those face tags were lost.

### Source Discovery

- Collects face regions from the logical still-image source (HEIC/RAW), not just the generated preview.
- Finds both `photo.xmp` and `photo.ext.xmp` sidecars, searching originals, the configured sidecar directory, and `.photoprism`.
- When multiple sidecars exist, prefers the most recently modified one (full-name wins on ties); a present sidecar is authoritative over embedded XMP.
- Applies the imported regions to the photo's primary JPEG or PNG file.
- Limits the behavior to still images; videos, PDFs, and animated images are excluded.

### Reconciliation

- Updates existing XMP markers in place instead of creating duplicates.
- Matches each region to the best-overlapping marker so two nearby faces don't collapse onto one.
- Removes stale markers when a region is deleted from the sidecar, and restores a marker's clustered/auto name (or leaves it for review) when its XMP name goes away.
- Never overwrites manually-assigned names or geometry; source priority is respected.
- Preserves rejected (invalid) markers and never resurrects them.
- Reconciles changed sidecars during incremental indexing, even when the original and primary preview are unchanged.
- Also runs in the deferred vision/metadata worker path, so XMP names import even when face detection is deferred or disabled.
- Treats import as best-effort: a failed XMP read or reconcile is logged without failing the index or discarding AI-detected faces.
- A forced re-index backfills regions for photos indexed before the feature was enabled.

### Parsing Robustness

- Supports rectangular and circular (diameter) regions, region rotation, and all eight EXIF orientations (including mirror/transpose).
- Falls back to the source image's real dimensions and orientation when the XMP has no `AppliedToDimensions`.
- Reads additional writer dialects: MWG-RS (name/title/seeAlso/Extensions PersonInImage), Microsoft MP:RegionInfo, and ACDSee (DLYArea preferred over ALGArea).
- Clips edge-cropped faces to the frame instead of dropping them.
- Handles ExifTool array compaction safely: when optional per-region arrays are sparse, regions are reported as partial and existing markers are left as-is rather than mislabeled or deleted. A malformed sidecar returns an error and never wipes existing markers.
- Adds a scoped `Marker.Delete()` that refuses to run an unscoped delete on an empty UID.

### Testing

- Embedded HEIC XMP regions and RAW/DNG sidecars (`photo.xmp` and `photo.dng.xmp`).
- Forced re-index backfill and incremental sidecar-only updates.
- Marker name and geometry updates without duplicates; stale-marker deletion and face-count recount.
- Manual/rejected marker preservation; malformed-sidecar and partial-parse safety.

Manual indexing tests were performed with embedded and sidecar XMP for JPEG, PNG, TIFF, WebP, AVIF, HEIC, DNG, and CR2. All 12 embedded/sidecar matrix variants imported exactly one XMP marker onto the correct primary file.


lastzero:
Thanks for the fix — the behavior is correct and the integration tests pass locally (`go build ./...`, `golangci-lint`, and the new tests are all green). Two things I'd like to see addressed before merge:

### Test coverage

The new helpers `ApplyXmpFaces`, `isXmpFaceSource`, and `xmpFaceSources` are only exercised indirectly through the end-to-end tests. Per our testing conventions, each new function (including unexported helpers) should have its own focused `Test*` with at least a success and an error/no-op case — e.g. `ApplyXmpFaces` has clear branches (`file == nil` → error, `m == nil`/empty hash → no-op) that aren't covered directly.

### Performance on large libraries

`PHOTOPRISM_XMP_FACES` is enabled by default in the bundled configs, so this runs on every index for standard installs. The concern is that the change recomputes `RelatedFiles(false)` from scratch inside `collectXmpFaces`/`xmpFaceSources` and again in the `IndexRelated` reconciliation, even though indexing already built that group. `RelatedFiles` does two directory globs (`readdir`) plus per-match stats, and the `IndexRelated` path also adds a `PrimaryFile` DB query and an eager `file.Markers()` load per photo — all incurred even for the common case of a photo with no XMP face regions. On libraries with millions of files indexed over CIFS/NFS, those extra `readdir`/`stat` round-trips add up.

Possible mitigations:

- Reuse the already-resolved group: thread `related.Main`/the known sources into `xmpFaceSources` instead of re-globbing.
- Gate the `IndexRelated` reconciliation so it only runs when `related.Main` is a distinct logical source that can carry XMP faces (HEIC/HEIF/RAW), rather than for every image including standalone JPEGs.
- Defer the `file.Markers()` load until after the empty-faces check so photos without XMP faces skip that query.

Happy to help prototype the group-reuse approach if useful.

lastzero:
Thanks for the follow-up commit, @omerdduran — the focused tests for `ApplyXmpFaces`, `isXmpFaceSource`, and `xmpFaceSources` are exactly what was needed, and the full suites are green locally (`go build ./...`, `golangci-lint` clean on the changed files, plus the complete `internal/meta`, `internal/entity`, and `internal/photoprism` packages).

Before we merge, the performance point from my earlier review is still open, and while validating the branch I confirmed two paths where the new authoritative-delete sweep silently removes previously imported face markers. Details and measurements below.

## Blocking

### 1. `collectXmpFaces` still re-globs the directory for every photo

`xmpFaceSources` continues to call `m.RelatedFiles(false)` from scratch in `internal/photoprism/index_faces_xmp.go`, for every primary file on every index — including standalone JPEGs that have no XMP at all. `RelatedFiles` runs **two** `filepath.Glob` calls (each a full `readdir` + sort of the directory) plus a `stat`/`NewMediaFile` per match, and the indexer has already resolved that exact group before it calls `UserMediaFile`.

Measured on a warm local SSD, `xmpFaceSources` + `FindAll` versus the previous `FindFirst` lookup:

| Files in directory | Before | This PR | Factor |
|--------------------|--------|---------|--------|
| 100                | 66 µs  | 255 µs  | 3.9x   |
| 1,000              | 26 µs  | 338 µs  | 12.9x  |
| 5,000              | 32 µs  | 1.57 ms | 48.9x  |

The cost scales with the number of files in the directory, not with whether XMP is present. Over CIFS/NFS the `readdir` round-trips dominate and this gets considerably worse.

There is a second, less visible cost on the embedded path: `collectXmpFaces` calls `source.MetaData()` on the `related.Main` instance that `RelatedFiles` has just created, so its `metaOnce` cache is cold and it performs a full EXIF parse plus ExifTool JSON read of the RAW or HEIC file per photo. For a JPEG + HEIC pair in a two-file directory with warm ExifTool caches, that is 3.97 ms to 7.19 ms, or +3.2 ms per photo.

Threading the caller's already-resolved `RelatedFiles` (or at least `related.Main`) into `collectXmpFaces` / `xmpFaceSources` addresses both: it removes the globs and reuses the `MediaFile` instances the indexer already holds, so the metadata cache stays warm.

### 2. `entity.PrimaryFile(photoUID)` runs before anything establishes there is work to do

In `internal/photoprism/index_related.go` the gate is `!related.ContainsPreview() && isXmpFaceSource(related.Main)`. Generated previews are written to `SidecarPath()` (see `fs.FileName(f.FileName(), w.conf.SidecarPath(), ...)` in `convert_image.go`), so they are not part of the originals glob and `ContainsPreview()` is false when a HEIC or RAW is re-indexed. Every changed or newly added HEIC/RAW photo therefore pays `Db().Unscoped().First(... file_primary = 1 AND photo_uid = ?)`, then `file.Markers()` as a second query, then the full `collectXmpFaces` filesystem cost above — all before anything checks whether the source carries regions.

My earlier suggestion to defer `file.Markers()` is only half applicable now: with authoritative deletes, an empty face set is meaningful, so the markers really are needed. But `PrimaryFile` can move after `collectXmpFaces`, and a cheap "no sidecar present and no embedded regions" short-circuit would skip both queries in the common case.

### 3. A region-less `.xmp` sidecar deletes previously imported markers

`collectXmpFaces` treats any present sidecar as authoritative, never falls back to embedded XMP, and does not distinguish "this sidecar contains no region block" from "the user removed all regions".

Reproduced with `internal/photoprism/testdata/xmp-faces/embedded-mwg.jpg` (one embedded region) plus a well-formed `.xmp` containing only `xmp:Rating`:

```
embedded faces without sidecar: 1
faces with region-less sidecar: 0, partial=false
reconcile changed=1
RESULT: existing SrcXmp marker was DELETED
```

Region-less sidecars are entirely ordinary — darktable and Lightroom develop settings, rating-only sidecars, and so on. PhotoPrism itself passes an existing `.xmp` to `darktable-cli` (`internal/raw/convert.go`), so sidecars beside RAW files are an expected part of the workflow. Please treat a sidecar as authoritative only when it actually contains a region container, and otherwise fall back to the embedded packet.

### 4. MWG regions without an explicit `Type` are dropped, and the empty result is authoritative

Both parse paths were tightened. Previously an absent type was accepted (`if t := ...; t != "" && !EqualFold(t, "Face")`); this PR requires it, in `xmp_document.go` (`if !strings.EqualFold(relFirst(li, relMwgType), "Face")`) and in `json_exiftool.go` (`if !typesAligned || ...`). `mwg-rs:Type` is optional in the MWG specification, so writers that omit it are valid.

Measured against `internal/meta/testdata/faces/mwg-multi.json`:

```
with RegionType:    faces=2 partial=false
without RegionType: faces=0 partial=false
```

Because `jsonArrayFit(nil, n)` returns `(false, false)`, a wholly absent `RegionType` is not flagged partial — so the sweep treats the empty result as authoritative and deletes every marker previously imported from such a writer. Please either restore the earlier leniency (absent type implies a face) or, at minimum, set `partial` when `len(types) == 0 && n > 0` so nothing is deleted.

## Should fix

### 5. One related source's metadata error discards the primary file's own valid regions

In `collectXmpFaces`, the source loop does `if data.Error != nil { return xmpFaceSet{}, ... }`. `meta.Data.Error` is a routine, debug-level condition in `MediaFile.MetaData()`. A RAW or HEIC sibling that fails EXIF parsing therefore kills XMP import for a JPEG primary that parsed perfectly well, and emits a warning per photo. Skipping that source and continuing would be safer.

### 6. `partial` has no equivalent on the sidecar path

`XmpDocument.FacesWithOptions` has no partial concept, so a sidecar whose regions fail to normalize (for example `unit="pixel"` with no `AppliedToDimensions` and no source dimensions) also yields an authoritative empty set and triggers the delete sweep. The ExifTool path's `recordFit` accounting is careful; the XPath path currently has none.

### 7. Documentation is now out of date

Our XMP metadata documentation still states that a marker is *not* auto-deleted when its region disappears, and that overlap reconciliation goes through `Markers.Overlapping` — both of which this PR changes. It also does not yet describe sidecar-over-embedded authority, the partial-parse flag, circular and rotated regions, the full eight-orientation mapping, or the new ACDSee / MWG title / `seeAlso` dialects, and the code map omits `ApplyXmpFaces`, `xmpFaceSources`, `isXmpFaceSource`, `restoreMarkerName`, `Marker.Delete`, `RelatedFiles.ContainsPreview`, `XMPWithOptions`, and `FacesWithOptions`. Since #5712 has "Documentation MUST be updated" checked, please refresh it alongside this change.

## Nice to have

- `restoreMarkerName` has no dedicated `Test*` yet; it is only covered indirectly through the `RemovedXmpName*` subtests. Same class as the gap flagged earlier. (`Marker.Delete` is properly covered with both a success and an empty-UID case.)
- `matched = append(matched, true)` stays index-aligned with `markers.Append` only because `Markers.Append` always appends exactly one element. A one-line comment would protect that invariant.

## What already works well

- Best-overlap binding preserves the `> face.OverlapThreshold` semantics of `Markers.Overlapping` while fixing a real bug where two nearby regions could collapse onto a single marker.
- `Marker.Delete()` correctly refuses an unscoped delete; since `MarkerUID` is the primary key, the delete is row-scoped once that check passes.
- Replacing the silent `log.Warnf` plus `return false` in `applyXmpName` with proper error propagation is a clear improvement.
- Treating XMP import as best-effort in `ApplyDetectedFaces`, so a malformed sidecar cannot discard AI-detected faces, is the right call and is well tested.
- The ExifTool array-compaction analysis is careful, thorough work, and test coverage overall is broad.

Suggested order: items 3 and 4 first, since they are small and localized and prevent silent marker loss, then the group-reuse refactor for items 1 and 2. Happy to help prototype the group-reuse approach if you would rather not carry it in this PR.

omerdduran:
Thanks for the detailed review and the performance measurements. I pushed `0b08089f6` (`Faces: Harden XMP region authority and reuse the file group`) and addressed the remaining points:

- Normal index and import paths now seed each `MediaFile` with the already-resolved group main through `SetRelatedMain`. `xmpFaceSources` reuses that same instance, avoiding the per-photo `RelatedFiles(false)` re-glob and keeping the logical source’s metadata cache warm. Deferred worker paths that do not have group context retain the fallback.
- `IndexRelated` now collects the XMP regions before calling `entity.PrimaryFile`. When no region container is declared, it skips both the primary-file lookup and marker loading. A declared-but-empty region container still proceeds so intentional face removal remains authoritative.
- I introduced `meta.FaceRegions` with separate `Declared` and `Partial` flags. A rating-only or develop-only sidecar is no longer authoritative; it is skipped in favor of the next sidecar that declares regions or the embedded packet. The destructive sweep only runs for a declared, fully resolved region set.
- MWG-RS and ACDSee regions without an explicit `Type` are accepted on both the XPath and ExifTool paths, while explicitly declared non-face types are still skipped.
- An unreadable related source is now skipped instead of discarding valid regions from the other sources. The result is marked partial so it cannot delete unmatched markers.
- The sidecar parser now tracks declared versus successfully resolved regions, so normalization failures also mark the result partial and suppress destructive reconciliation.
- I updated `internal/meta/README.md` and the private XMP specification in `e1625b4`, including source authority, group reuse, sidecar/embedded precedence, partial parsing, supported geometry and dialects, reconciliation behavior, and the expanded code map.
- I also added the focused `TestRestoreMarkerName` coverage and documented the `Markers.Append` / `matched` index-alignment invariant.

All package tests pass, and the changed-code lint reports 0 issues. I’d appreciate another look when you have time.

Issue #4669, opened 2024-12-16
AI: Improve Facial Recognition


graciousgrey:
**As a user of PhotoPrism's face recognition feature, I would like to see a more capable model and vector database being used to improve performance and accuracy of matches.**

Since the face detection and recognition functionality was first implemented in PhotoPrism, there has been a lot of progress in both models and database technology, for example:

- https://mariadb.com/kb/en/vector-functions/

Related Issues:

- https://github.com/photoprism/photoprism/issues/3124
- https://github.com/photoprism/photoprism/issues/4328
- https://github.com/photoprism/photoprism/issues/1587
- https://github.com/photoprism/photoprism/issues/5167

theshadow27:
Questions for discussion/refinement:

1) **Database Support:** [MariaDB](https://mariadb.com/kb/en/vector-functions/) only, or also [sqlite](https://github.com/asg017/sqlite-vec)?
2) **Backwards Compatibility:** All-in (add columns/indexes to existing tables) or incremental (new tables with FK to old tables, e.g. `markers_v`)? Will this be a 'major' release that requires users to upgrade to the latest DB/install extensions in order to move to?
3) **Scope of enhancement:** Drop in to replace current `match` with vector distance search preserving all other logic, or redesign to leverage HNSW indexes with more advanced methods (HDBSCAN, temporal clustering, iterative learning, remove minimization/optimization). Incorporate a time dimension or metric to assist with children/aging?
4) **Architecture Reevaluation:** Keep re-implementing algorithms in Go, or gRPC to a python container dedicated to face recognition, clustering, and matching? This could render the first 3 considerations moot - all existing code remains, with feature flag to use the bypass it when the service is present.

lastzero:
@theshadow27 Thanks for your questions!

1. Since the vector features are only available with the latest/upcoming version of MariaDB, we will/should keep the current implementation for SQLite until it eventually offers the same or similar functionality (I believe there are patches in the works, but feel free to do your own research and report back).
2. As for determining the best upgrade/migration strategy, that would be part of this issue, i.e. we need to first test for feasibility and then work out a plan before implementing the final solution. We could use a different/new database table to store the face embeddings if the same one cannot/should not be used, e.g. for performance reasons. However, adding a second column with the native vector type (or replacing the current one), if the database is a supported MariaDB version, seems easiest to implement.
[...]

Issue #1587, opened 2021-10-03
People: Improve facial recognition for young children


ScuttleSE:
The model used to categorize faces seems to be poorly trained with younger children.

I have a cluster with hundreds of faces ages 0-10, dozens of different people, but Photoprism thinks it's the same face.

lastzero:
Yes, that's the case. Funding will fix this.

lastzero:
Maybe we should exclude children from recognition, so that you need to name every face individually.

lastzero:
Overall, we're happy to have such specific issues 🥳

tootbrute:
> Maybe we should exclude children from recognition, so that you need to name every face individually.

Another idea is make it easier to "undo" a bad match. I have a folder for a child's face. I clicked to see what photos it found of the face. 90% of the photos are the wrong person.

It would be great if I could select all the wrong photos, then go to the action button at the bottom and "clear faces".

Of course, it would be great if it recognized children better, but this might help people clean up their faces database in other instances as well. For myself at this moment, I think I may have to go back and just delete this person to remove all those faces that are wrong.

lastzero:
We've added a special rule for baby faces to the latest update that excludes them from automatic matching / clustering. This won't change already indexed / matched faces though.

tootbrute:
> We've added a special rule for baby faces to the latest update that excludes them from automatic matching / clustering. This won't change already indexed / matched faces though.

Wonderful!

Now, what is the current suggestion on how to clear bad matches or delete a whole person? --> docker-compose exec photoprism photoprism faces index?

I hope I can avoid reindexing everything because with 50,000 pictures, that face matching hammers my server for 24 hours, but I can't think of another way.

I see I can hide a person, but not 'delete' that person. The only way I've found this is to go into each picture and click the 'eject' button on the person. This is not feasible with 100+ pictures.

lastzero:
> Now, what is the current suggestion on how to clear bad matches or delete a whole person? --> docker-compose exec photoprism photoprism faces index?

If you didn't invest too much time in naming faces, you may reset existing faces & people and then ONLY reindex faces:

```
docker-compose exec photoprism photoprism faces reset -f
docker-compose exec photoprism photoprism faces index
```

Skip the `docker-compose exec photoprism` prefix when not using Docker Compose, or you're running the commands from inside the container. You can't perform these operations from the UI.

[...]

Issue #5703, opened 2026-06-28
AI: Upgrade ONNX Runtime from 1.26 to 1.29 with optional GPU support


photoprismci:
**As a PhotoPrism user running machine-learning features, I want ONNX Runtime upgraded from 1.26 to 1.29 with optional GPU acceleration, so that inference stays current with upstream fixes and can run substantially faster on supported hardware.**

PhotoPrism uses ONNX Runtime for ML inference (face detection, embeddings, and related vision tasks) through the cgo binding [`yalue/onnxruntime_go`](https://github.com/yalue/onnxruntime_go). The runtime currently shipped is **1.28.0** (CPU build). Upstream has released [**v1.29.0**](https://github.com/microsoft/onnxruntime/releases/tag/v1.29.0), and the matching archives (CPU + GPU) are already mirrored on our download server.

> **Retargeted 1.28 → 1.29.** This issue was originally opened against 1.27 and then retargeted to 1.28, which has since landed: the binding moved to `v1.32.x`, the install script default was raised to 1.28.0, and the base images were rebuilt against it. Upstream then released 1.29.0, and the binding followed with a release embedding its headers, so the same three-step chain applies once more. The earlier steps are complete and superseded rather than outstanding — the checklist below tracks the 1.29 step.

### Precondition — Go binding compatibility

The default runtime version must not be bumped independently of the cgo binding. The binding embeds a copy of the ONNX Runtime C API header and requests that exact API version at initialization, so a mismatch is not a soft failure — the shared library refuses the requested version and every ONNX code path fails at load:

```
The requested API version [29] is not available, only API versions [1, 28] are supported in this build.
```

- `yalue/onnxruntime_go` **v1.32.1** embeds the **1.28.0** headers (`ORT_API_VERSION 28`).
- `yalue/onnxruntime_go` **v1.35.0** embeds the **1.29.0** headers (`ORT_API_VERSION 29`) — the exact match for this target.

The two therefore move together, and the published base images are what make the pairing real: a green local test run only proves the developer's own machine has the new runtime installed, not that `photoprism/develop:*` carries it.

### Ordering

The change is a chain rather than a single edit, and the middle step is easy to skip:

[...]

PR #5758, opened 2026-07-30
AI: Make the face embedding model configurable #4669 #1587

omerdduran:
Adds a `PHOTOPRISM_FACE_MODEL` option so the face embedding model can be selected without a rebuild, moves embedding inference to ONNX Runtime alongside the existing SCRFD detector, benchmarks the candidates against the FaceNet weights we have shipped since 2021, and makes SFace the model new libraries start with.

Relates to #4669 and #1587.

## Why

On ordinary adult portraits the current model is fine, which is why this has been easy to miss. It falls apart on the two groups the issues describe. Measured on 5310 labeled images with the benchmark added here (TAR at FAR = 0.1 %, higher is better):

| Model | general (LFW) | asian (CASIA-FaceV5) | children (FG-NET 0-12) | cross-age (FG-NET) |
|:---|---:|---:|---:|---:|
| `facenet` (current) | 0.9253 | **0.7938** | **0.2562** | 0.3471 |
| `sface` | 0.9464 | 0.9982 | 0.4934 | 0.5163 |
| `auraface` | 0.9383 | 0.9826 | 0.5051 | 0.5070 |
| `arcface_mbf` | 0.9479 | 0.9996 | 0.5750 | 0.5856 |
| `arcface_r50` | **0.9519** | **1.0000** | **0.7758** | **0.7575** |

All models are indistinguishable on LFW (EER 4.4-4.6 %). Indexing the same images on a dev instance shows what the numbers mean in practice: of the face clusters FaceNet forms on the Asian subset, 67 % merge more than one person, and the largest single cluster merges 138 distinct identities.

**SFace is the choice.** It is Apache-2.0, 39 MB, the fastest of the five at 13.4 ms per face, and its 128-dimensional output quarters the stored embedding payload. AuraFace was added and measured specifically because it is the permissively licensed member of the architecture family that tops the table — it does not recover that advantage, trailing SFace on three of four subsets at eight times the cost per face, so it stays registered as a fallback rather than a recommendation.

## What changed

- **`PHOTOPRISM_FACE_MODEL` / `--face-model`** — `auto`, `none`, `facenet`, `sface`, `auraface`, `arcface_r50`, `arcface_mbf`. Reported by `photoprism config` and `photoprism faces config`.
- **Model registry** (`internal/ai/face/models.go`) carrying each model's input size, channel order, mean/scale, embedding length, alignment mode and weight license, so the CLI help and config report come from one source. The preprocessing constants were taken from the OpenCV and InsightFace reference implementations rather than inferred.
- **SCRFD landmarks are decoded.** The bundled `scrfd_500m_bnkps` model has always predicted five keypoints; they were never read. `internal/ai/face/align.go` now warps them onto the standard 112x112 template. This is not optional for ArcFace-style models: for the same face rotated 20 degrees in plane, the aligned embedding distance is 0.374 versus 0.925 unaligned.
- **`face.Embedder` interface** so TensorFlow and ONNX models share one call site, plus a generic ONNX embedder that refuses to load a file whose output width does not match the registry.
- **Embedding provenance.** `faces.embed_model` and `markers.embed_model` record which model produced each vector, `entity.Face.Match` refuses cross-model comparisons, and `photoprism faces audit` reports the cluster count per model. Columns are added by GORM auto-migration.
- **Per-model thresholds.** `ClusterDist`, `ClusterRadius` and `MatchDist` live on each registry entry, derived by `TestCalibrateFaceThresholds`. One global set fits exactly one embedding space: at FaceNet's accept distance of 0.82, SFace reaches TAR 0.5875 against FaceNet's 0.8185 on the same images, which would have read as "the new model is worse". The shipped FaceNet configuration costs 1.43 % false automatic matches; every calibrated model targets a tenth of that, and still beats FaceNet's true accept rate.
- **`photoprism faces migrate --to MODEL`** with `--dry-run`, which reports valid and invalid markers, manually identified people, and marker counts per model before changing anything. It preserves manual subject assignments, checkpoints regenerated embeddings so it can resume, and atomically replaces clusters before rebuilding automatic matches.
- **Two defects found on the way.** `face.Clusters.Contains` treated the `-1` distance returned for non-comparable embeddings as *inside* the cluster radius, so any vector of an unexpected length was silently classified as a background face and discarded. And the bundled child/background reference samples are FaceNet-space vectors, which the 512-dimensional models would have been compared against purely because the lengths match. Both filters measure inert (see below) and are now deleted, together with `face.Clusters`, whose only consumers they were.
- **Install scripts.** SFace is Apache-2.0 and is now part of `make dep` through `dep-onnx`, so it ships in the published images — a model new libraries default to has to be present. AuraFace is Apache-2.0 as well but its graph is 261 MB, so `scripts/download-auraface.sh` is an explicit opt-in and is wired into no build target; its file is deliberately renamed from the upstream `glintr100.onnx`, because a different model ships under that same name elsewhere and preprocessing cannot be read from an ONNX graph. The InsightFace weights are published for non-commercial research only, so `scripts/download-arcface.sh` requires `ARCFACE_ACCEPT_LICENSE=1` and they are never bundled.

## Default and upgrade behavior

**A library that already holds face vectors keeps the model that produced them.** `auto` asks the library before it consults the preference list, so upgrading an existing installation never changes its embedding space on its own; only a library with no vectors starts on SFace. Moving an existing library is an explicit `FACE_MODEL` change followed by `photoprism faces migrate`.

Two details make that guarantee hold, and both are worth a reviewer's attention:

1. **The schema is migrated after the configuration is propagated.** On the first start after an upgrade, `markers` has no `embed_model` column yet and the provenance query fails. A library that predates the column cannot hold anything but FaceNet vectors, so that case counts face markers instead and resolves to FaceNet. Without this, every existing library would have resolved to SFace on the first start and quietly written into a second, incomparable space.
2. **A resolution found before the database is connected is not cached**, so an early lookup cannot freeze the preference list into place.

Both paths have regression tests, including one that drops the column to reproduce the upgrade.

**Image size grows by 39 MB.** Published images now carry FaceNet and SFace together, which is required for the transition: existing libraries need FaceNet until they migrate. Dropping it is a separate decision for later.

## Error rates at the shipped thresholds

Every model is calibrated to a tenth of the shipped FaceNet error budget, and that holds on adult faces. It does not hold on children, so the number is recorded here rather than left for a reviewer to find.

Measured at each model's own accept distance (`ClusterRadius + MatchDist`: 0.82 for FaceNet, 1.06 for SFace), over the same images, the same detections and the same pairs for both models:

| Model | Subset | Accept | TAR | FAR |
|:---|:---|---:|---:|---:|
| `facenet` | general (LFW) | 0.82 | 0.8296 | 0.0000 |
| `sface` | general (LFW) | 1.06 | 0.9474 | 0.0000 |
| `facenet` | children (FG-NET 0-12) | 0.82 | 0.4323 | 0.0151 |
| `sface` | children (FG-NET 0-12) | 1.06 | **0.8000** | **0.0731** |

On adults both models sit at zero false accepts. On children SFace nearly doubles the true accept rate, and raises false accepts from 30 to 145 over the same 1984 negative pairs.

**This is an open decision, not a settled trade-off.** Two readings, and this PR does not pick between them:

- 7.31 % is above the 1.43 % global budget the shipped FaceNet configuration costs, and roughly fifty times the per-model target the calibration harness aims for. By the standard this PR sets for itself, the children distribution is out of budget.
- FaceNet does not usefully recognize children today at all (0.4323). Trading 30 wrong merges for 115 additional correct ones, on exactly the population #1587 is about, is arguably the better product outcome.

Resolving it means adding a child subset to `TestCalibrateFaceThresholds` and re-deriving the thresholds, which is the same harness work as the `ClusterDist` item below.

Method note: this is a pairwise verification measure, not a clustering simulation. It reports how often two different children fall inside the accept distance, which is what drives a wrong merge, rather than the end-to-end merge rate. FG-NET is small and deliberately hard, so the gap between the two models is the reliable figure rather than the absolute rate.

## Reference sample filters: measured, then removed

`IsChild` and `IsBackground` compared an embedding against bundled FaceNet-space reference samples. Rather than regenerate those samples in SFace space, I measured what they caught under FaceNet, with both filters forced on:

| Filter | Input | Size | Fired |
|:---|:---|:---|---:|
| `IsChild` | FG-NET ages 0-12 | 168 embeddings | **0** |
| `IsBackground` | 54 non-portrait images from `assets/samples` | 5 detections | **0** |
| `IsBackground` | LFW + FG-NET faces | 524 embeddings | 0 *(correct, must not fire)* |

Neither filter fired on anything, so **nothing is lost by removing them: they were already inert under FaceNet.**

`Children` held two reference vectors at radius `ClusterRadius` (0.42), which is tighter than the 0.82 accept distance for the same person, so it can only match a near-duplicate of those two specific children. The nearest of the 168 real children sits at distance 0.540 and the median at 1.030.

`Background`'s 24 samples were collected against Pigo, the detector SCRFD replaced, as its own comment recorded. SCRFD does not produce the plant and sky false positives they describe. The false positives it does produce are animal faces, two pug detections in the sample set, and those land 0.876 from the nearest background sample, whose radius is 0.76.

Regenerating the samples would therefore have reproduced a filter that does not fire, so they are deleted rather than carried forward. Replacing them with something that works — an age signal, or a detector-confidence gate — is a separate decision and out of scope here.

## Changes after review

The review raised four things that had to be resolved before merge. Each is covered by a regression test that fails against the previous commit.

**Face clusters could be built from two embedding spaces.** `Face.Match` guarded only the receiver's provenance, while the vector it was compared against carried none — and FaceNet, AuraFace and both ArcFace variants are all 512-wide, so the length check could not separate them. `Faces.Optimize` could therefore merge clusters from different models into one centroid, and a person-naming click in the UI reached the same path.

Provenance is now derived from the producer rather than from configuration: `face.Face` carries the model that generated its vector, set where the embedder and the vector last coexist, and `NewFace` / `SetEmbeddings` take it as a parameter instead of reading a global. `Match` and `ResolveCollision` take the counterparty's model, `Marker.SetFace` refuses a cross-space assignment, `ManuallyAddedFaces` is filtered, and `query.Faces` gained a `MatchableFaces` variant so the audit keeps its cross-model view. Note that `face.ModelsComparable` is asymmetric — it treats its second argument as the reference — so stored-vs-stored comparisons go through a new `SameEmbeddingSpace` helper.

**`faces migrate` finalized even when it re-embedded nothing.** A file-level read failure was counted per marker and the loop continued, so an unreadable originals mount produced `Migrated: 0` and still ran the finalize, which deletes every cluster row and blanks the embeddings. The error was returned *after* the destruction was committed. The finalize is now refused when nothing was migrated or when the failure ratio exceeds 10 %, overridable with `--force`; markers with no file are reported apart from failures so a library with orphans does not refuse forever.

Related: the skip decision used strict equality against the target, so a library predating the provenance column was never idempotent — `--to facenet` re-embedded everything into the space it was already in. Both the skip and `FinalizeFaceMigration`'s inverse now treat a blank model as FaceNet. These two had to change together: fixing only the skip turns it into skip-then-delete.

**A failed embedder crashed the process.** When the model fails to load, the marker query is deliberately left unfiltered, so mixed-dimension vectors reached DBSCAN — and `alg.EuclideanDist` indexed the shorter operand without a length guard, panicking inside a worker goroutine that no `recover` covers. The clusterers now reject ragged input at the entry point, the distance functions return `NaN` rather than panicking or silently truncating, and clustering refuses to run when the embedder failed to load or the samples differ in length.

**Smaller items.** A configured model whose weights are missing now disables embeddings instead of falling forward into another vector space — which matters because `assets/` is baked into the image, so an upgrade removes opt-in models. Embeddings returned by a remote vision endpoint are validated and never attributed by guesswork. `--face-collision-dist` and `--face-epsilon-dist` warn when out of range instead of falling back silently. And two config tests compared a getter against a mutable package var, which made `make test-mariadb` fail deterministically.

## Still open

- The children false accept rate above is an unresolved calibration question, not a resolved trade-off.
- `ClusterDist` is derived from pairwise distance equivalence rather than from a clustering simulation, so fragmentation and merge behavior are not measured yet.
- `faces migrate` preserves `SrcManual` identities; XMP-sourced names keep their marker but seed no replacement cluster, and are not counted in the summary.
- `--dry-run` cannot predict files that are absent from disk, only markers with no `file_uid`.
- The worker guards around migration are process-local, so a CLI migration cannot see a running server's indexer. A cross-process lock is the real fix.

## Benchmark

Skipped unless a dataset is provided, so normal test runs are unaffected:

```bash
PHOTOPRISM_TEST_FACE_DATASET=/path/to/dataset \
PHOTOPRISM_TEST_FACE_ALIGNED_SUBSETS=asian \
  go test ./internal/ai/face -run TestBenchmarkEmbeddingModels -count=1 -v -timeout 120m
```

Layout is one directory per identity, optionally grouped by subset (`dataset/<subset>/<person>/*.jpg`). Every model scores the same detections and negative pairs are sampled with a fixed stride, so the ranking cannot shift with detector variance and a rerun produces an identical pair set. That held in practice: across a refactor of the harness, all 28 subset accuracy rows reproduced byte for byte and only the wall-clock `ms/Face` column moved.

The datasets used above were obtained from third-party mirrors and are not committed. The Asian subset is distributed as pre-aligned 112x112 crops, which gives the ArcFace-style models exactly their training input while handing FaceNet a tighter crop than its own convention, so part of its deficit there may be crop mismatch. The children and cross-age subsets went through PhotoPrism's own detector and alignment, and FaceNet is still last by a wide margin. Provenance should be re-verified against the official distributions before any of these figures is published outside the project.

## Tests

`go build ./...`, `make fmt-go`, `make check-make-help`, and `golangci-lint` are clean for the lines this branch adds. `go test` passes for `./pkg/vector/...`, `./internal/ai/...`, `./internal/config`, `./internal/entity`, `./internal/entity/query`, `./internal/photoprism`, `./internal/commands`, and `./internal/api`.

The same packages pass against MariaDB. Note that `./pkg/fs/duf` fails in a container where mount enumeration returns nothing; it is unrelated to this branch and fails the same way on `develop`.


lastzero:
Thank you for this — it's a genuinely impressive piece of work, and one of the most carefully prepared contributions we've received in a while. The model registry as a single source for preprocessing contracts, the `face.Embedder` abstraction, the decision to keep `facenet` as the default while the thresholds are still FaceNet-tuned, and the honest "what this does not do" section all made the review much easier. The two incidental bug fixes are valuable on their own: `Clusters.Contains` treating the `-1` non-comparable distance as a match, and `Magnitude()` computing itself via `Dist(NullEmbedding)`. Decoding the SCRFD landmarks that the bundled model has always predicted is a nice catch as well.

I reviewed and ran the branch locally. `go build ./...`, `gofmt`, `shellcheck`, and `make check-make-help` are clean, `golangci-lint` reports zero issues in the files this PR touches, and the Go tests pass for `./internal/ai/...`, `./internal/config`, `./internal/entity`, `./internal/entity/query`, and `./internal/photoprism`. After `make dep-sface` the ONNX embedder tests pass too, including the landmark decoding against the real SCRFD model and the alignment-distance check.

Below are the findings we would want resolved before we can merge. We also agree with your own gating conditions — recalibrated thresholds and regenerated child/background samples — before the default can move, so there is no time pressure on that part.

## Should be resolved before merge

**1. A configured model whose weights are not installed silently falls back to FaceNet with the wrong provenance.**

When `ConfigureEmbedder` fails because the model file is missing, it logs a warning but leaves `configuredModel` set to the requested name while `activeEmbedder` stays `nil`. `vision.Model.FaceModel()` then falls through to the TensorFlow FaceNet path, and `face.EmbeddingModelName()` still reports the requested model. I confirmed the resulting state for `arcface_r50` with the weights absent:

```
ConfigureEmbedder err = faces: stat .../arcface/w600k_r50.onnx: no such file or directory
ActiveEmbedder        = <nil>
EmbeddingModelName()  = "arcface_r50"
ExpectedDims()        = 512
SamplesComparable()   = false
```

Because ArcFace and FaceNet are both 512 wide, the dimension guard in `Face.SetEmbeddings` passes, so `faces.embed_model` and `markers.embed_model` are stamped `arcface_r50` on FaceNet vectors. `SameEmbeddingModel()` will later accept those rows as comparable against real ArcFace vectors. At the same time `SamplesComparable()` returns false, which deactivates the child and background filters — the very safeguard your description identifies as what currently keeps juvenile faces out of automatic matching.

Since the ArcFace weights are never bundled and SFace is not part of `make dep`, this is the most likely first outcome for anyone who sets the option. Resolving to `face.ModelNone` when initialization fails, and/or applying the `Installed()` check in `Config.FaceModel()` to explicit names the way it is already applied for `auto`, would close it.

**2. `onnxEmbedder.Run` dereferences a closed session.**

`Close()` sets `e.session = nil` under the mutex, and `Run()` takes the same mutex and calls `e.session.Run(...)` without a nil check. `onnxruntime_go`'s `RunWithOptions` dereferences the session on its first statement, so this is a nil-pointer panic rather than a returned error. `ConfigureEmbedder` closes the previous embedder, and `Config.Propagate()` is reachable at runtime (for example when an administrator saves config options), while indexing workers hold the embedder returned by `vision.Model.FaceModel()` across the whole `GenerateEmbeddings` loop. A nil check at the top of `Run` should be enough.

**3. `markers.embed_model` is written but never read, and one write path skips it.**

In `entity.File.AddFaces`, the path that upgrades an embedding-less marker persists with `Updates(Values{"embeddings_json": ..., "landmarks_json": ...})`. `embed_model` is not in that column list, so the database row keeps an empty value while `SetEmbeddings` sets the field in memory only. Nothing currently reads `markers.embed_model` — only `faces.embed_model` is consumed, by `Face.SameEmbeddingModel` and the audit query. Either give the column a consumer and fix the update path, or leave it out until it has one.

**4. The ONNX embedding path is not exercised in a default checkout.**

Because `make dep` does not install SFace, `TestONNXEmbedder_Run`, `TestONNXEmbedder_AlignmentReducesPoseDistance`, `TestNewONNXEmbedder/SFace`, `TestNewONNXEmbedder/DimensionMismatch`, and `TestONNXEmbedder_Close/Idempotent` all skip silently, so the entire new inference path goes untested unless someone knows to run `make dep-sface` first. The tests themselves are good — they pass once the model is present. Making the dependency explicit in the development environment, or otherwise making the skip visible, would keep this path covered.

**5. `scripts/download-sface.sh` primary URL is not published yet.**

`https://dl.photoprism.app/onnx/models/face_recognition_sface_2021dec.onnx` currently returns 404, so `make dep-sface` always takes the GitHub fallback. The checksum verification means this is safe, and no change is needed on your side — we will publish the file. Noting it so the behavior is not mistaken for a script bug.

**6. `scripts/download-arcface.sh` does not verify a pinned checksum.**

Unlike the SFace script, it records the hash after downloading rather than checking it against a known value, so a changed upstream release asset would install without complaint. Pinning both hashes would make the two scripts consistent.

## Minor

- `EmbeddingModel.Bundled`, `License`, and `Redistributable()` are never read outside `models.go` and its tests. `Bundled: true` on SFace is also not accurate today, since it is neither installed by `make dep` nor present in the images.
- After a model switch, `Face.SetEmbeddings` rejects legacy clusters with a dimension error during optimize and merge until `faces reset` runs. This is documented, but it will be noisy in the logs, so a clearer message there may help operators.

Thanks again — the benchmark harness and the measurements behind it are exactly the kind of evidence that makes a change like this reviewable, and the cross-age and children numbers are compelling. Happy to take another look once the items above are addressed.

lastzero:
A sequencing note rather than a review: GPU/CUDA support for ONNX Runtime (#5703) is on hold until this PR is merged, so the two changes don't collide.

The CUDA execution-provider selection would land in `engine_onnx.go` and in the new `embedder_onnx.go` this PR adds, and once the embedder moves to ONNX there are two sessions to place on the GPU instead of one. That is much better built on top of your `Embedder` abstraction than merged into it, so #5703's remaining GPU items wait for this.

Two practical notes while you are in this package, since `develop` moved under you in `359df6641`:

- On your next rebase, `internal/ai/face/model_test.go` still merges cleanly (verified with `git merge-tree`), but `internal/ai/face/README.md` conflicts on exactly one line — the `**Last Updated:**` date. Keep yours and it is resolved; the sentence that commit added to the ONNX binding paragraph merges without help.
- That commit also changes `TestNet` behavior. It no longer skips when the detector fails to initialize while `assets/models/scrfd/scrfd.onnx` is present — that is a hard failure now, because a skip there silently hid an ONNX Runtime version mismatch behind a green `ok`. A genuinely missing model still skips, so nothing changes for a normal checkout.

lastzero:
@omerdduran Two hardening suggestions on the matching path — follow-ups rather than blockers, since the behavior predates this PR. Moving the thresholds into the model registry is what makes them worth doing.

**1. `SampleRadius` is clamped on write but read back unclamped.** `SetEmbeddings` and `UpdateMatchStats` both cap it at `face.ClusterRadius`; `Face.Match` and `faceCandidate.match` use the stored column as-is. Now that thresholds are per model and derived from `TestCalibrateFaceThresholds`, that column is only meaningful in the calibration that produced it — when the numbers are re-derived, existing rows keep radii from the previous calibration until a match run happens to rewrite them through `UpdateMatchStats`. Applying the same clamp where the value is consumed makes the invariant hold for stored data, and has the side benefit that a changed `--face-cluster-radius` takes effect on the next run rather than only after the rows are rewritten. `faces migrate` is already fine here — it goes through `entity.NewFace`.

**2. Consider a non-configurable ceiling on the accept distance itself.** `faceThreshold` bounds `face-cluster-radius` and `face-match-dist` at 1.5 each, independently, so the effective cutoff `SampleRadius + MatchDist` can be configured up to 3.0. Every model's vectors go through `NewEmbedding` and are L2-normalized regardless of dims, so distances live on [0, 2] for all of them, and two independent unit vectors sit at √2 ≈ 1.41 in expectation. A cutoff at or above that isn't an aggressive setting, it's "accept everything" — and the merges it produces outlive the setting that caused them. Your calibrated maxima top out at 1.22, so a ceiling around 1.4 would leave room for future models while putting the degenerate range out of reach. Worth noting this bound is model-independent even though the thresholds are not.

Both fit in one helper on `entity.Face` returning the clamped cutoff, with `faceCandidate` storing that instead of the raw radius — `Face.Match` and `faceCandidate.match` are a copy-pasted pair today, and the audit and query log lines recompute the same sum a third time.

Minor, while you're in `faceThreshold`: an out-of-range value falls back silently, so `--face-match-dist=2` quietly becomes the calibrated default with no indication why.

lastzero:
Thanks - I reviewed and ran the branch again at `8b10a55d5`. Everything from the earlier rounds is
addressed: the not-installed fallback, the nil session, marker provenance, the ArcFace checksums, the
read-side accept clamp and the out-of-range threshold warning all landed. `face.AcceptDist` as the single
definition of the cutoff, clamped on read as well as on write, is a clean result - the four consumption
sites all route through it and `grep` finds no raw `SampleRadius + MatchDist` outside tests.

State on my machine: `go build ./...`, `gofmt`, `shellcheck` and `make check-make-help` are clean,
`golangci-lint` adds no new findings in the files this PR touches, and the tests pass for
`./internal/ai/...`, `./internal/config`, `./internal/entity`, `./internal/entity/query`,
`./internal/photoprism`, `./internal/commands`, `./pkg/capture` and `./pkg/fs`. SFace was installed here,
so the ONNX embedder tests ran rather than skipping.

Most of what follows clusters around one thing: the window in which a library holds vectors of two widths.

## Should be resolved before merge

**1. Mixed embedding widths reach an unguarded loop, and two paths get there.**

`EmbeddingsMidpoint` guards its mean loop with `if len(emb) != dim { continue }`, but the radius loop
below it iterates `emb[i]` for `i := range dim` with no guard.

- `PUT /api/v1/markers/:uid` -> `Marker.Face()` -> `UpdateMarker` -> `OptimizeFor(subjUID)` ->
  `query.ManuallyAddedFaces` (no `embed_model` filter) -> `MergeFaces` -> `NewFace`. Naming a person who
  already has a manual cluster from the previous model gives `index out of range [128] with length 128`,
  and `OptimizeFor` runs synchronously after `marker.SaveForm(frm)` has already committed.
- `Cluster()` passes `face.EmbeddingModelName()` into `query.Embeddings`, and `whereEmbeddingModel`
  applies no filter when that is empty, so both widths reach `alg.EuclideanDist`. Worth knowing: there are
  two functions with that name. `pkg/vector.EuclideanDist` has an `a.Dim() != b.Dim()` guard returning
  NaN; the `pkg/vector/alg` one that `faces_cluster.go` actually hands to `DBSCANWithProgress` does not.
  The index-out-of-range then lands in `nearestWorker`, which is a goroutine, so `Faces.start`'s
  `recover()` does not see it.

Guarding the radius loop and filtering `ManuallyAddedFaces` by model covers both.
`TestBackgroundSamplesMidpoint/EmbeddingWithDifferentLength` looks like it pins this, but it puts the
shorter embedding first, so `dim` is 1 and the loop never over-indexes - and it asserts `radius == 2.01`,
which is the result of truncating a 2-dim vector to its first component.

**2. `Face.Match` validates the receiver's provenance, not the incoming vector's.**

`m.SameEmbeddingModel()` covers the cluster side only. FaceNet, AuraFace, arcface_r50 and arcface_mbf are
all 512-dim, so the length check is no backstop. `query.Faces(true,false,false,false)` has no model
filter, so `Audit` and `ResolveFaceCollisions` compute real distances between two unrelated spaces, and
with `--fix` a legitimate cluster picks up `Collisions++` and a narrowed `CollisionRadius`.

The bigger consequence is one level up: `MergeFaces` has no provenance filter either, and
`SetEmbeddings` sets `m.EmbedModel = face.EmbeddingModelName()` unconditionally. A cross-model merge
therefore writes a centroid that means nothing in either space, stamps it with the *current* model, and
clamps `SampleRadius` to `ClusterRadius` - the widest permitted accept distance. After that
`auditEmbeddingModels`, `FacesFromOtherModels` and `FaceMigrationCounts` all report clean, so the operator
is told no migration is needed. I could not produce an actual false accept from it (cross-space distances
sit near sqrt(2) and the ceiling is at most ~1.11), so the reliable harm is the false provenance and the
audit that no longer answers "does this library need migrating".

Passing the producing model into `NewFace`/`SetEmbeddings` instead of reading a global, and requiring
`SameEmbeddingModel()` in `Marker.Face()` and in `MergeFaces`' candidate set, closes both. The same
one-sidedness is in `f1.Match(...)` in `faces_audit.go` and `merge[0].Match(...)` in `faces_optimize.go`,
which are safe today only because `Dist` returns `-1` on a width mismatch.

**3. `Face.Match`'s min-distance loop is order-dependent.**

`if d := e.Dist(faceEmbedding); d < dist || dist < 0 { dist = d }` has no length guard, so a `-1` compares
as smaller than any real distance and overwrites a valid minimum. Same set, reordered:
`match=false dist=-1` versus `match=true dist=0.05`. It is the only one of four sibling implementations
without the guard - `minMarkerDistance`, `minEmbeddingDistance` and `Marker.SetFace` all have it.
Extracting the guarded one and calling it from all four would be the durable fix.

**4. The two "unknown current model" policies contradict each other.**

`ConfigureEmbedder` collapses an init failure into `ModelNone`, so `EmbeddingModelName()` becomes empty.
In that state `whereEmbeddingModel` applies no restriction (deliberately - the comment explains why),
while `ModelsComparable(stored, "")` returns `stored == ""`, excluding exactly the rows the configured
model wrote. Only one can be the intended policy. The observable end state is that `buildFaceIndex` drops
every row so matching is off library-wide, indexing keeps producing embedding-less markers at debug level,
`CountUnmatchedFaceMarkers` still counts them so the worker spins each cycle, and
`Config.FaceModel()` still returns the configured name because the explicit-and-installed early return
never re-checks whether the embedder actually loaded. Total signal is one warn line at boot. This is also
the state that makes path 2 of item 1 reachable.

It would also help to keep an embedder *failure* distinguishable from `FACE_MODEL=none` - a third state
rather than `ModelNone` - so it can log on every worker cycle instead of once.

**5. `faces config` reports a different model than the one in force, and `faces migrate` cannot fail.**

`InitCoreConfig` does not connect the database, so `libraryFaceModel()` returns empty and the preference
list wins. Same binary, same library:

```
--- faces config ---
face-model  sface
--- faces migrate --to sface --dry-run ---
faces: migration target sface does not match configured model facenet
EXIT=0
```

The `if c.db != nil` caching guard is right; it is the returned and printed value that is not. And this is
the command an operator runs before deciding whether to migrate.

Separately, `FacesMigrateIncompleteError` is not a `cli.ExitCoder`, and `main` prints plain errors without
forcing an exit code, so `photoprism faces migrate --yes && echo ok` prints `ok` whether the migration
ran, was refused, or cleared markers' vectors. `internal/ai/face/README.md` states the opposite. Wrapping
in `cli.Exit(err.Error(), 1)` would settle it.

**6. Three migration behaviors worth a second look.**

- `FinalizeFaceMigration` clears `embeddings_json` for every marker not on the target model, including the
  ones in `failedMarkerUIDs`, and there is no abort threshold. Recovery is circular: `File.AddFace`
  refills an embedding-less marker only if the detector finds the face, which is the condition that just
  failed. Hand-drawn markers, profiles and small faces fail the 41 % overlap match in
  `detectMigrationEmbeddings`; for `--to facenet`, `cropMigrationEmbeddings` falls back to the original
  file, which `fs.DecodeImageFile` cannot read for RAW/HEIC/video, so a purged thumb cache fails a lot of
  markers at once. The dry-run report does not say how many markers stand to lose their vectors.
- `internal/mutex` is in-process, so `Migrate`'s `IndexWorker`/`MetaWorker`/`VisionWorker` checks and
  `FacesWorker.Start()` are invisible to a running server -
  `docker compose exec photoprism photoprism faces migrate --to sface --yes` passes all of them. Finalize
  then deletes every `faces` row while the server may be writing `markers.face_id`. The identity snapshot
  and `sameFaceMigrationIdentities` guard handle a concurrent rename well, but the rollback leaves
  new-model markers alongside old-model clusters, i.e. matching fully dark until someone re-runs.
- `PlanMigration` requires `target == configured` and `dominantFaceModel` picks by count, so an
  interrupted migration cannot be resumed once `FACE_MODEL` is unset - `auto` resolves back to the old
  model. The per-file checkpointing itself is sound; it is the target selection that is not resumable.
  Persisting the in-progress target, or saying so in the error text, would cover it.

## Performance

- **Every process start now scans `markers`.** `Config.Propagate` -> `FaceModel()` ->
  `libraryFaceModel()` -> `MarkerEmbeddingModels()` runs
  `WHERE marker_type = 'face' AND LENGTH(embeddings_json) > 0 GROUP BY embed_model`.
  `markers.embed_model` has no `index` tag (`faces.embed_model` does), and the `LENGTH()` predicate forces
  a row scan anyway. It is cached per process, but that means once per CLI invocation too, not only at
  server start. Also `facesMigrateAction` calls `PlanMigration` and then `Migrate` calls it again, so the
  count queries run twice.
- **Aligned crops are capped at the `Fit720` detection thumbnail.** `GenerateEmbeddings` decodes the same
  file the detector used, so a 60 px face is upscaled ~1.9x into 112x112 from already-downsampled pixels.
  The box path deliberately went the other way -
  `crop.ImageFromThumb` -> `openIdealThumbFile` -> `findIdealThumbFileName` picks a *larger* cached
  thumbnail for small faces. Landmarks are absolute coordinates, so scaling them into a larger thumb by
  the width ratio looks tractable. Note the decode count did improve: one thumb decode per file instead of
  one `crop.ImageFromThumb` per face.
  Related: `benchmarkCrop` aligns from the full-resolution dataset image, so the shipped per-model
  thresholds were not measured on the crops production actually generates. Worth recording as a known gap
  either way.
- **The keypoint clamp skews the warp.** In `parseDetections`, `det.kps` is clamped to
  `[0, origWidth]`/`[0, origHeight]`. The decode itself matches InsightFace `distance2kps`, but the
  reference clamps boxes rather than keypoints, and `similarityTransform` is a least-squares fit over all
  five points - so one snapped landmark rotates and scales the whole crop. A face at the frame edge with a
  predicted eye 8 px outside a 60 px face moves that point 13 % of the face width. `sampleBilinear`
  already contributes transparent black out of bounds, which is what `warpAffine` does, so dropping both
  `clampFloat32` calls should be enough.
- Smaller: `onnxEmbedder.Run` serializes all inference behind one mutex. That is correct - it is the same
  lock `Close` takes, which is what makes a reconfiguration mid-inference degrade safely - and the
  TensorFlow path had the same shape, so no regression. Worth noting only because both ONNX sessions
  default to `NumCPU/2` intra- and inter-op threads, so they can oversubscribe a small host. And the
  ~92 MB TensorFlow SavedModel cached in `vision.Model.faceModel` is never released after switching to
  ONNX. `FinalizeFaceMigration` holds `DELETE FROM faces` plus three whole-table `markers` updates in one
  transaction, which is an exclusive write lock for the duration on SQLite.

## Model installation

`scripts/dist/download-models.sh` is now the checksum-pinned registry for model installs and replaces the
individual `scripts/download-*.sh` scripts - its header carves out only "models with their own
installation requirements". It already carries `sface` and `auraface` at exactly the checksums in
`models.go` (`0ba9fbfa...`, `a7933ea5...`), with mirror plus publisher fallback, an up-to-date check
against the installed file, dated backups, and a `--url` override that still enforces the checksum.
`dep-tensorflow` already calls it.

So `scripts/download-sface.sh` and `scripts/download-auraface.sh` can go, with `dep-onnx` calling
`scripts/dist/download-models.sh sface` directly rather than via a new `dep-sface` target.
`scripts/download-arcface.sh` is exactly the carve-out case - accept-license gate, zip extraction from a
release pack, no mirror - so it stays as it is.

Two follow-ons: `TestEmbeddingModelChecksums` should read the registry line
(`name|url|fallback|sha256|type|dir|file`) instead of the per-model scripts, and lose its `t.Skipf` -
right now it skips silently when a script is missing, so deleting those scripts would leave it passing
without checking anything. And with the shell registry owning every URL, `ModelInfo.SourceUrl` is a second
copy of it; the Go registry could keep the preprocessing contract and drop the URL.

## For the next detector

We are planning to add YuNet next, and two of the items above are worth doing in this PR rather than
leaving to that change, because both feed the alignment path:

- The keypoint clamp, since it perturbs the crops these thresholds are calibrated from and the vectors
  this PR persists. Measuring detector-swap drift later is only interpretable against an unskewed
  baseline.
- Which image the aligned crop is warped from. If that is not settled here, a detector swap changes the
  detector and the alignment source at the same time and the two cannot be separated.

Also, once a different detector is the default, `DetectorModel` has to move with it or its warn-only
checksum will log on every install. The warn-rather-than-refuse choice itself reads right to me -
recall on the next run is a much cheaper failure than a library of incomparable vectors.

## Also worth deciding

- `AcceptDistMax = 1.4` works as implemented and needs no change from you - this is just so the README
  sentence does not mislead a later reader. The text says a cutoff at or above ~1.41 "accepts every pair"
  and that the bound "follows from normalization alone". The averaging argument is right, but 1.41 is an
  *average* distance between two unrelated faces, and in practice a noticeable share of unrelated pairs
  land closer than 1.4. So the ceiling is a backstop that stops a misconfiguration from being
  catastrophic, rather than 1.4 being a safe setting. Worth softening the wording. Related: `ThresholdMax`
  is 1.5, i.e. above the ceiling, so a value between 1.4 and 1.5 passes flag validation and is then
  quietly reduced to 1.4 while the config report still shows what was set. We will sort that part out on
  our side when merging.
- The background filter was enabled by default before this change (`IgnoreBackground = true`, with
  `--face-allow-background` defaulting to false), so removing it is a behavior change for every existing
  library and wants a release note. The removal itself is necessary rather than optional -
  `Clusters.Contains` compared via `Dist`, which returns `-1` on a width mismatch, so under any 128-dim
  model every embedding would have been classified as background.
- `--face-skip-children` and `--face-allow-background` are gone with no `Hidden: true` shim, so a compose
  `command:` line carrying either now fails to start. The env vars are ignored silently and both options
  were `yaml:"-"`, so `options.yml` is unaffected. The removal is otherwise clean - no dangling references
  anywhere.
- `(*Model).FaceModel()` returns `face.ActiveEmbedder()` before it looks at `m.Name`, so a custom
  `vision.yml` face model is silently ignored once SFace is active - and when it *is* used,
  `EmbeddingModelName()` returns the registry name rather than the `vision.yml` one, so its vectors are
  stamped `facenet`. `(*Model).ModelName()` already has the right answer.
- `auto` asks the shared database via `libraryFaceModel()` but asks the local filesystem via
  `installedFaceModel()`. Two instances on one database with different installed models resolve
  differently, each hides the other's vectors, and `faces migrate` on one rejects the other's model.
- Both `embed_model` columns come from AutoMigrate, which `migrations run` only triggers via `--failed` or
  a version bump - dropping the columns and running `migrations run` exits 0 and creates neither.
- `similarityTransform`'s comment says degenerate landmark sets return an error, but `[[a,-b],[b,a]]` has
  `det = a^2 + b^2`, so it cannot reflect by construction and `variance <= 0` catches only all-five-
  identical. Mirrored landmarks fit at scale 0.225 with a 26.7 px residual on a 112 px template and no
  error, and `faceCropImage`'s fallback only triggers on `AlignPoints() == false`. Rejecting on residual
  or scale would let it fall back to the box crop. The math itself is right - it matches the closed-form
  Umeyama fit for every non-reflecting input.
- `internal/ai/face/test.go` (~270 KB of `BackgroundSamples`) is unreferenced now, and being in a
  non-`_test.go` file it still links into the binary.
- `Embeddings.Contains` treats the `-1` non-comparable distance as inside the radius - the same shape as
  the `Clusters.Contains` fix. No production caller today, so it is only a trap for whoever wires it up
  next, but the sentinel now means opposite things in two methods on the same type.

## Comments, tests and docs

One test gap is worth calling out on its own: `Faces.Migrate` never executes past `if opt.DryRun`, so the
batch loop, per-file checkpointing, `buildFaceMigrationClusters`, `FinalizeFaceMigration` and the
`w.start` + `Audit` tail have no end-to-end coverage - which is why items 5 and 6 are invisible to a green
suite. An end-to-end migration test would be the single most valuable addition here. A few others assert
less than they appear to: `TestFinalizeFaceMigration` creates no `faces` rows, so the delete it is named
for is a no-op; the `faces_audit_test.go` subtests only assert on functions other than the ones they call;
and the provenance write in `Marker.SetEmbeddings` and the collision gate in `faceCandidate.match` can
both be deleted with everything still green.

On docs, the quick ones: the `FACE_MODEL` row in `internal/ai/face/README.md` omits `auraface`, which
contradicts the model table above it and the CLI help; `face.QualityThreshold` and
`PHOTOPRISM_FACENET_DISABLED` are referenced but do not exist; `internal/ai/vision/README.md` still
presents face embeddings as `vision.yml`-owned; and there is a little British spelling
("cancelled", "initialise"). Also, we keep package READMEs as descriptions of current behavior rather than
of the change, so the before/after benchmark table and the "now runs at ~155 ns/op, down from ~242" style
lines are worth rewriting as plain current values.

We have the rest of the comment, test and documentation notes written up and will share or fold them in
separately - no need to chase them from this comment.

Thanks again - the registry as one source for the preprocessing contract, decoding the landmarks the
bundled model always predicted, and clamping the accept distance where it is consumed are all real
improvements over what they replace, and the measurements behind the model choice made this reviewable.
Happy to take another look once the items above are addressed.

lastzero:
Re-checked at `b703db5` - build is clean and `./internal/ai/face/...`, `./internal/entity` and
`./internal/commands` pass. That is nearly all of it.

Cleared since the last round: the mixed-width crash on both paths, the provenance items, the
model-missing behavior, the migration abort, the marker index, the migrate exit code, and the
`faces config` model. Four worth calling out:

- The `EmbeddingsMidpoint` mean bug you found - divided by the full count including the vectors the loop
  skips - was mine to have caught.
- You extracted the min-distance loop instead of patching `Face.Match`, so all four call sites now share
  one guarded `Embeddings.Dist`. I had been about to say that item was optional now; the extraction is
  better than what I originally suggested and it closes the latent `Embeddings.Dist` case as well.
- Rejecting a poor landmark fit on the RMS residual, alongside dropping the clamp, is exactly right -
  `faceCropImage` now actually reaches its fallback instead of warping from a bad fit.
- `RegisterDb()` plus the new `face-model-status` row answers more than I asked for, since it also
  surfaces a model that failed to load rather than just naming one.

Still open, roughly in order of value:

1. **The aligned crop is still warped from the `Fit720` detection thumbnail.** This is the other half of
   the alignment pair and the one that most affects quality: a 60 px face is upscaled ~1.9x into 112x112
   from already-downsampled pixels, where the box path picks a larger cached thumb via
   `findIdealThumbFileName`. Landmarks are absolute coordinates, so scaling them into a bigger thumb is
   most of the change.
2. **The startup scan.** `MarkerEmbeddingModels` still runs `LENGTH(embeddings_json) > 0 ... GROUP BY
   embed_model` across the whole markers table. The index helps the grouping but the `LENGTH()` predicate
   still reads every row, and this runs once per process - so every CLI invocation pays it, not just
   server start. A cheaper probe, or resolving lazily instead of during `Propagate`, would both work.
3. **`PlanMigration` runs twice** - once in `facesMigrateAction` and again inside `Migrate`. Passing the
   plan through would halve the count queries.
4. **Duplicate normalization in clustering.** `query.Embeddings()` normalizes every vector via
   `UnmarshalEmbeddings`, then `EmbeddingsMidpoint` normalizes the same slices again in its mean loop -
   DBSCAN only reads them in between, so the second pass is pure waste, and it is also what mutates the
   caller's data. It cannot simply be dropped, because `MergeFaces`, `Marker.Face()` and
   `buildFaceMigrationClusters` pass vectors straight from `json.Unmarshal` that were never normalized.
   An early return in `normalizeEmbedding` when the sum of squares is already 1 costs nothing on a
   genuine call and halves every redundant one.
5. **Thread counts.** Both sessions set intra-op *and* inter-op to `NumCPU/2`. Inter-op only helps graphs
   with independent branches, and these are sequential CNNs, so setting it to 1 halves the demand for
   free. The auto value should also account for index workers, since `onnxEngine.Detect` takes no lock
   and up to `IndexWorkers()` detections run at once with their own pools. `Config.indexWorkers()` is the
   pattern to follow. Please keep this on `PHOTOPRISM_FACE_ENGINE_THREADS` rather than adding a second
   option, since detection and embedding always run together. If unclear, I can take care of this when
   merging.
6. **Moving the model installs to `scripts/dist/download-models.sh`**, which already carries `sface` and
   `auraface` at the same pinned checksums.

Nice touch renaming `test.go` to `samples_test.go` - that keeps the sample fixtures out of the binary.

lastzero:
Thanks - I went through the branch again at `410277c47` and ran it on both drivers. The seven commits
since `b703db524` close most of what was left: the aligned crop no longer warps from the detection
thumbnail, `auto` resolves from the indexed provenance column instead of scanning the blob, the plan is
passed into `Migrate` rather than derived twice, the redundant normalization pass is gone, and SFace and
AuraFace now install from the shared registry with `TestEmbeddingModelChecksums` reading the registry row.
Preserving automatic assignments and re-seeding each cluster from all of them is a real improvement over
what it replaced, and the hidden-cluster restore is a thoughtful piece of work - it is not obvious that
markers are the only thing that survives to carry that decision.

Two things are red, though, and I could not merge on top of them.

## Must be resolved before merge

**1. `TestMergeFaces/SameSubjects` fails, so `make test-go` and `make test-mariadb` are both red.**

```
expected: "WRFY77XXC4ITVUSIBCDEM3PRDKBILC4M"
actual  : "5K5PKAKVOM3E7A5E65V657TPFTTUMB37"
```

On linux/amd64 with Go 1.26.6, on SQLite and on MariaDB. Baselined with the same invocation: `develop`
passes with `4FD6...`, `b703db524` passes, and it starts failing at `8602d7001` - the commit that added
the already-normalized early return. So the value in the test was never produced here, on any driver.

The assertion pins a base32 SHA-1 of a float64 centroid, which means it pins the last bits of an
arithmetic result rather than a behavior, and re-pinning it to a third constant would only move the
problem. `SampleRadius` in the same subtest still passes and is what actually covers the merge, so I would
drop the literal and either assert the ID is non-empty or compare it against an `entity.NewFace(...)`
computed in the test.

Worth knowing why it moved at all: a cluster ID is a content hash of its centroid, so anything that
changes how the centroid rounds renames every cluster in every library. That is harmless here, since the
migration recreates them anyway, but it is the reason the constant is brittle.

**2. `TestFaces_migrate` leaves the process-wide embedder changed, and `TestFaces_migrationEmbedder` then
fails on MariaDB.**

```
make reset-acceptance
PHOTOPRISM_TEST_DRIVER=mysql PHOTOPRISM_TEST_DSN=... \
  go test -parallel 1 -count 1 -cpu 1 -tags="slow,develop" ./internal/photoprism

--- FAIL: TestFaces_migrationEmbedder
    faces: loaded embedding model facenet, expected sface
```

It passes when run alone and passes against a database that was not reset, so it depends on state another
test leaves behind. The `-v` log shows the FaceNet threshold defaults (0.42 / 0.64 / 0.4) being
re-propagated inside each `TestFaces_migrate` subtest and never going back: `newMigrateTestConfig` builds
a config with an empty models path, which reconfigures the process-wide embedder, and
`restoreEmbedderSettings` captures `face.ConfiguredModel()` - the global that was already changed - rather
than a value known to be good. The restore error is discarded too (`_ = face.ConfigureEmbedder(restore)`),
so a restore that does not take is silent.

This is worth more than a green build to you: `TestFaces_migrate` is the end-to-end coverage the earlier
round asked for, so if it runs against an embedder it did not intend, its own assertions prove less than
they look like they do.

I checked whether the same shape reaches production, and it does not - `ConfigureEmbedder` sets
`configuredModel = ModelNone` when an ONNX init fails, so `Model.FaceModel()` returns nil instead of
quietly falling through to TensorFlow. That part holds.

**3. `FaceEngineThreads()` also decides *when* face detection runs, and the new formula flips it.**

`FaceEngineShouldRun` uses the thread count as a machine-capability proxy: `RunOnIndex` needs
`FaceEngineThreads() > 2` and `RunNewlyIndexed` needs `<= 2`. With MariaDB and no override,
`indexWorkers()` returns `cpus/2`, so `NumCPU / IndexWorkers()` comes out as **2 on every host regardless
of core count** - which moves detection off the index path and onto the newly-indexed pass for
essentially every production deployment. Measured here on SQLite with 28 CPUs (`sqlite-cap`, 4 workers):
7 threads, down from 14. And on a low-memory host `IndexWorkers()` is 1, so the count doubles to `NumCPU`
exactly where memory is scarce.

Separately, the same value feeds both sessions (`internal/config/config.go`, the detector and the embedder
settings a few lines apart), and the embedder is serialized behind `onnxEmbedder.Run`'s mutex. The
reasoning in the comment - detection takes no lock, so one pool runs per indexing worker - is right for
the detector and inverted for the embedder, which never runs concurrently with itself and now gets fewer
threads than before. With AuraFace at roughly 8x SFace per face that is the wrong direction.

Giving the embedder its own count, and deriving the run-mode gate from something other than a per-session
thread count, would settle both.

**4. Some documentation the last commits made false.**

We keep docs describing current behavior in the same commit that changes it, and these all sit in files
the branch already edits:

- `internal/config/flags.go`, `face-engine-threads`: "0 uses half the available CPU cores". That is
  `--help` text and it feeds the generated config reference.
- `internal/ai/face/README.md`, the detector bullet and the `FACE_ENGINE_THREADS` row: same claim,
  `runtime.NumCPU()/2`.
- `internal/ai/face/README.md`, the `auto` paragraph: "A library that already holds face vectors keeps the
  model that produced them, because resolving to a different one would leave every stored cluster
  incomparable with everything indexed afterwards." `RecordedMarkerEmbeddingModels` excludes
  `embed_model = ''`, so a legacy library with a single recorded SFace marker now resolves to `sface`.
  That sentence is exactly what the change breaks.
- `internal/ai/face/README.md`, the migration paragraph: "seeds each replacement cluster from all of them"
  stopped being true with `centroidSamples`, and neither the outlier exclusion nor the hidden-cluster
  restore is described.
- Nothing yet says the aligned crop is warped from a larger rendition than the one the detector measured
  in. That is the most consequential behavior change in this batch, and it is the thing a later reader
  will need to know before touching the alignment path again.
- `Last Updated` still says August 17.

## Smaller things

- `crop.ImageFromIdealThumb` and `photoprism.ConfigFileName` are new exported functions with no test of
  their own; `RecordedMarkerEmbeddingModels` only has a subtest hanging off `TestMarkerEmbeddingModels`.
- `centroidSamples` compares `marker.Embeddings().Dist(midpoint) <= registered.ClusterDist`, and `Dist`
  returns `-1` when nothing is comparable, so a non-comparable marker counts as a sample. The query keeps
  it unreachable today, but it is the same shape you fixed in `Clusters.Contains` and `Embeddings.Contains`
  and the sentinel now reads three different ways in three places.
- `internal/ai/face/models_test.go`: the old four-line `TestEmbeddingModelChecksums` description was left
  above `readModelScript`.
- A few non-ASCII characters crept into comments - em dashes in `embedding.go` and `models.go`, and
  `sqrt(2)` written as a symbol in `config.go` and `config_test.go`. We keep comments plain ASCII.
- `dominantFaceModel`'s comment still says "most of the counted vectors", which is now most of the
  *recorded* ones.

## Things we will take at merge - no need to chase them

The removed-flag shim, the background-filter release note, the `vision.yml` face model precedence, `auto`
as a per-node answer to a per-database question, a numbered migration for the two `embed_model` columns,
the assertions missing from `faces_audit_test.go`, and the package READMEs and code maps outside
`internal/ai/face`. Same for the two migration items from last round that are design calls rather than
defects: there is still no cross-process interlock, and a finalize that rolls back leaves new-model
markers beside old-model clusters.

A few observations from this round that are ours to decide rather than yours to fix, recorded so they are
not lost:

- `auto` ignoring blank-model markers means an operator who tries `sface` on a few photos and then unsets
  `FACE_MODEL` flips the whole library to `sface` and hides every FaceNet vector. A log line at resolution
  time would make that visible.
- Embeddings are no longer a pure function of the original and the model: the aligned crop reads whichever
  cached rendition satisfies the template width, so it depends on the thumb ladder and on `--thumb-size`.
  Deterministic for a fixed configuration, since thumbnails are generated before detection, but raising
  `--thumb-size` later changes vectors on re-index.
- The finalize now issues one `UPDATE` per assigned marker inside the single whole-table transaction,
  where it used to issue one per manual marker. On your own measurement that is 66 against 10; on a large
  curated library it is easily a hundred times more.
- Widening the identity snapshot from manual markers to every assigned one makes a concurrent server run
  abort the finalize much more readily. The rollback is sound, so this is a trade rather than a bug.

Everything else checks out: build, `gofmt`, `shellcheck` and `make check-make-help` are clean,
`golangci-lint` reports the same count as `develop` with nothing new in the files you touch, and the
branch still merges cleanly. Happy to take another look as soon as the two failures are sorted.

lastzero:
**Well done, Ömer! This looks ready to merge now.** 🎉

Computing the merged ID from the same embeddings is the correct approach. Thank you for investigating why it differed instead of simply re-pinning it. The fused multiply-add on arm64 is the reason for the difference and is useful information for the next time a float-derived constant appears in a test.

Two things you found that I had not:

- `FaceEngineThreads` wrote its derived value back into the options, so whichever getter ran first would have frozen its answer for the other one. Splitting the two counts would have been half a fix without that.
- Reading the restore settings back from the global was the actual mechanism, not just the trigger, so one missed restore spread to every capture after it. Taking the snapshot once in `TestMain` and failing the test when a restore does not take is a better guarantee than what was there before.

The gate reading the undivided count is also the better call - it is the machine's capability, so it should not follow the database driver.

I'm going to merge this shortly, and then I'll finish the rest on `develop`. Nothing further is needed from you right now.

PR #5765, opened 2026-08-06
Metadata: Complete XMP and DCMI tags for the report #2260

photoprismci:
Completes the `xmp:` and `dc:` struct tags on `meta.Data` so the metadata
compatibility report reflects what the XMP reader actually supports.

The tags are read only by `internal/meta/report.go` via reflection, which
renders `photoprism show metadata` and the published compatibility table.
`report.go` is their sole consumer, so this changes documentation output
only — parsing is driven by the `meta:` tags and by the XPath chains in
`internal/meta/xmp_document.go`, neither of which is touched.

### Changes

- `xmp:` tags: 9 fields to 32, transcribed from the reader's priority
  chains and namespaced, so `exif:` / `exifEX:` / `aux:` are distinguishable.
- Corrections, not just additions: `TakenAt` claimed `DateCreated,CreationTime`
  but resolves `photoshop:DateCreated -> exif:DateTimeOriginal -> xmp:CreateDate`;
  `CameraModel` claimed `CameraModel,Model` but only reads `tiff:Model`.
- `GPSLatitude` / `GPSLongitude` now report `exif:GPSLatitude` / `exif:GPSLongitude`.
  The reader populates coordinates from XMP; the empty cells read as
  "XMP GPS is unsupported", which is wrong.
- `dc:` tags: 1 field to 6, covering the Dublin Core properties the reader
  consults.
- Adds `internal/meta/report_test.go`, matching `internal/thumb/report_test.go`.
- Corrects the doc comment, which named a `show metadata-fields` command
  that does not exist.

### Not included

`Lat`, `Lng`, `TakenNs`, `TimeZone` and `Orientation` are tagged `meta:"-"`,
so `Report()` emits no row for them. Surfacing them means changing the
report filter, which is a separate call about whether computed fields
belong in a user-facing table.

The test pins the tag values but cannot mechanically verify them against
the chains, which are unexported compiled expressions. It catches typos
and deletions, not semantic drift.

The published table at https://www.photoprism.app/kb/metadata/ is a pasted
copy; regenerated in photoprism/photoprism-web#4.

### Acceptance Criteria

- [x] Every field the XMP reader populates and that `Report()` emits a row
      for carries an `xmp:` tag. (MUST)
- [x] Tag values match the reader's priority chains in order. (MUST)
- [x] No parsing behavior changes. (MUST)
- [x] `go test ./internal/meta/...` passes. (MUST)

### DIFF
--- a/internal/meta/data.go
+++ b/internal/meta/data.go
@@ -18,20 +18,24 @@ const (
 //
 // Note: the meta:"…", xmp:"…", and dc:"…" struct tags below are read by
 // internal/meta/report.go via reflection to render the metadata-source
-// columns of `photoprism show metadata-fields`. Do not delete them as
+// columns of `photoprism show metadata`. Do not delete them as
 // "vestigial" — they are documentation that ships in the CLI report.
+//
+// The xmp:"…" tags list the namespaced properties the XMP reader consults,
+// in the priority order its chains apply (internal/meta/xmp_document.go);
+// the dc:"…" tags name the Dublin Core properties among them.
 type Data struct {
 	FileName         string        `meta:"FileName"`
 	MimeType         string        `meta:"MIMEType" report:"-"`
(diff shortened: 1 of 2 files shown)


lastzero:
These changes should now be published.

PR #5783, opened 2026-08-15
Retro fit postgres branch improvements

### LINKED ISSUE #692: moments: near \"INNER\": syntax error (update folder dates) & (update album dates)
I have created a docker container on a QNAP NAS running the latest version of their Container Station ("docker") platform.

When I am doing an import of photo's, I am seeing these errors flagged in the log.

```
time="2020-12-10T03:08:04+08:00" level=error msg="moments: near \"INNER\": syntax error (update folder dates)"                              
time="2020-12-10T03:08:04+08:00" level=error msg="moments: near \"INNER\": syntax error (update album dates)"                               
time="2020-12-10T03:08:04+08:00" level=info msg="Import completed in 34 s"                                                                  
```

Let me know what additional information you may need to troubleshoot this.

keif888:
### Description

These changes implement improvements that have been discovered whilst producing the Postgres branch.  The majority are improved tests, either addressing ToDo issues, or adding post checking to ensure that a successful call to a function performed as expected.  Some of the tests around nil results are there to confirm behaviour of Gorm v1, so that Gorm v2 can be confirmed to function the same way.  It is now possible to test the help functions for commands and capture all output from commands.

Non test changes:
- Detect and use RunNonInteractively at function execution time so that it can be tested
- Add ability to check if a db is open and pingable
- Ensure that the migrate and version tables exist when init-ing the test database
- Flush the Keyword and PhotoKeyword cache when all other cache's are being flushed
- Wrap the error reporter in a Mutex so it can be shut down cleanly
- Add missing SQLite query for updating Album and Folder dates.
- Ensure that RemovePeopleAndFaces doesn't error if there are no people or portraits

#### Related Issues

- Fixes #692 by adding valid SQLite queries

### Acceptance Criteria


lastzero:
Thanks for this - it's a big net improvement, and the test-quality work in particular is very welcome.

I validated the branch locally. Both suites are green, including the private `plus`, `pro`, and `portal` extensions:

- `make test-go` (SQLite) - 144 packages, exit 0
- `make test-mariadb` - 107 packages, exit 0
- `go build ./...`, `go vet`, `gofmt -l`, and `golangci-lint --new-from-rev develop` all clean

There are a few things I'd like your input on before merging. I'll take care of the smaller cleanups afterwards.

### 1. `entity/error.go` - the worker can latch off permanently

The deferred cleanup in `LogEvents` releases the `logEvents` gate before it calls `mutex.ErrorWorker.Stop()`:

```go
logEvents.CompareAndSwap(true, false)   // gate released
event.Unsubscribe(s)
mutex.ErrorWorker.Stop()                // busy/canceled cleared only here
```

A `LogWarningsAndErrors()` landing in that window wins the CAS, gets `"already running"` from `Start()`, discards the error, and never spawns a worker. Nothing resets `logEvents` afterwards, so the `errors` table stops receiving rows for the rest of the process with no diagnostic, which makes the `// This should not be possible…` comment inaccurate. `Config.MigrateDb` and `Config.InitTestDb` are both callers, and ~25 CLI subcommands `defer conf.Shutdown()` → `mutex.CancelAll()`.

Moving `Stop()` above the CAS and logging the discarded error would close it. More broadly: `logEvents` already guarantees single-start, so `Canceled()` looks like the only thing the Activity adds. Could you say a bit about what the Activity is solving? If it's only the shutdown path, gating on `HasDbProvider()` may be enough and would leave one gate instead of two that can disagree.

Related: `WorkersRunning()`'s new comment says `ErrorWorker` is "started on startup and stopped on shutdown". It's actually started lazily per `MigrateDb`/`InitTestDb`, and shutdown only *cancels* it. `Stop()` runs from the goroutine's own defer, which fires only if another log event arrives after the cancel.

### 2. SQLite album/folder dates are shifted by the stored UTC offset

`mattn/go-sqlite3` binds `time.Time` with the offset attached, and SQLite's date functions convert to UTC before formatting:

```
sqlite> select strftime('%Y-%m-%d', '2020-01-01 00:30:00+01:00');
2019-12-31
```

This is common in practice. In one of my dev indexes, 38 of 48 rows carry a non-UTC offset. A folder whose newest qualifying photo falls within the offset of local midnight would get `album_year=2019, album_month=12, album_day=31` for a January 2020 folder.

There's a second, SQLite-only effect: `MAX()` over TEXT-with-offset orders by local wall clock, while MariaDB's `MAX()` orders by instant, so a folder mixing `Local`-zone and named-zone photos selects a different row per driver.

Both go away with a plain substring extraction, which is also what MySQL's `YEAR()/MONTH()/DAY()` do to a naive `DATETIME`:

```sql
SET album_year  = substr(taken_max, 1, 4),
    album_month = substr(taken_max, 6, 2),
    album_day   = substr(taken_max, 9, 2)
```

That has the side benefit of removing the NULL path: `strftime` returns NULL on an unparseable value, and `Album.AlbumYear` / `Folder.FolderYear` are plain `int`.

The fixtures all use `time.UTC`, so the new `TestUpdateAlbumDates` / `TestUpdateFolderDates` assertions can't catch this. They're good assertions otherwise (both fail on `develop`, which is exactly what I want to see). A case with an offset-bearing `taken_at_local` would cover it.

### 3. Performance of the same two statements on large libraries

I benchmarked these against a 4M-photo / 40k-folder database built from the real schema. The plans are good - `MATERIALIZE p` over `SCAN photos USING INDEX idx_photos_path_name`, then an indexed seek into `albums` / `folders`, with no temp B-tree or sort. Timing is linear: 0.5 s per statement at 1M photos, 1.9 s at 4M, warm cache on NVMe.

Two things I'd like to get right before this reaches users with big libraries:

**It's a full scan of `photos` added to every `moments` run.** SQLite users pay nothing for these today, since the current code falls through to `default: return nil`. The scan has to touch every row for `taken_src`, `photo_quality`, `deleted_at`, and `taken_at_local`, so it becomes I/O bound once `photos` stops fitting in the page cache - my numbers are a best case, and the kind of NAS hardware in #692 would be substantially slower. That's amplified by two things: PhotoPrism doesn't enable `journal_mode=WAL`, so the write lock blocks readers as well, and the SQLite DSN sets `_busy_timeout=5000` (`pkg/dsn/driver.go`), so concurrent work fails with "database is locked" past 5 s. At 4M photos the two statements already total ~3.8 s warm.

**Every run rewrites every row.** Re-running with all values already correct still reports `changes() = 40001` and takes the same time - 80k no-op row updates and ~2 MB of journal churn per `moments` run at that size, plus index maintenance on `idx_albums_ymd` and `idx_albums_country_year_month`. Adding an inequality guard skips the writes once the dates have settled and shortens the lock window:

```sql
AND (albums.album_year  <> substr(taken_max, 1, 4)
  OR albums.album_month <> substr(taken_max, 6, 2)
  OR albums.album_day   <> substr(taken_max, 9, 2))
```

The existing MySQL statement has the same property, so I'm happy to bring that one to parity separately - I mainly want the SQLite and eventual Postgres forms to start out with it.

### 4. `TestImportSQL`

`ImportSQL` runs `q.Raw(stmt).Scan(&result)` and discards the error, and `connectDb` calls `LogMode(false)`, so gorm v1 only prints from `AddError` at `detailedLogMode`. A failing statement therefore produces no output at all, and `assert.NotContains(buffer, "level=error")` holds either way. Asserting `c.Db().HasTable("importtest")` between the create and the drop would pin it.

The test also sets `c.options.DatabaseDSN` from the raw `PHOTOPRISM_TEST_DSN`, which bypasses `entity.TestDbDSN`'s per-package `acceptance_<pkg>_<hash>` isolation and runs the DDL against the shared `acceptance` database. The pool opened by `connectDb()` is never closed.

### 5. Fixture changes - a couple of intent questions

- **`Photo23`**: GORM v1 overwrites `PhotoLabel.PhotoID` from the parent before insert, so the first argument to `LabelFixtures.PhotoLabel(...)` is ignored at persist time. `10000018 → 10000023` changes nothing in the database (and `Photo23.ID` is `1000023`, not `10000023`). The hunk's only real effect is dropping the `likeLabel` association, which is what drives the `TestPhoto_Merge` `1 → 2` change. Was that the intent? Seven other fixtures still pass `10000018`.
- **`TestFindClientByNodeUUID`**: `assert.Equal(t, m.UserUID, UserFixtures.Get("node").UserUID)` compares `""` to `""`, since there's no `"node"` key in `UserFixtures`. Was a matching user fixture meant to land with the new client fixture?
- **`"cows"` album**: not referenced by any test. Its only current effect is shifting the moment offsets in `search/albums_test.go` from 2 to 3. Is it staged for something later?

### 6. Two smaller ones

- `photo_title_test.go`: the exact-title assertion became `HasSuffix(m.PhotoTitle, "/ 2008")` plus a loop over `SubjectNames()`. When that slice is empty, which is the order-dependence being worked around, `"Unknown / 2008"` passes. The polluter looks like `TestMarker_SaveForm`; restoring the subject names in a `t.Cleanup` there would let the exact assertion stay.
- `NewTestContext`: `app.HideHelp` flipped to `false` and the `cmd.HideHelp = true` workaround was dropped, re-enabling help generation that the removed TODO described as panicking. It works in my runs, and I assume the new grandparent `cli.NewContext` supplies the parent that was missing. Could you confirm that's the mechanism? It's a harness-wide change worth its own line in the description.

Also, the doc comment on `resetConfigAndDB` mentions `PHOTOPRISM_TEST_DSN_NAME`, which doesn't exist elsewhere in the repo, and a Postgres path that can't run here since no Postgres GORM dialect is registered.

Thanks again - happy to merge once these are settled.

keif888:
I will work through the code changes and re-submit.

**1. entity/error.go**
 _Could you say a bit about what the Activity is solving?_
The mutex code was added because in cases where a large number of errors have been created, the async LogEvents function gets backlogged.  If a shutdown is issued whilst these messages are being written, then the HasDbProvider may return true, but the subsequent Db().Create would panic when the main go routine triggered the close of the database.  Without the mutex I couldn't find a reliable way to gate against a database connection being closed mid transaction.
This should be a testing issue, but it's possible a user could have many errors, and shutdown to address them, but this could then result in a panic.

**2. SQLite album/folder dates are shifted by the stored UTC offset**
I will see if I can change a fixture to force a failure, and the correct the code.

**3. Performance of the same two statements on large libraries**
I will look at how to optimise the moments process to see if there is some way to reduce the number of albums being processed, or reduce the impact on SQLite users.  If I can improve it I will apply to all DBMS' if appropriate.

**4. TestImportSQL**
I will revisit this test.  It's only working in the Postgres branch because the logging is set to Error with a comment.  ```// Log level  <-- This should be Silent to match GORM1, set to Error before release...```
I believe that the lack of switching databases is a hold over from my reversion to the entity.TestDbDSN from my previous way of doing it.

**5. Fixture changes - a couple of intent questions**
- Photo23 - typo when transcribing from Postgres to develop.  I changed the number, and removed the landscape label to that and Photo24 (which I missed when transcribing) so that test TestPhoto_IdenticalIdentical would work when run **on its own**.
- TestFindClientByNodeUUID A bug in a test that happens to not fail.  I will correct.
- cows album is used in some gorm v2 tests, that aren't included, and 2 tests that I should have included but missed, I will fix.

**6. Two smaller ones**
- I will add cleanup to TestMarker_SaveForm.
- HideHelp = false panic was caused by the missing parentCtx in the original cli.NewContext.


And the comments, that's what I get for copying the comment and not updating it for the "old" way of doing things.
Postgres branch has PHOTOPRISM_TEST_DSN_NAME so you can swap between pre-configured DSN's.

keif888:
**2. SQLite album/folder dates are shifted by the stored UTC offset**
I updated an existing fixture as follows:
```
	"Photo03": { // JPG + JPG (same instance ID) + MP4, Indexed, Place from metadata
		ID:               1000003,
		PhotoUID:         "ps6sg6be2lvl0yh0",
		TakenAt:          time.Date(1990, 4, 18, 1, 0, 0, 0, time.UTC),
		TakenAtLocal:     time.Date(1990, 4, 18, 1, 0, 0, 0, time.FixedZone("UTC+8", +8*60*60)),
```
And running the TestUpdateAlbumDates test against SQLite and MariaDB caused both to fail, as the test expects ```1990-04-18``` but got ```1990-04-17``` from both DBMS'.
MariaDB stored ```1990-04-17 17:00:00``` and SQLite stored ```1990-04-18 01:00:00+08:00```

The AlbumPath and PhotoPath are both sourced from the TakenAt field, so it is currently possible to have a TakenAtLocal value that is for the following month due to time offset differences between TakenAt (UTC) and TakenAtLocal.
At this point the SQLite query is producing the same result as the MariaDB query, so it is working as developed.
The issue is that this may not be working as expected/designed.

Options:
- Leave as is, so the Album's Year, Month and Day are based in the Local time of the photo, and may not match the paths
- Change the Album's Year, Month and Day so they are based on the times in TakenAt, which will match the paths

lastzero:
I re-checked at `556abdd` to see what's already resolved:

- **The `entity/error.go` latch is closed.** Moving `mutex.ErrorWorker.Stop()` above the `CompareAndSwap` does it - I traced every exit path from `LogEvents` and they all run the same defer, so the "gate released while still busy" sequence is gone.
- **My question about what the Activity is for is answered.** The shutdown race you described is a real one, and a `HasDbProvider()` check on its own would not have covered it, since the database can still close between that check and the `Create` that follows. `CancelAll()` running ahead of `CloseDb()` is what gives the worker a chance to see it coming, so the second gate earns its place.

Two things that I got wrong and are worth noting, since you spent time on both of them:

- **I was wrong to flag the UTC shift, so please stop chasing it.** You were right that you could not prove it. `go-sql-driver` writes with `Loc=UTC`, so MariaDB stores the UTC wall clock and `DAY()` returns the UTC day, while SQLite stores the offset and `strftime` converts it to UTC. The two agree. I also checked my own dev index: 38 of 48 rows carry a `+01:00` offset, and **zero** rows change date, so no folder is affected. Neither of the two options you offered is needed. I would still take `substr` when you next touch those statements, because it matches what the Go path in `folder.go` does and it removes the `strftime` -> NULL path into a plain `int` column, but that is a preference and not a blocker.
- **I read `_synchronous=NORMAL` as a durability change, and it is not one.** It is already go-sqlite3's default, so `develop` runs at NORMAL today and nothing moves. The only real change is `journal_mode`, and since rollback-journal + NORMAL is the combination that can corrupt on power loss while WAL + NORMAL cannot, WAL is the safer default rather than the riskier one. That part I am glad you did.

On the SQLite skip in #692: thank you for digging out the original error, that settles it. It was a syntax problem, not a performance or locking complaint, so there is no historical reason to keep SQLite excluded. WAL also answers the half of my item 3 that mattered most - writers no longer block readers, so a long moments statement no longer freezes the UI.

It would be great if we could resolve these issues before merging, listed in order of priority:

1. **`Config.DatabaseFile()` needs to go back to a path-only derivation.** Routing it through `dsn.Parse().SQLiteFilename()` means the regex parses a filesystem path as if it were a credential string, and several legal configurations now resolve wrongly:

   | storage path or DSN | before | now |
   | --- | --- | --- |
   | `/photoprism/my@storage/index.db` | correct | `storage/index.db` (relative) |
   | `/srv/photos (main)/index.db` | correct | `/srv/photos (main/index.db` |
   | `file:index.db` | `index.db` | `file:index.db` |
   | `.../index.sqlite`, `:memory:` | correct | `""` |

   This matters because `backup.RestoreDatabase` calls `tables.Drop(c.Db())` before it runs `sqlite3 <DatabaseFile()>`, and `sqlite3 ""` succeeds against a private temporary database. So on an affected install a restore empties the index, writes the dump nowhere, and exits 0. Either keep the old cut (`TrimPrefix "file:"` then split on `?`), or have `SQLiteFilename` take the configured driver from the caller instead of re-sniffing it, and trim `file:` on both branches. Worth adding the relative-`file:`, non-`.db`, `@` and `(`/`)` cases to `TestDSN_SQLiteFilename` - none of the six current cases cover a shape that breaks. Independently, `RestoreDatabase` should refuse to run when `DatabaseFile()` is empty, before it drops anything.

2. **`dsnPattern` should stay anchored, and `name` should keep accepting `?`.** Dropping the leading `^` means an input the pattern cannot match from position 0 no longer fails - the engine restarts at a later offset and returns a partial match with the prefix silently discarded. Narrowing `name` from `.*?` to `[^?]+` is what makes those inputs unmatchable in the first place, so the two changes compound. I isolated it on one string:

   ```
   PR pattern            matched "a=1?b=2"   user=""           name="a=1"
   PR pattern + ^        no match
   PR pattern + ^ + .*?  full match          user="photoprism" name="photoprism?a=1"
   ```

   Restoring both keeps `User`, `Password`, `Server` and `Name` derived from the whole string. It would also be good for `MaskPassword` to fail closed rather than return its input when the pattern does not match at all, so a parse failure cannot be mistaken for "there is no password here". A second case is already visible without any of that: `user:pass@tcp(host:3306)/` (a MySQL DSN with no database name, which the driver accepts) now yields `Server="tcp"` and `Name="host:3306)/"`, so `Host()` returns `tcp` and `mariadb-dump` gets the wrong arguments. Worth a `TestParse` case.

3. **I would rather take the WAL switch as its own pull request.** It is a good change and I want it, but it is a persistent, on-disk change for every SQLite installation, and it is currently arriving inside a PR whose description and acceptance criteria are about test quality. Three things need deciding in one place rather than piecemeal:

   - A leftover `-wal` is applied to a database file that was replaced underneath it. I reproduced this: with a stale `index.db-wal` present, `cp -f backup.db index.db` and reopen, and the previous run's rows are there while `integrity_check` says `ok`. That is exactly what `acceptance-sqlite-restart-%` does, and what `RestoreDBFromCache` / `NewMinimalTestConfigWithDb` do when they remove and rewrite only the `.db`. Everywhere the code or the Makefile removes or writes a SQLite database, the `-wal` and `-shm` need to go with it. `make reset-sqlite` misses the per-package ones today, and `.dockerignore` did not get the two suffixes that `.gitignore` did.
   - Where the sidecars cannot be created, the connection fails outright rather than falling back. Same directory, same database: `develop` connects and writes, the WAL DSN fails at `Ping` with `attempt to write a readonly database`, which surfaces to the user only after `connectDb`'s 60 s retry loop as "waiting for the database". A startup probe that reports the real reason, or an opt-out, would keep the NAS and network-share installs out of trouble.
   - A clean `photoprism stop` checkpoints and removes the sidecars, but a kill or a power loss does not, so `index.db` is no longer self-contained at rest. `PRAGMA wal_checkpoint(TRUNCATE)` in `CloseDb` would restore that property cheaply. Our own `backup` is fine either way, since `sqlite3 .dump` is WAL-aware.

Everything else I would rather fix myself after merging than hold this up for: the `wal_checkpoint` result being discarded (`busy=1` comes back with a nil error, so an incomplete checkpoint can be cached), `TestImportSQL`, the fixture questions, `photo_title_test.go`, a couple of assertions that cannot fail (`assert.NotNil` on a `time.Time` value, and `os.MkdirAll(..., os.ModeDir)` in `folders_search_test.go`, where the mode has no permission bits), and the doc updates the WAL default implies. Your answers on the fixture items were all I needed.

On the moments RBAR idea: yes, that shape is fine by me, and with WAL in place the locking pressure you were designing around is much lower, so do not feel you have to make each chunk tiny. One note on the guard - prefer `<>` over `>`, since a deletion can legitimately move a folder's date backwards and a `>` test would never write that case.

keif888:
I have reverted the SQLite WAL changes, and they are in a separate branch that I will submit separately after addressing concerns:
1. Config.DatabaseFile() needs to go back to a path-only derivation
2. dsnPattern should stay anchored, and name should keep accepting ?
3. I would rather take the WAL switch as its own pull request.

All other points raised have been addressed.

The MariaDB queries for moments updates have been updated to ensure extra updates aren't done.  As an FYI, MariaDB returned 0 rows updated before applying the gate on taken_max <> year/month/date for album and folder as it doesn't do the update as nothing had changed.  MariaDB is smarter than SQLite in this context.

lastzero:
Thanks! I re-checked at `486b1e37` and all three are resolved:

- **`Config.DatabaseFile()`** is back to the path-only derivation. `internal/config/config_db.go` no longer routes through `dsn.Parse().SQLiteFilename()`, so the `@`, `(`/`)`, `file:` and non-`.db` shapes all resolve the way they do on `develop` again.
- **`pkg/dsn/`** is byte-identical to `develop`, so the anchor and the `name` group are both back. Nothing further needed there.
- **WAL as its own PR** works for me. When you open it, please carry the three sub-items from my last comment with it (sidecar removal wherever a `.db` is removed or rewritten, the read-only-directory failure mode, and a `wal_checkpoint` on close), plus `.dockerignore` and `make reset-sqlite`.

That leaves the moments change. Two things I found while looking at it.

### The MariaDB "0 rows" is the protocol, not the optimizer

Both are true, but they are separate mechanisms and only one of them is the gate:

- **MariaDB.** The server does skip the row write when every assigned column already holds the value, and the client protocol reports *changed* rows rather than *matched* rows unless `CLIENT_FOUND_ROWS` is set. Our DSN params are `charset`/`collation`/`parseTime` only, so `go-sql-driver` runs with that flag off and `RowsAffected` is the changed count. That is why you saw 0 before the gate existed.
- **SQLite.** `changes()` counts every row the statement processed, same value or not - `UPDATE t SET a=1` against a row already holding 1 returns 1. So on SQLite the gate is genuinely load-bearing for the count, and without it the second pass would report every folder.

So the asymmetry is real, but it means the gate is doing something on one branch and nothing on the other.

### The MariaDB gate does not compare like with like

`p.taken_max` is `MAX(taken_at_local)`, a `DATETIME`, while `STR_TO_DATE(CONCAT(...), '%Y-%c-%e')` is a `DATE` at midnight. The `<>` is therefore true for every row whose photo has a non-zero time of day, which is nearly all of them. On the `1990/04` fixture:

```
taken_max            gate date     taken_max <> gate_date
1990-04-17 17:00:00  1990-04-17    1
```

That is *after* the album has been set to the correct 17th, so the gate matches the row on the second pass too - the `assert.Equal(t, 0, actual)` assertions pass through the changed-rows path, not through the predicate. Either compare dates on both sides:

```sql
AND DATE(p.taken_max) <> STR_TO_DATE(CONCAT(album_year, '-', album_month, '-', album_day), '%Y-%c-%e')
```

or drop the MariaDB gate entirely, since the server is already skipping the write. I would take the first, so the two branches read the same.

One thing to keep in mind either way: when the stored `y/m/d` cannot form a valid date, `STR_TO_DATE` returns `NULL`, the predicate is `NULL`, and the row is skipped - which is exactly the row you would want repaired. The common unset case is safe (`STR_TO_DATE('0-0-0', '%Y-%c-%e')` is `2000-00-00`, not `NULL`), so this is only about genuinely out-of-range values, but an `OR STR_TO_DATE(...) IS NULL` costs nothing.

### The SQLite range clip is off by one at both ends

```go
if album.AlbumYear > 1000 && album.AlbumYear < 10000 && album.AlbumMonth > 0 && album.AlbumMonth < 13 && album.AlbumDay > 0 && album.AlbumDay < 31 {
```

`< 31` excludes the 31st and `> 1000` excludes the year 1000, both of which MariaDB accepts. A folder correctly dated the 31st never satisfies the range check, so `albumDate` stays at the `1000-01-01` sentinel, the gate is always true, the `UPDATE` runs, and SQLite counts the row - so moments logs a non-zero "updated" count for that folder on every single run. `>= 1000` and `<= 31` fixes it. Same two lines in `folders.go`.

### Optional, if you feel like it

The per-folder loop costs one statement per folder plus one per folder album, all inside `mutex.Index`. You could get the same result with one `SELECT photo_path, MAX(DATE(taken_at_local)) ... GROUP BY photo_path`, compare in Go against the album/folder rows you already loaded, and issue an `UPDATE` only for the ones that differ. That drops the read side to a single query, makes the count exact by construction, and removes the sentinel-date logic above entirely. Not a blocker - the current shape is fine by me.

### Smaller cleanups

- `folders.go`: `// SQLite has potential locking issues ... on all foders at once` - typo.
- `folders.go`: `log.Errorf("album: set folder dates on %s ...")` - should be the `folders:` prefix.
- `moments.go`: both `updated %d ... dates` lines fire on every run, including the steady-state 0. Worth gating on `updated > 0`.

Once the gate comparison and the range clip are in I will test again and merge.

keif888:
MariaDB queries changed to handle DATE casting to ensure match works, 0 value handling to ensure STR_TO_DATE doesn't fail depending on sql_mode, and COALESCE to ensure that invalid dates don't prevent updates.
SQLite queries changed to 2 queries and as needed updates per call (optional suggestion).

lastzero:
Thanks - I re-checked at `55639c6a`. Everything from my last comment is in:

- **The MariaDB gate now compares like with like.** `DATE(p.taken_max) <> COALESCE(STR_TO_DATE(...), DATE('1000-01-01'))` gates correctly, and I confirmed it holds under `sql_mode='STRICT_ALL_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE'`. The `COALESCE` also closes the out-of-range case I mentioned, so the `IS NULL` disjunct is not needed. Nicely done.
- **The typo, the log prefix and the two unconditional `Infof` lines** are all fixed.

The range clip is gone because you took the optional single-query suggestion instead. That was my suggestion and I still think it is the better shape, but the rewrite needs a few guards that the set-based statement was getting for free. Four of them change what lands in the database, so I would like them on this branch rather than after the merge. All four are in the same fifteen lines.

I tested each of these on both drivers, with the same test file, so the SQLite-vs-MariaDB column below is a real difference in behavior and not a difference in setup.

### 1. The SQLite album update needs the `album_type` guard

The MariaDB statement has `WHERE albums.album_type = 'folder'`. The Go loop does not:

```go
UnscopedDb().Exec(`UPDATE albums SET album_year = ?, album_month = ?, album_day = ? WHERE album_path = ?`, ...)
```

So when `album.AlbumPath` is empty, that statement matches every album whose `album_path` is `''` - which is all of them, because `NewMonthAlbum`, `NewMomentsAlbum` and manual albums never set the column and the Go zero value is `""`. In the fixture set that is 32 of 36 albums.

Measured, with one photo added at the originals root and one folder album carrying an empty path:

```
SQLite   UpdateAlbumDates -> updated=35
         moment album "Scratch Moment": 2011/5/9 -> 2004/3/7
MariaDB  UpdateAlbumDates -> updated=2
         moment album "Scratch Moment": 2011/5/9   (untouched)
```

Both preconditions are reachable on a current install, not just on a legacy database:

- `clearDuplicateFolderAlbumPaths` ends with `UPDATE albums SET album_path = NULL ...` and runs from `Album.UpdateFolder` whenever two folder albums collide on a path. Gorm scans that NULL into the `string` field as `""`, so the row comes back from `AlbumsByType(entity.AlbumFolder, true)` with an empty path.
- `FindFolderAlbum` already carries a `Where("(album_path IS NULL OR album_path = '')")` branch for "legacy rows created before album_path was persisted", so rows with a literal `''` exist in the field too.
- A photo directly in `originals/` has `photo_path = ""`, which is the map key that makes the two meet.

Two things make this worth fixing before the merge rather than after. `FindMonthAlbum` looks up `album_year = ? AND album_month = ? AND album_type = 'month'`, so once every month album shares one date it returns `nil` for every real month and the next `Moments.Start()` creates duplicate month albums, then renames whichever one now matches through `UpdateTitleAndLocation`. And in the NULL variant it never settles: `album_path = ''` does not match NULL, so the row that triggered it is never repaired and the same mass update runs again on every pass.

Adding `album_type = 'folder'` shrinks the blast radius but does not close it, since another empty-path folder album is still a match. I would skip the empty path outright and key the update on the row you already have:

```go
if album.AlbumPath == "" {
    continue
}
...
UnscopedDb().Exec(`UPDATE albums SET album_year = ?, album_month = ?, album_day = ? WHERE album_uid = ?`, ..., album.AlbumUID)
```

`album_uid` also removes the double update when two folder albums share a path.

### 2. `time.Date` normalizes instead of rejecting, so an invalid stored date reads as correct

```go
if takenMax != time.Date(folder.FolderYear, time.Month(folder.FolderMonth), folder.FolderDay, 0, 0, 0, 0, time.UTC) {
```

`time.Date` rolls out-of-range components over rather than refusing them:

```
(2024, 6,  0)  -> 2024-05-31
(2025, 1,  0)  -> 2024-12-31
(2008, 0, 31)  -> 2007-12-31
(2020, 1, 32)  -> 2020-02-01
(2020, 2, 31)  -> 2020-03-02
```

`folder_day = 0` is not an edge case - it is the initial state of every folder row, because `NewFolder` sets year and month from the directory mtime and leaves the day alone. All four folder fixtures have `FolderDay: 0`, and `TestUpdateFolderDates` opens by asserting it.

So a folder stored as `(2024, 6, 0)` whose newest photo is 2024-05-31 normalizes onto exactly that date, the comparison says equal, and the row is skipped. Measured:

```
folder (2024,6,0), newest photo 2024-05-31
SQLite   -> 2024/6/0    (unchanged, on this run and every future run)
MariaDB  -> 2024/5/31
```

This is the same case `fa5c81bd` added `folder_day = 0 OR ...` to the MariaDB `WHERE` for. The two halves of the PR went in opposite directions on it. Comparing component-wise removes the normalization surface entirely and matches what the `SET` clause writes:

```go
if takenMax.Year() != folder.FolderYear ||
    int(takenMax.Month()) != folder.FolderMonth ||
    takenMax.Day() != folder.FolderDay {
```

### 3. The folder update needs the `root` column

`(path, root)` is the unique index on `folders`, and `AllFolders(true)` returns every root, but `UPDATE folders ... WHERE path = ?` matches on the path alone. So a path that exists under both `originals` and `import` gets both rows stamped from the originals photos, and the loop then visits the second row and writes it again. Measured on both drivers - the `import` row came out with the originals dates.

The MariaDB join has the same gap, so this one is not new, but SQLite installs never reached that code before. `AND root = ?` fixes the SQLite side; happy to take the MariaDB half separately if you would rather keep this diff small.

### 4. An unknown maximum date is written rather than skipped

```go
photoPathDates[photoPath.PhotoPath] = defaultDate  // 1000-01-01
```

MariaDB skips those rows with `p.taken_max IS NOT NULL`. The Go path puts the sentinel in the map, the comparison then differs from any real stored date, and the sentinel is written out. Measured: an album on `2020/6/1` came back as `1000/1/1`.

`MAX()` ignores NULLs, so this needs every qualifying photo in one path to have a `taken_at_local` that SQLite's `date()` cannot read - rare, but the repair direction is backwards when it happens. Leaving the key out of the map entirely gives you MariaDB's behavior, and the `ok` check already handles it.

### One transaction around the loop

Each `UnscopedDb().Exec` in the loop is its own implicit transaction, so on `journal_mode=delete` each one costs a journal create, a write, and two fsyncs. Measured here, 10,000 updates on a folders-shaped table:

```
autocommit, one statement at a time   88.96 s
same 10,000 inside one transaction     2.54 s
```

That whole time is spent holding `mutex.Index`, which photo save, batch edit, upload, import, the labels page and the marker endpoints all contend for. And the first run after upgrade is the wide one: SQLite never stamped these columns before, so essentially every folder and every folder album differs on that pass. Wrapping the loop in a single `Begin()`/`Commit()` keeps your per-row shape and gets almost all of it back. With the WAL change now out of this PR there is nothing else softening it.

### Smaller things

- `getPhotoPathMaxDates` has no test. New functions need focused coverage including unexported helpers, and the NULL branch and the parse-error branch are exactly the two paths items 4 above and the abort behavior live in.
- `log.Errorf("album: set album dates on %s ...", album.AlbumTitle)` and the folder equivalent pass the raw title. `moments.go` uses `clean.Log(a.AlbumTitle)` a few lines away - worth matching, since these lines reach the Web UI.
- If `getPhotoPathMaxDates` fails to parse one date it returns and both functions do nothing at all for that run. A `continue` would degrade to skipping one path instead. Low priority, since SQLite's `date()` returns NULL rather than an unparseable string for everything I could throw at it, including years past 9999.
- `getPhotoPathMaxDates` is the only `get`-prefixed function in the package, its doc comment is missing the closing period, and the loop variable `photoPath` actually holds a `PhotoMax`.
- `updated = 0` is redundant after the named return, there is a trailing space in the `folders.go` SQL literal, and that file now uses both `folder:` and `folders:` as a log prefix.
- The photo `GROUP BY` scan runs twice per moments run, once from each caller. Not worth restructuring for, just noting it.

### Before the merge

Your tests pass on both drivers, and so does the rest of the suite. The fixtures simply do not reach any of this - one root, non-empty unique paths, and no photo at the originals root. Two small additions would close that gap nicely: a root-level photo alongside an empty-path folder album covers item 1, and a folder left at `day = 0` covers item 2.

The rewrite is the right shape and I am glad it is in - it just moved enough surface to be worth one more pass. Items 1 to 4 plus the transaction, and I will re-test and merge. Everything under "Smaller things" I will pick up on `develop` afterwards, so no need to spend time on those here.

keif888:
**1. The SQLite album update needs the album_type guard**
I have refactored the SQLite update so that it uses album id instead of the path.  It already read only the required album_types, but used the incorrect field for the update where clause.
I have refactored the MariaDB update and the SQLite go code so that empty album_path's are excluded.

**2. time.Date normalizes instead of rejecting, so an invalid stored date reads as correct**
I have refactored the time.Date to time.Parse in both Album and Folder updates.  If an invalid date made from the 3 Y/M/D parts is found, then 1000-01-01 will be used as the comparison date to ensure that updates are made appropriately.
This has the same end result as the MariaDB query.

**3. The folder update needs the root column**
I have refactored the SQLite folder update to use the folder_uid, and the set of folders being retrieved for analysis is now filtered by the RootOriginals (in both SQLite and MariaDB), so the other roots will not be included or updated.
As an empty path is a valid location, these have not been excluded (unlike Albums).

**4. An unknown maximum date is written rather than skipped**
I have refactored the helper function to ignore the nil dates returned from the max date query.

**5. One transaction around the loop**
I have set a 5000 paths per commit limit, so that large libraries will not be locked up for the entire time that the update runs for.
This should reduce the number of writes and hopefully keep within the 5s timeout on end user hardware.
I have only desk checked this (with 2 albums, a commit count of 1 and logging statements) as I don't have a large test database to test this against.  The code was committing appropriately.

**6. Smaller things**
- getPhotoPathMaxDates has been renamed to photoPathMaxDates
- photoPathMaxDates has tests of success and forced bad date (only for SQLite)
- photoPathMaxDates will skip null dates
- photoPathMaxDates will report on bad dates and continue.  The only way for this to happen is a bug in the backend database engine.
- logging has been cleansed
- updated = 0 has been removed
- logging prefixes have been standardised
- The moments processing has not been refactored to remove the extra GROUP BY for SQLite, as that requires MariaDB (and then Postgres) to move from the SET based processing to RBAR based processing.

**7. Before the merge**
I have added a test TestUpdateAlbumDates/BlankAlbumPath, which updates an album and photo to having no path, and then confirming that the updated album is not updated, with deferred cleanups to revert the changes.
I have added a test TestUpdateFolderDates/FolderOnAnotherRoot, which creates a new folder on the Import root, and the same path as a fixture album, sets the FolderDay, saves it, and then verifies that it is not updated by UpdateFolderDates.

lastzero:
Thanks - re-tested at `27248bf1` on SQLite and MariaDB, and all four are closed:

- **Album update.** Keying on `id` and skipping the empty path is the right cut, and `album_path <> ''` keeps the MariaDB branch in step. My repro now leaves the moment album alone and reports 1 instead of 35.
- **Invalid dates.** `time.Parse` is a better fix than the range clip I suggested - it rejects instead of normalizing, so there is nothing left to reason about. Every out-of-range triple I tried falls to the sentinel, `(2024,6,0)` and `(2020,2,31)` included, which is what the MariaDB disjunct does.
- **Folder root.** `RootOriginals` on both branches plus `folder_uid` closes it twice over; the import-root folder keeps its own dates.
- **Null max dates.** Dropping them from the map is right, and `ForcedBadDate` pins it with the 36 -> 35 count.

`BlankAlbumPath` and `FolderOnAnotherRoot` would both have caught the original problems, which is the part I care about most.

Two things came in with the fixes. **Both are already fixed on our side** - you will see them in the merge, so please do not push again for them.

### `pathCount` is never incremented in `UpdateFolderDates`

`albums.go` increments after a successful update; `folders.go` declares, tests `== 5000` and resets, but never increments - so the batch commit never fires and the whole folder pass runs as one transaction, the case the limit exists to avoid. It matters at the size you were designing for: one transaction over 100k folders runs ~25 s against a 5000 ms `_busy_timeout`, so a concurrent writer errors instead of waiting. Dropping the threshold to 1 confirmed it - zero commits without the increment, thirteen with it, all subtests still passing.

### `tx.Rollback()` never returns nil

```go
if errRB := tx.Rollback(); errRB != nil {
```

Gorm v1's `Rollback()` returns `*gorm.DB` and always hands back `s`, so this fires on every error path, and `%v` formats the whole struct into a line that reaches the Web UI. `tx.Rollback().Error` is the check you want. Same line in both files.

### One test I added

Your new subtests use `(0,0,0)`, which is invalid and therefore always updates - a different path from a triple that normalizes *onto* the target. `ZeroDay` covers that one: a folder at `(1990,4,0)` with a newest photo of 1990-03-31, which `time.Date` would have resolved to the same date and skipped. It goes red if anyone swaps `time.Parse` back.

### After the merge

I have already merged this locally, so it should land on `develop` shortly, with a few refinements on top - renaming the inner type in `photoPathMaxDates` that shares the function's name, and replacing the hardcoded fixture counts in its tests. On the extra `GROUP BY`, I agree with you: keeping MariaDB set-based is worth one extra scan.

**Thanks for sticking with it through five rounds! The final changes are great, including the tests, which are much more robust now.** 🎉

PR #5789, opened 2026-08-19
Viewer: Report equirectangular frame for fisheye RAW #5711

omerdduran:
### Description

These changes fix dewarped fisheye DNGs opening in the flat viewer instead of the spherical one.

The conversion itself was never at fault — the dewarp produces a correct 2:1 equirectangular derivative with the right GPano tag, and `Panorama` is `true`. The problem was the payload `/api/v1/photos/view` hands to the lightbox, which is built from two helpers on the search result:

- `MediaProjection()` correctly reports `equirectangular`, because it prefers the dewarp derivative over the fisheye original when one is present.
- `MediaInfo()` supplies `Width`/`Height`, and its `MediaRaw` branch returned the **RAW original's** frame — 3264×6528, ratio 0.5.

`is360Equirectangular()` in `frontend/src/common/sphere.js` requires both to agree: an explicit `equirectangular` projection is authoritative only when the known frame is within 2:1 ±0.15, so that a partial or cylindrical panorama tagged equirectangular does not render as a vertically distorted sphere. The photo therefore failed the ratio test and fell back to the flat viewer. Changing the photo type from RAW to Image in the UI worked around it by taking a different branch.

This is a reporting bug only, so no re-index and no re-conversion are needed — the derivatives on disk are already correct.

#### Root Cause

`a0ef7a0c` added the `HasFisheyeOriginal()` derivative preference to `MediaInfo()`, but only to the `MediaVideo`/`MediaLive` branch. The `MediaRaw` branch predates it (added to show a RAW's native resolution in the Info sidebar) and never got the same treatment. That also explains why 360° videos were unaffected.

#### Changes

`internal/entity/search/photos_results.go` — the `MediaRaw` branch now prefers an equirectangular derivative's frame, mirroring the block the video branch already has. The existing RAW loop is untouched below it, and the primary file's hash is preserved so album cover selection and the lightbox slide key keep their current value.

Ordinary RAW files cannot be affected: the new block is unreachable unless `HasFisheyeOriginal()` is true, which requires `file_projection ∈ {fisheye, dual_fisheye}` on the primary row or a merged file. Normal DNGs carry no projection and `MediaFile.FisheyeDng()` returns false, so their code path is byte-identical. A fisheye DNG whose dewarp failed or is disabled finds no derivative and falls through to the previous behavior as well.

One intentional side effect: the Info sidebar now shows the derivative's frame (e.g. 5760×2880) rather than the RAW's 3264×6528 for fisheye DNGs, which is the correct reading for a 360° photo.

`MediaImage` is deliberately left alone. It already works deterministically rather than by accident: ExifTool normalizes an `.insp`'s codec to `jpeg`, so the `FileCodec != "jpeg"` filter always skips the original and falls through to the primary derivative.

#### Related Issues

- #5711 — Media: Render original-format 360° files (Insta360 .insv/.insp, fisheye DNG) via FFmpeg

### Testing

Four tests were added, three in `TestPhoto_MediaInfo` and one at the DTO level:

- `RawFisheyeEquirectangularDerivativePreferred` — the reported failure.
- `RawStackedEquirectangularNotPreferred` — an unrelated 360° file stacked on a normal RAW must not replace its dimensions.
- `RawFisheyeWithoutDerivative` — a failed or disabled dewarp falls back to the original's frame.
- `TestPhoto_ViewerResult_FisheyeRaw` — asserts that the DTO the lightbox actually consumes reports the projection **and** a 2:1 frame *together*. `MediaInfo()` and `MediaProjection()` are each individually defensible; only their combination in `viewer.Result` was inconsistent, so this is the test that would have caught the bug.

The new tests were verified to fail without the production change, reproducing the reported values exactly (3264×6528, ratio 0.5). The pre-existing `Raw` subtest and the two new guard cases pass both with and without the fix, confirming that normal RAW behavior is unchanged.

```
go build ./...                                        # pass
make fmt-go                                           # no changes produced
go test ./internal/entity/search -count=1             # ok
golangci-lint run ./internal/entity/search/...        # no findings in the changed files
```

Not covered here: the two sample DNGs were not exercised end to end, since they are not available in this environment and the repo has no fisheye-DNG fixture. Coverage is synthetic at the result/DTO level, which is where the defect lives. Live confirmation is a matter of opening both files in the lightbox after rebuilding.

### Acceptance Criteria

The documentation and database criteria do not apply: there are no new configuration options, no potentially disruptive changes, and no schema or query changes. The browser criterion is left unchecked because the fix has not yet been confirmed against the sample files in a running instance.


lastzero:
Thanks for the fix! 🔧 Your diagnosis holds up: `is360Equirectangular()` requires the projection and a ~2:1 frame to agree, `MediaProjection()` already returned `equirectangular`, and the `MediaRaw` arm was reporting the DNG's frame. Removing the production hunk makes exactly `RawFisheyeEquirectangularDerivativePreferred` and `TestPhoto_ViewerResult_FisheyeRaw` fail while the two guard cases pass either way, as you describe. The branch merges cleanly and the package is green.

I would like to suggest a smaller change, because I think the premise can be tightened.

### The primary file is already the derivative

For a dewarped fisheye DNG the developed JPEG is the file that carries `file_primary = 1`, and it is the file the dewarp is written to. `PhotoResults.Merge()` copies that row's columns onto the `Photo`, so `m.FileWidth` / `m.FileHeight` already hold 5760x2880 and `m.FileProjection` is already `equirectangular` before the switch runs.

So the defect is not a missing derivative lookup - it is the RAW loop below overwriting correct values with the DNG's frame. The new pre-loop searches `m.Files` for a file that `m` itself already points at.

### Why the search is worth avoiding

The returned tuple mixes provenance: the hash comes from `m`, the frame and codec from `f`, and nothing ties them together. For a still image the frame is not informational - `is360Equirectangular()` uses it to route, and `lightbox.vue` uses it to size the slide (that branch covers `media.Raw`). What is rendered comes from `Thumbs`, which the DTO builds from `m.FileWidth`, `m.FileHeight` and `m.FileHash`. When the winning `f` is not the primary, both the routing test and the slide sizing are computed against a file that is not on screen.

The video arm is a different case, and its shape is right there: a browser needs the playable file's hash, mime and codec, so substituting the file is correct. A still has no equivalent need - the browser cannot render a 360 DNG natively, which is why the JPEG exists - so there is nothing to substitute.

States where the loop picks the wrong row, all reachable with default settings:

- Stacking by matching place and time metadata is on by default, so a 360 camera writing two files for one capture merges them. An `.insv` plus a fisheye `.dng` gives one photo holding an equirectangular *video* derivative alongside an equirectangular image; `photo_type` resolves to `raw` because raw outranks video, and the loop has no `MediaType` filter, so a raw-typed result can report `avc1` / `video/mp4`.
- An `.insp` plus a fisheye `.dng` gives two dewarped JPEGs with different frames, and which one is reported depends on row order.
- `file_projection` is set from GPano metadata for any image file including a RAW, and the fisheye override only fires when the DNG is recognized as fisheye. A 360 RAW that keeps `equirectangular` is itself an eligible match, so the loop can return the RAW's frame - the value the change exists to stop reporting.

### Suggested change

```go
case entity.MediaRaw:
	// A dewarped fisheye original is displayed through its equirectangular derivative, which is
	// the primary file, so report that row rather than the RAW's own portrait frame.
	if m.HasFisheyeOriginal() && m.FileWidth > 0 && m.FileHeight > 0 &&
		projection.Type(m.FileProjection).Equal(projection.Equirectangular.String()) {
		return m.FileHash, m.FileCodec, clean.ContentType(m.FileMime), m.FileWidth, m.FileHeight
	}
```

`file_codec`, `file_mime`, `file_width`, `file_height` and `file_projection` are all in the viewer's column set, so no query change is needed. It is smaller than the current hunk, and the hash necessarily names the file whose frame is reported, so none of the states above can arise.

### Tests

Both new fixtures set `FileHash: "primary-jpeg"` while the equirectangular entry in `Files` carries a different hash, and leave `Photo.FileProjection` empty. `Merge()` always copies the seed row into `Files[0]`, so `m.FileHash` is always `Files[0].FileHash` in a real result - the fixtures describe an arrangement the search path cannot produce, and they assert the split hash and frame as the expected outcome.

Adopting the version above means reshaping them: set the photo-level columns from the equirectangular JPEG, let `Files[0]` mirror it, and keep the DNG as the second entry. A fixture in that shape passes.

### Minor

- The description says the Info sidebar will now show the derivative's frame. For a loaded record it will not - `Photo.originalFile()` returns the first non-JPEG file, so the file row still reports the DNG's 3264x6528. The DTO's dimensions only reach the sidebar on the thumbnail-model fallback.
- The note that ExifTool normalizes an `.insp`'s codec to `jpeg` holds only when ExifTool metadata was read; the native Exif reader sets no codec, so with ExifTool disabled the codec is empty and the `MediaImage` filter does not skip the original. Harmless today, since such a file is ~2:1 anyway, but worth not relying on later.
- Not required here: under `order=size` the query excludes files whose root is `sidecar`, which is where the dewarp derivative lives, so that sort keeps these photos in the flat viewer either way. A separate issue.

Issue #5770, opened 2026-08-06
NSFW: Add support for ONNX-based models


photoprismci:
**As a PhotoPrism user, I want NSFW detection to run on an ONNX-based model, so that sensitive images are recognized more reliably by a smaller and faster installation.**

PhotoPrism currently detects NSFW content with TensorFlow, while face detection already runs on ONNX Runtime. Adding ONNX support here as well is a step toward maintaining a single inference runtime instead of two.

### Why

- **Smaller installation.** The bundled TensorFlow library is large and ships in every Docker image. Once the remaining TensorFlow-based features have been migrated, it can be removed entirely.
- **One runtime to maintain.** ONNX Runtime is already included for face detection.
- **Build portability.** TensorFlow is not available on every platform. See #5591 and #5609, where PhotoPrism cannot be built while TensorFlow is a hard requirement.
- **A better choice of models.** Recent NSFW detection models are generally published in ONNX format, and several are considerably newer than what is in use today. Users who want to select a different model can then do so without converting between formats.
- **Quality and speed.** Newer models are expected to recognize sensitive images more reliably, and ONNX Runtime may be a better fit for this workload.

### Expected Behavior

The stored NSFW flag and the existing settings keep working, so no re-scan is required.

The replacement models classify an image as safe or not safe, rather than into the five categories used today. The category scores are not part of the public API, so this is not expected to affect users directly — but because a new model's scores are on a different scale, the detection threshold has to be re-checked before any default changes, so that roughly the same images keep being flagged.

### Acceptance Criteria

- [ ] NSFW detection MUST be able to run through ONNX Runtime without requiring TensorFlow.
- [ ] Existing NSFW settings and stored flags MUST keep working without a re-scan.
- [ ] The model MUST be selectable through the configuration without rebuilding PhotoPrism.
- [ ] The detection threshold MUST be re-checked against real photos before the default model changes, since scores are not comparable between models.
- [ ] Detection SHOULD be at least as reliable as the current model at recognizing sensitive images.
- [ ] Indexing time and memory use SHOULD NOT increase noticeably.
- [ ] Additional optional models MAY be offered for users who want a different trade-off between accuracy and speed.

### Related

- #3230 — NSFW filter not moving files to private
- #5591, #5609 — building PhotoPrism without TensorFlow
- #5011 — configuration of custom classification models
- #5703 — ONNX Runtime upgrade
- #5704 — optional CoreML acceleration for ONNX on macOS

Issue #5769, opened 2026-08-06
Labels: Add support for ONNX-based models


photoprismci:
**As a PhotoPrism user, I want image labels to be generated with ONNX-based models, so that I get better labels from a smaller and faster installation.**

PhotoPrism currently classifies images with TensorFlow, while face detection already runs on ONNX Runtime. Adding ONNX support for labels as well is a step toward maintaining a single inference runtime instead of two.

### Why

- **Smaller installation.** The bundled TensorFlow library is large and ships in every Docker image. Once the remaining TensorFlow-based features have been migrated, it can be removed entirely.
- **One runtime to maintain.** ONNX Runtime is already included for face detection, so labels would reuse what is there.
- **Build portability.** TensorFlow is not available on every platform. See #5591 and #5609, where PhotoPrism cannot be built while TensorFlow is a hard requirement.
- **A better choice of models.** Many more models, and more recent ones, are published in ONNX format than for TensorFlow 2. This matters most for users who are comfortable selecting a different model but not converting between formats.
- **Quality and speed.** Modern classifiers are expected to label more accurately than the current model, and ONNX Runtime may be a better fit for this workload.

### Expected Behavior

The label vocabulary stays the same, so existing labels remain valid and no re-indexing is required. The model can be selected through the configuration without rebuilding, and custom models supplied by users continue to work (see #5011).

### Acceptance Criteria

- [ ] Image classification MUST be able to run through ONNX Runtime without requiring TensorFlow.
- [ ] The existing label vocabulary MUST be preserved, so stored labels stay valid and no re-indexing is needed.
- [ ] The model MUST be selectable through the configuration without rebuilding PhotoPrism.
- [ ] Custom models supplied by users MUST continue to be supported.
- [ ] The bundled model SHOULD label at least as accurately as the current one, measured on real photos rather than on published benchmark figures.
- [ ] Indexing time and memory use SHOULD NOT increase noticeably.
- [ ] The size of the bundled model SHOULD stay the same or get smaller.
- [ ] Additional optional models MAY be offered for users who want a different trade-off between accuracy and speed.

### Related

- #448 — request for better labeling results
- #175 — scene category classification
- #5591, #5609 — building PhotoPrism without TensorFlow
- #5011 — configuration of custom classification models
- #5703 — ONNX Runtime upgrade
- #5704 — optional CoreML acceleration for ONNX on macOS
- #1287 — CLIP-powered semantic search, tracked separately and not part of this issue

PR #5809, opened 2026-09-04
Vision: Add selectable ONNX label and NSFW models

omerdduran:
### Description

This draft completes the ONNX model migration for image labeling and NSFW detection while preserving PhotoPrism's existing fixed label taxonomy and making content-screening failures explicit.

The implementation adds selectable, checksum-pinned ONNX models with model-specific preprocessing, graph validation, stable output handling, custom model support, reproducible export tooling, download metadata, runtime reporting, and opt-in corpus benchmarks. Explicit `vision.yml` disablement is preserved when `LABEL_MODEL` or `NSFW_MODEL` is `auto`.

### Label Classification

The label pipeline supports four registered ImageNet-1k classifiers through the existing label rules and embedded 1,000-class vocabulary:

- EfficientFormerV2-S1
- EfficientFormerV2-S2
- EfficientNet-B0 RA
- RepViT-M1.0

EfficientFormerV2-S2 is the bundled default. On the reviewed 402-picture Wikimedia corpus, it provided the highest visible-label coverage and left the fewest pictures without labels:

| Model | Coverage | No Labels | Mean Top-1 | ARM64 p50 |
|:------|---------:|----------:|-----------:|----------:|
| EfficientNet-B0 | 54.2% | 184 | 43.5% | 49 ms |
| EfficientFormerV2-S1 | 73.4% | 107 | 58.3% | 52 ms |
| **EfficientFormerV2-S2** | **81.8%** | **73** | **65.5%** | 80 ms |
| RepViT-M1.0 | 68.7% | 126 | 53.1% | 68 ms |

A broader local 2,500-image calibration corpus has now been prepared with 500 COCO multi-object images, 500 ImageNetV2 single-object images, 500 scene/travel images, 500 personal-photo-style images, and 500 edge or degraded images. Its files are unique, decode successfully, retain source provenance, and group derived images with their originals for validation. Source labels still need mapping and manual review against PhotoPrism's visible taxonomy before this corpus can be used for final confidence-threshold calibration.

### NSFW Detection

The NSFW pipeline supports five registered ONNX detectors:

- AdamCodd ViT-Base FP32
- AdamCodd ViT-Base INT8
- Falconsai
- Freepik
- Yahoo OpenNSFW

The result contract distinguishes `safe`, `unsafe`, and `unavailable`, so a missing model, decode failure, initialization error, or inference failure cannot silently become a safe result. Upload screening handles directly decodable images as well as generated previews for supported RAW and video media. Yahoo's BGR preprocessing applies normalization after arranging channels in the declared tensor order.

#### NSFW Calibration Results

All five ONNX candidates and the legacy TensorFlow detector were evaluated on the same 2,500 images:

- 500 manually reviewed unsafe SIMAS photographs;
- 500 manually reviewed matched-safe SIMAS photographs;
- 500 source-labeled adult illustrations;
- 500 source-labeled hentai images; and
- 500 source-labeled neutral images.

The corpus contains 1,500 unsafe and 1,000 safe images, so threshold selection uses balanced accuracy instead of raw accuracy. For each model, the selected millesimal threshold stays within 0.3 percentage points of maximum balanced accuracy while preferring recall. Five-fold group-stratified validation checks threshold stability.

| Model | Threshold | Recall | False-Positive Rate | Balanced Accuracy | ARM64 p50 |
|:------|----------:|-------:|--------------------:|------------------:|----------:|
| AdamCodd FP32 | 0.718 | **93.4%** | 15.7% | 88.9% | 465 ms |
| AdamCodd INT8 | 0.760 | 93.0% | 15.4% | 88.8% | 260 ms |
| Falconsai | 0.529 | 91.8% | **4.4%** | **93.7%** | 156 ms |
| Freepik | 0.992 | 88.7% | 5.4% | 91.7% | 979 ms |
| Yahoo OpenNSFW | 0.327 | 93.2% | 6.8% | 93.2% | **20 ms** |
| Legacy TensorFlow | 0.468 | 94.2% | 34.4% | 79.9% | 44 ms |

Yahoo remains the bundled default. Falconsai leads balanced accuracy by 0.5 percentage points, while Yahoo provides 1.4 percentage points more recall, runs nearly eight times faster on the measured ARM64 host, and uses a much smaller artifact. Every registered ONNX model now carries its separately calibrated default threshold.

The added illustration, hentai, and neutral groups are source-labeled rather than an independently reviewed final holdout. Manual review of candidate/incumbent disagreements and a final must-flag holdout remain release gates.

### Configuration & Behavior

- `PHOTOPRISM_LABEL_MODEL` selects `auto`, `none`, or a registered/custom label model.
- `PHOTOPRISM_NSFW_MODEL` selects `auto`, `none`, or a registered/custom NSFW model.
- Automatic local NSFW thresholds resolve from the selected model; explicit operator thresholds still override them.
- Named models fail closed instead of silently substituting different weights.
- Configuration, runtime reporting, model downloads, package documentation, API documentation, provenance manifests, license records, and translations are aligned.

### Related Issues

- Related to #5769
- Related to #5770

### Validation

- `make fmt-go`
- `make build-go`
- `go test ./internal/ai/nsfw -count=1`
- `go test ./internal/ai/vision -run 'TestResolvedNSFWThreshold|TestBuiltinNSFW|TestNSFW' -count=1`
- `go test ./internal/config -run 'TestConfig_(LabelModel|NSFWModel|LoadVisionConfig)$' -count=1`
- `go test ./internal/api -run '^TestNsfwRejectsUpload$' -count=1`
- `go test ./internal/api -run '^TestPostVisionNsfw$' -count=1`
- `make -C specs lint`
- Isolated SQLite CLI matrix confirms disabled default and custom label/NSFW models remain disabled with `auto`.
- A short-lived local HTTP instance started successfully and returned `{"status":"operational"}` from `/api/v1/status`.

GitHub Actions is not enabled for this repository, so validation is performed through the local project test and lint targets.

### Draft Follow-Ups

- Publish the nine registered ONNX artifacts and verify their public checksums.
- Map and manually review the new 2,500-image label corpus against PhotoPrism's visible taxonomy, then calibrate per-model confidence thresholds.
- Complete the independent NSFW holdout and manually review candidate/incumbent disagreements before release.

### Acceptance Criteria

- [x] Label and NSFW ONNX engines are implemented with selectable registered models.
- [x] Configuration, model metadata, download tooling, documentation, and API output are aligned.
- [x] Automated tests cover model validation, preprocessing, inference, concurrency, configuration, upload screening, and failure behavior.
- [x] Auto-selected default label and NSFW models retain explicit `vision.yml` disablement.
- [x] Yahoo BGR input normalization follows tensor channel order.
- [x] Registered NSFW models have corpus-calibrated default thresholds.
- [x] The previous Yahoo explicit-fixture regression passes at the calibrated default threshold.
- [ ] Registered artifacts are publicly downloadable with the documented checksums.
- [ ] The expanded label corpus is reviewed and confidence thresholds are calibrated.
- [ ] The independent NSFW holdout and disagreement review are complete.


lastzero:
Thanks for the thorough work on this - the overall design is in good shape, and a few things are clearly right and worth keeping as-is:

- The three-valued `nsfw.Result` with `unavailable` as the zero value is the right shape, and having `UnmarshalJSON` map an unrecognized status to `unavailable` rather than to a clearance is a nice touch.
- Fail-closed upload screening genuinely closes a gap: the old code errored on any non-JPEG and the handler skipped it, so PNG/HEIC/WebP/RAW/video uploads were never screened at all.
- Checksum verification is properly enforced - it runs before the runtime loads and before the graph is parsed, rather than being recorded as decorative metadata.
- `ModelInfo.Merge` is fill-only, so a model file can never override its own pinned preprocessing.
- The config plumbing follows our option checklist exactly, including `DocDefault` rather than `Value:` for a default that is resolved at runtime.
- `swagger.json` regenerates byte-identical, and `gofmt`, `goimports` and `shellcheck` are clean.

I'll help with mirroring the models on our download server and testing the thresholds against our samples so you can focus on the items below.

## Blocking

**1. The default classifier's vocabulary is no longer installed.** `classify.imageNetModel` sets `LabelFile: "../nasnet/labels.txt"`, but `dep-models` in the `Makefile` dropped `nasnet` from its list. A clean checkout now installs the ONNX classifier without a `labels.txt`, `loadLabels` fails, and labeling turns off. Existing dev containers only pass because the old `nasnet` directory is still on disk. Either keep `nasnet` in `dep-models`, or publish a vocabulary file next to each ONNX label model and point `LabelFile` at that.

**2. A legacy `vision.yml` breaks uploads permanently.** `ConfigValues.Load` disables TensorFlow models for `ModelTypeLabels` only. A `Type: nsfw` entry with a `TensorFlow:` block - which is what `photoprism vision save` used to write - survives enabled, falls through to the custom-ONNX branch looking for `nsfw/nsfw.onnx`, fails to initialize, and from then on every upload is rejected with 403 "Upload might be offensive" for the life of the process. Reproduced locally. We do want TensorFlow gone, but existing installs need a smooth path: please extend the same "migrate it to ONNX" guard to `ModelTypeNsfw`. Worth reconsidering the user-facing side too, since "the detector is unavailable" and "this image is offensive" currently produce the same 403 and the whole upload batch is deleted either way.

**3. Test suite.** `internal/api` and `internal/ai/vision` currently fail:
- `TestPostVisionLabels` and `TestPostVisionNsfw` *panic* on a nil `apiResponse.Model` instead of failing an assertion, which aborts the rest of the package and hides every test after them. A nil check would turn both into readable failures.
- `TestGenerateLabels` hard-fails when the artifact is missing, where its peers skip. Giving it the same `requireRegisteredModel` treatment keeps a fresh clone green.
- `vision_labels_test.go` relaxed `assert.Len(..., 2)` to `assert.NotEmpty(...)` for the two-image case. If two images should still produce two results, please keep the stronger assertion.
- `go vet ./internal/ai/vision` exits 1 with two copylocks findings (`Clone()` in `model.go`, and `config_test.go:381`). The `//nolint:govet` only silences golangci-lint; `go vet` still reports it. `NasnetModel.Clone()` already exists and does what the test wants.

## Should fix

**4. Converter environment in `TempPreview`.** `convert_preview.go` is the only place in the package that hands the full process environment to the child process via `append(os.Environ(), ...)`. The other seven call sites (`convert_image.go` x3, `convert_fix.go`, `convert_video_avc.go`, `convert_sidecar_json.go`, `mediafile_thumbs.go`) build from nil, so the converter receives only `HOME` and `LD_LIBRARY_PATH`. Please match that convention.

**5. `Thresholds.NSFW` changed meaning without a migration.** Before this PR the field only gated label-derived NSFW; the model path used fixed constants (0.75 on upload, 0.98 for indexing, both behind a `Neutral > 0.25` veto). It is now the detector's operating point, and `NSFWIsSet()` is just `> 0`. The old `Save()` wrote `NSFW: 75` unconditionally, so every upgraded install pins 0.75 and suppresses each model's calibrated default, while a fresh install of the same version behaves differently. The number is also not comparable across models, since it used to gate `max(porn, sexy, hentai)` from a 5-class model with a veto that is now deliberately gone. Two things would help: a way to express "unset" other than deleting the line (a sentinel such as `auto` or `-1`), and treating a loaded `75` that arrives next to TensorFlow default models as unset.

**6. `onnx.Normalization.Mean` has two contradictory conventions.** `buildBlob` in `nsfw/model.go` indexes the mean by tensor plane (`m.mean[rIndex]`) - that is the fix in 696bd2861 and it is correct for Yahoo. `buildBlob` in `classify/model.go` indexes by source channel (`m.mean[0]` for red), as does `face/embedder_onnx.go`. They agree for RGB and disagree for BGR: with `Mean: {104, 117, 123}` the classify path subtracts 104 from red and 123 from blue, exactly inverted. Latent today because all four registered classifiers are RGB, but reachable through a custom BGR label model, which is one of the acceptance criteria. `classify.TestModelBuildBlob` uses a uniform mean in its BGR case so it passes either way; the nsfw counterpart pins the convention properly and is the shape to copy. Please settle on one convention and state it in the `Normalization` doc comment.

**7. Upload screening bypasses the model's calibrated threshold.** `nsfwRejectsUpload` passes `nsfw.UploadThreshold` (0.75) as `fallback`, and in `nsfwInternal` a non-nil `fallback` suppresses `detector.DefaultThreshold()`. So a photo scoring 0.78 is rejected on upload but left public by the indexer, on the same instance with the same model. That sits oddly beside the README's own rationale that a threshold tuned for one model does not transfer to another.

**8. Omitting `UnsafeClassIndex` silently inverts a custom detector.** The field is `omitempty`, so `0` is indistinguishable from unset and validates as in-range. Someone wiring up a `[safe, unsafe]` binary graph and leaving it out gets `probabilities[0]` - the safe probability - used as the unsafe score. A `*int`, or requiring the field explicitly when `Reduction == softmax-unsafe`, would remove the trap. Related: I checked our specs, and while the per-model values are recorded in our model research data, none of `Reduction`, `UnsafeClassIndex`, `NeutralClassIndex` or `DefaultThreshold` is documented anywhere as a settable `vision.yml` field - not internally, and not in `internal/ai/vision/README.md`, whose field table lists `LabelFile` and `CanonicalOrder` and stops there. Since this PR makes all four operator-facing, they each need an entry. We can take care of the internal reference if you cover the READMEs.

**9. `auto` never checks what is installed.** `classify.AutoModelPreference` is referenced only from tests. `LABEL_MODEL=auto` and `NSFW_MODEL=auto` both resolve unconditionally to `DefaultModelName()`, with no check that the artifact is actually present. The face code already solves this: `Installed(modelsPath)` on `face.Detector` / `face.EmbeddingModel`, and `installedFaceModel()` in `internal/config/config_faces.go`, which walks `face.AutoModelPreference` and returns the first candidate on disk, falling back to `ModelNone`. Reusing that shape here - plus one error at load naming `scripts/dist/download-models.sh <name>` when the selected artifact is missing - would turn several failures that are currently silent until the first photo into one clear line at startup.

**10. `EffectiveNSFWModel` ignores `Disabled`.** It calls `vision.Config.Model(...)`, which skips disabled entries, so a disabled NSFW model falls through to `nsfw.DefaultModelName()` and `photoprism show config` reports `nsfw-model: yahoo_open_nsfw` with a real-looking path while detection is off. `EffectiveLabelModel` gets this right via `configuredVisionModel(...)` plus an explicit `Disabled` check - the same treatment would fix it. The tests mirror the asymmetry: the labels case asserts the effective value, the NSFW case only asserts the `Disabled` flag.

I'm happy to take another look once these findings have been addressed :)

omerdduran:
Thanks for the thorough review. The ImageNet vocabulary is now kept as a plain, embedded `labels.txt`, so the default classifier no longer depends on files left behind by the old NASNet installation. Custom models can still load their own vocabulary from disk.

I also added a migration path for legacy TensorFlow NSFW entries. They are disabled cleanly instead of being mistaken for custom ONNX models. Uploads now distinguish between an unsafe result and an unavailable detector: unsafe content returns 403, while an enabled detector that cannot run returns 503.

The NSFW threshold handling has been cleaned up as well. `-1` now means “use the model’s calibrated default,” and an old `NSFW: 75` setting that belongs to the default TensorFlow configuration is migrated to automatic mode. Explicit thresholds on custom ONNX models are preserved. The separate upload fallback was removed, so upload screening and indexing use the same model-specific threshold when they run the dedicated NSFW detector.

For preprocessing, normalization values are now consistently interpreted in tensor-channel order after `ColorOrder` is applied. I fixed the classifier and face paths and added non-uniform BGR tests so an accidental channel reversal is covered.

The model configuration is safer now too: class indexes are pointers, which makes an omitted value different from an explicit zero, and the required index is validated for the relevant reduction. The reduction, class indexes, and default threshold are documented in the READMEs.

Automatic model selection now checks which artifacts are actually installed, following the existing face-model approach. If no registered model is available, it resolves to `none`; if someone explicitly selects a missing registered model, the startup log includes the exact download command. Disabled NSFW models are also reported as disabled instead of falling back to a misleading default model and path.

I fixed the test issues you found, restored the exact two-result assertion, removed the `go vet` copylock findings, and made `TempPreview` use the same restricted converter environment as the other call sites. Swagger and the related documentation were regenerated as well.

Thanks again

lastzero:
Thanks - this is a big step forward, and the parts that were hardest to get right are right.

The nine model artifacts are mirrored now, so `make dep-models` works again and the suite runs: `go test ./internal/ai/...` and `go test ./internal/api` are both green, `go build` is clean, and `golangci-lint` reports nothing in the files these commits touch. I verified two of the published checksums against the pinned values by hand. That unblocks everything round 1 could not check.

On the ten items from last time: eight are fixed, two are partly fixed. Details where it matters:

| # | Item | Status |
|---|------|--------|
| 1 | Default classifier vocabulary | Fixed |
| 2 | Legacy TensorFlow NSFW entry | Fixed |
| 3 | Failing api and vision tests | Fixed |
| 4 | `TempPreview` converter environment | Fixed |
| 5 | `Thresholds.NSFW` migration | Partly - see 3 and 4 below |
| 6 | `Normalization.Mean` convention | Fixed |
| 7 | Upload bypassed the calibrated threshold | Fixed |
| 8 | `UnsafeClassIndex` omitempty | Fixed |
| 9 | `auto` never checked what is installed | Fixed, but see 1 below |
| 10 | `EffectiveNSFWModel` ignored `Disabled` | Partly - see 5 below |

Two of those deserve more than a checkmark. The vocabulary is byte-identical to the bundled derivative rather than the published 1000-class list - that distinction is easy to miss and getting it wrong would have shifted every prediction by one index, so thank you for catching which file to copy. And the normalization fix holds up under mutation: reverting all three `buildBlob` sites to the old indexing turns the new BGR tests red, which the old uniform-mean test would not have done.

## Please fix

**1. `NSFW_MODEL=auto` with no resolvable artifact now admits uploads unscreened.** This is the one to look at first, because it changes the direction of a safety control. `installedNSFWModel()` returns `none`, `applyNSFWModel` sets `Disabled = true`, `Config.Model(ModelTypeNsfw)` skips disabled entries and returns nil, and `nsfwUploadStatus` then returns `StatusSafe` - so with `UPLOAD_NSFW=false`, which is the default and means "please screen", every upload is admitted. Before these commits `auto` resolved unconditionally to the default model, `Init()` failed on the missing file, and the request failed closed. The explicit-name path still fails closed with 503; only `auto` flipped.

Two ways in, both reproduced: an assets directory populated by an older release (it has `nasnet/` and `nsfw/` but no `yahoo_open_nsfw/`), and an unreadable models directory - `fs.FileExists` is `os.Stat`, so EACCES is indistinguishable from absence, and a root-owned bind mount with a non-root PhotoPrism silently turns screening off. The boot warning from `reportUnscreenedUploads` does fire, which is a real improvement over no signal at all, but the request path logs only at Debug.

Worth separating "the operator disabled it" (`NSFW_MODEL=none`, admit) from "automatic selection found nothing" (503). Related: `vision save` persists that `Disabled: true`, and `applyDefaultModels` carries it forward, so installing the model afterwards does not re-enable screening - only an explicit `NSFW_MODEL=<name>` recovers. An auto-resolution artifact probably should not outlive the condition that produced it.

**2. A remote detector's explicit "unsafe" without a numeric score is read as safe.** In `Result.Decide`, a response of `{"status":"unsafe"}` sets the status but leaves `Score` at 0; because the status is present, the class-score fallback in the `else if r.HasScores()` arm is unreachable, so `NewResult(0, 0.75)` decides `StatusSafe`. `{"status":"unsafe","Porn":0.99}` reads safe for the same reason. Only reachable with a configured remote endpoint, so default installs are unaffected - but it inverts the verdict rather than dropping it. Reaching the class-score fallback whenever `Score` is absent, rather than only when `Status` is absent, would cover both shapes.

**3. An existing `Thresholds.NSFW: 0` changes meaning from "use the default" to "flag everything".** On develop `GetNSFW()` returns 75 whenever `NSFW <= 0`, so a hand-written `0` means "default". Now `NSFWIsSet()` is `>= 0` and `GetNSFW()` returns 0 verbatim, so `labels.IsNSFW(0)` matches any label and marks every newly indexed photo private, and the detector runs at threshold 0.0 so every upload is rejected. The `75 -> auto` migration does not cover `0`. Migrating `0` alongside `75`, or at least warning when an explicit `0` is loaded, would close the upgrade gap - the new semantics themselves look deliberate and are fine.

**4. A legacy `vision.yml` can never be set to exactly 75.** The `legacyDefaultTensorFlow && NSFW == 75 -> auto` rewrite in `Load` runs on every load, not once. An operator whose file still carries the (now disabled) TensorFlow nsfw stanza can set 74 or 76 but not 75, and nothing says why. Logging when the migration fires would make the lever visible.

**5. A disabled custom detector is still reported as active.** The `Disabled -> ModelNone` branch sits inside `if setting == ModelAuto`, so with an explicit `NSFW_MODEL=my_custom` the early return skips it: `photoprism config` reports `nsfw-model: my_custom` with an `onnx` runtime while `Model(ModelTypeNsfw)` is nil and nothing is screened. `EffectiveLabelModel` has the same shape. Moving the check above the early return, or consulting `configuredVisionModel` on both paths, would finish item 10.

**6. In a mixed batch, undecided outranks unsafe.** `StatusUnavailable` overwrites unconditionally, so a batch containing one file scored 0.99 unsafe and one RAW whose preview conversion fails answers 503 rather than 403. Nothing is admitted, so this is classification rather than safety - but 503 reads as transient and a client that retries on it will re-upload the same batch indefinitely, while the logs show an outage rather than a rejection. A determination should probably outrank the absence of one.

## Smaller things

- `Installed()` uses `fs.FileExists` in both registries, so a zero-length artifact counts as installed and shadows a complete lower-priority model. `fs.FileExistsNotEmpty` costs nothing here.
- The `Thresholds` doc comment still says "expressed as percentages (0-100)" while the default is now `-1` and `vision save` writes it. The sentinel deserves its own line on the `NSFW` field - you already had to change two tests from `Thresholds{}` to `Thresholds{NSFW: NSFWThresholdAuto}` for exactly this reason.
- `internal/ai/vision/README.md` says `-1` selects the detector's calibrated default for "both the dedicated model and the labels fast-path". It does not for the labels path: `GetNSFW()` returns 75 for any negative value, and the same applies to a remote nsfw service. Worth stating, since an operator tuning that path would otherwise look in the wrong place.
- `internal/ai/nsfw/README.md` describes flagged files being "deleted on the spot" versus a batch deletion for undecided. `removeScreenedUploads(uploads)` deletes the whole batch either way.
- `internal/ai/vision/README.md` field table omits `Path`, though `internal/ai/classify/README.md` tells operators to override it. And neither README says a custom detector that omits `DefaultThreshold` falls back to 0.98.
- Still open from round 1: the `ModelYahoo` comment calls Yahoo a "control model" (it is the bundled default now); `index_mediafile.go` still says "Classify images with TensorFlow"; `internal/ai/onnx/README.md` still says NSFW "can adopt it unchanged when it moves from TensorFlow" and omits NSFW from its consumer list and checksum table.
- `internal/ai/classify/README.md` links a path that is not part of the public repository, so it 404s for anyone reading on GitHub. Please drop that bullet.
- The new `"Content screening is temporarily unavailable"` message is not in any catalog yet - it needs a `gettext-extract` pass, ideally as its own commit.

## Tests worth strengthening

- `TestDescriptionInstalled` and `TestModelDescriptionInstalled` compute the expected value by re-implementing `Installed()`, so they cannot fail for any implementation, and with an empty models directory both sides are false and nothing is asserted. `TestConfig_installedVisionModels` is the shape to copy - it installs S1, asserts S1, installs S2, asserts S2, so it actually pins the preference order.
- Nothing covers the `screeningStatus` aggregation loop, which is where item 6 above lives.
- `Model.Clone()` moved from a struct copy to a 26-field literal. It is correct today - I checked every field - but the copy picked up new fields automatically and the literal will not, and this PR adds four fields to that struct. A reflection-based test that fills every exported field, clones and compares would keep it honest.

## One question

What motivated making EfficientFormerV2-S2 the default over S1? The commit message states the change without a rationale, and #5769 asks for the bundled model to be measured on real photos rather than published figures. S2 is roughly twice S1's compute, so the interesting number is indexing time rather than accuracy alone - if you have the comparison, it is worth recording.

## Housekeeping from our side

We merged `develop` into the branch so you do not have to: the conflicts were ours, caused by #5828 (optional CUDA execution provider) landing today, and they were all additive collisions in `internal/config` plus one `Last Updated` line. Nothing semantic changed, and the affected packages are green on the merged branch.

That merge also brings in `onnx.NewSessionConfig` / `SessionConfig.NewSession`, which now own provider selection, the CPU fall back and a warm-up check for every ONNX session. `classify/model.go` and `nsfw/model.go` still call `onnxruntime.NewSessionOptions` directly, so after the merge they are the only ONNX consumers that would not honor `PHOTOPRISM_ONNX_PROVIDER`. Routing them through the helper is the intended shape - happy to take that ourselves if you would rather not widen the PR.

omerdduran:
Thanks again for the detailed follow-up and for mirroring the model artifacts. I addressed the review items in `c556cbbba`, with the documentation and translation updates in `1a112e166` and `3a8c4fcfc`.

1. When `NSFW_MODEL=auto` cannot find an installed artifact, the default detector entry now remains enabled. Initialization therefore remains fail-closed and uploads return 503 instead of being admitted unscreened. Installing the artifact allows automatic selection to recover without persisting an auto-generated disabled state.

2. `Result.Decide` now preserves an explicit `safe` or `unsafe` result when no numeric evidence is present. If legacy class probabilities are present without a scalar score, they are used to derive the unsafe score.

3. Legacy TensorFlow configurations now migrate both `NSFW: 0` and `NSFW: 75` to the automatic threshold sentinel.

4. The legacy threshold migration emits a warning that names the migrated value, so an operator can see why `0` or `75` was changed.

5. Disabled custom label and NSFW models are now reported as `none`, regardless of whether the corresponding model option is `auto` or explicitly names the model.

6. Batch screening now gives `unsafe` precedence over `unavailable`, with direct aggregation tests for both orderings.

The smaller findings were addressed as well:

- Installed-model discovery requires a non-empty artifact.
- The threshold documentation now describes the `-1` sentinel and distinguishes dedicated detector behavior from the labels and remote-service paths.
- The upload deletion documentation now reflects whole-batch deletion.
- The missing custom model fields and default-threshold fallback are documented.
- The stale TensorFlow/control-model wording and the private classifier link were removed.
- The ONNX documentation lists label and NSFW consumers.
- The content-screening-unavailable message was extracted into the locale catalogs.
- Installed-model tests now create missing, empty, and non-empty artifacts instead of reimplementing `Installed()`.
- The screening-status aggregation has direct tests.
- `Model.Clone()` now has a reflection-based exported-field coverage test.
- The label and NSFW runtimes now use `onnx.NewSessionConfig`, including provider selection, fallback, graph inspection, and warm-up behavior.

I also completed a broader NSFW calibration pass after these fixes. All five ONNX detectors and the legacy TensorFlow detector were evaluated on the same 2,500-image corpus. Yahoo remains the default at a calibrated threshold of `0.327`: it reached 93.2% balanced accuracy and 93.2% recall at a 20 ms ARM64 p50. Falconsai reached 93.7% balanced accuracy but had lower recall at 91.8% and was nearly eight times slower at 156 ms. The per-model thresholds and methodology are now recorded in the PR description and package documentation.

Regarding EfficientFormerV2-S2 versus S1: the original decision was based on the reviewed 402-image corpus. At the same confidence setting, S2 covered 81.8% of images versus 73.4% for S1, left 73 instead of 107 images without labels, and had a mean top-1 confidence of 65.5% versus 58.3%. The measured ARM64 p50 cost was 80 ms versus 52 ms.

On the newer 2,500-image corpus, S2 measured 60.3 ms p50 and S1 44.1 ms p50. S2 again produced more visible labels, but I do not consider that sufficient evidence of higher accuracy because the expanded corpus has not yet been mapped and manually reviewed against PhotoPrism's visible label taxonomy. For that reason, final per-model label confidence calibration remains an explicit draft/release gate rather than treating agreement with the sparse legacy NASNet output as ground truth.

The focused NSFW, vision, config, and API tests pass with these changes. The PR remains a draft while the label ground-truth mapping and independent NSFW holdout review are completed.

omerdduran:
One follow-up on threshold handling after reviewing the upload and indexing policies separately: I split the operating points again, but made them explicitly configurable instead of restoring hardcoded constants.

The new `vision.yml` fields are:

```yaml
Thresholds:
  NSFWUpload: -1
  NSFWIndex: -1
  NSFWLabels: 75
```

- `NSFWUpload` controls the dedicated local or remote NSFW detector in the upload gate.
- `NSFWIndex` controls the same dedicated detector during indexing and vision-worker runs.
- `NSFWLabels` controls the NSFW confidence returned through the Ollama/OpenAI labels fast-path.
- The existing `NSFW` field remains as a backward-compatible shared fallback wherever a path-specific field is omitted.

A path-specific `-1` explicitly selects automatic behavior, even if the shared `NSFW` field is set. For a local ONNX detector, automatic mode uses that model’s calibrated default, so Yahoo currently resolves to `0.327` independently for upload and indexing. Remote dedicated detectors and the labels fast-path use `0.75` when no calibrated local model default is available.

This also makes the historical policy expressible without hardcoding it:

```yaml
Thresholds:
  NSFWUpload: 75
  NSFWIndex: 98
  NSFWLabels: 75
```

The upload handler now calls `DetectNSFWUpload`, while indexing continues through `DetectNSFW`. The Ollama/OpenAI label path uses only `NSFWLabels`; local ONNX label classifiers do not emit an NSFW signal and continue to rely on the dedicated detector.

lastzero:
Thanks for the quick turnaround and for recording the calibration and the S2 measurements - that answers the question from last time, and the numbers are exactly what we needed to see.

The branch builds cleanly, `go vet` and `gofmt` are clean, and the tests for `internal/ai/nsfw`, `internal/ai/classify`, `internal/ai/vision`, `internal/api`, `internal/config`, `internal/photoprism` and `internal/workers` all pass. `make gettext-extract` reproduces the frontend catalogs exactly.

On the round-2 items:

| # | Item | Status |
|---|------|--------|
| 1 | `NSFW_MODEL=auto` without an installed artifact | Fixed as asked - but see "Two decisions" below, we are changing the policy |
| 2 | Remote "unsafe" without a score | Fixed for the reported shapes - one small note below |
| 3 | `Thresholds.NSFW: 0` | Fixed for legacy TensorFlow files - see 2 below |
| 4 | Legacy value 75 | Fixed |
| 5 | Disabled custom detector reported as active | Fixed |
| 6 | Undecided outranking unsafe in a batch | Fixed |

The smaller items are all addressed as well: non-empty artifacts, the rewritten `Installed()` tests, the reflection-based `Clone()` test, the `NewSessionConfig` routing, and the documentation fixes.

## Two decisions from our side

**The calibrated Yahoo threshold stays the default.** Since we are leaving TensorFlow behind entirely, the default should be the operating point of the new ONNX default model, not a value carried over from the old one. `0.327` is fine.

**Uploads should not be rejected with 503.** This reverses what we asked for in round 2, sorry for the back and forth. When a file cannot be screened - no model installed, a model that fails to load, or a file with no usable preview - the upload should be admitted rather than rejected. A file that is actually decided unsafe should still be rejected with 403 as it is now. Since an unscreened upload then goes unnoticed otherwise, please make it visible instead:

- Log at Warn level when an upload is admitted without a decision (today that path logs at Debug).
- Warn at startup when automatic selection finds no installed NSFW model. `reportUnscreenedUploads` stays quiet in that case today, because the default entry remains configured.
- The model's initialization error is cached, so screening only resumes after a restart once the artifact is installed. Either retry initialization when the file appears, or say so in the log message and the README.

`ErrContentScreeningUnavailable` and its catalog entries are then no longer needed, unless you find another use for them.

## Please fix

**1. The labels auto path still persists a disabled state.** `applyLabelModel` did not get the change you made to `applyNSFWModel`: with `LABEL_MODEL=auto` and no installed classifier, it still clones the default entry with `Disabled = true`. `vision save` writes that to `vision.yml`, and after the model is installed, `registered.Disabled = current.Disabled` carries it forward, so labels stay off until the operator notices and edits the file. The NSFW path shows the right shape.

**2. An explicit `0` outside a legacy TensorFlow file still means "flag everything".** The migration now covers `0` and `75`, but only when the file carries the old TensorFlow stanza. In any other `vision.yml` - for example one written for an Ollama or OpenAI labels model - `Thresholds.NSFW: 0` (or `NSFWLabels: 0`) resolves to 0, and since `Labels.IsNSFW` compares with `>=`, every label matches. The labels fast path in `index_mediafile.go` has no `DetectNsfw` guard, so every newly labeled photo becomes private even with `DETECT_NSFW=false`. Treating `0` like the automatic value (or at least requiring a confidence above 0 in `IsNSFW`) would make this safe regardless of how the file was written.

**3. Existing `Thresholds.NSFW` values change what they control.** Before this PR the shared value only affected the labels fast path; now it is also the fallback for the dedicated detector on upload and indexing. An operator who set `98` or `100` to quiet label-based flagging now gets a detector that practically never flags (a threshold of 100 requires a score of exactly 1.0), and nothing tells them. A one-time Warn when a shared `NSFW` value other than the defaults is loaded and no path-specific field is set would be enough, or migrate a legacy value to `NSFWLabels` only.

**4. The PR description does not cover `f63d49ae8`.** The three new `NSFWUpload` / `NSFWIndex` / `NSFWLabels` fields and `DetectNSFWUpload` are operator-facing and should be listed under "Configuration & Behavior", including the fallback rules. Please also settle the draft status: the PR is marked ready for review, while the description and your last comment still call it a draft and list the label-corpus review and the NSFW holdout as open release gates. The "publicly downloadable" box can be checked - all nine artifacts are published.

## Smaller things

- `Result.Decide` treats a non-zero `threshold` as evidence that a score was sent, and a JSON `null` score as present. This only concerns remote services, which are not our main use case, but dropping `|| r.Threshold != 0` and ignoring a null score costs nothing - removing that clause currently leaves every test green.
- `internal/api/README.md` still says the upload handler runs `vision.DetectNSFW`; it is `DetectNSFWUpload` now.
- `internal/ai/tensorflow/README.md` still describes the package as used for labels and NSFW, and `CODEMAP.md` lists only `internal/ai/face` as a consumer of `internal/ai/onnx`.
- `NSFWIsSet`, `GetNSFW`, `GetNSFWFloat32`, `nsfwInternal`, `NsfwThreshold` and `NsfwUploadThreshold` are now only called from tests, and the `NSFWIsSet` doc comment no longer describes what it does. Either remove them or update the comments.
- The new helpers (`DetectNSFWUpload`, `nsfwThreshold`, `nsfwValue`, the `GetNSFWUpload/Index/Labels` getters, `resolvedNSFWThresholdFor`) need their own focused tests.

## Tests worth strengthening

We mutation-tested the new code, and these changes leave every test green:

- Removing the `if !thresholdIsSet { threshold = detector.DefaultThreshold() }` branch in `nsfwInternalContext` - so nothing checks that automatic mode actually uses the model's calibrated threshold in the detection path.
- Replacing `OnnxProvider` with `onnx.DefaultProvider` at the four model factory sites in `vision/model.go`.
- Removing the `ONNX.Input` deep copy in `Model.Clone()` - the fixture has a nil `Input`, so the reflection test cannot see it. Filling slices and pointers in that fixture would close it.
- Switching the upload handler back to `DetectNSFW`: `SetNSFWFunc` discards the threshold context, so no API test can tell which threshold was requested.

## Housekeeping from our side

The backend catalog (`assets/locales/messages.pot`) is generated in a separate step on our side, so there is nothing to do there for you.

PR #5821, opened 2026-09-15
Docker: Prefer the compose plugin over legacy docker-compose

photoprismci:
Prefers the Compose plugin (`docker compose`) over the legacy `docker-compose`
binary wherever the plugin is available, keeping `HAS_DOCKER_COMPOSE_WITH_DASH`
as the fallback for hosts that only have the old binary.

The previous order picked `docker-compose` whenever `which docker-compose`
resolved, which on a current host is often a forwarding shim around the plugin.

Resolution by environment (verified with an isolated `PATH` per arm):

| `PATH` provides | before | after |
|---|---|---|
| plugin + legacy | `docker-compose` | `docker compose` |
| legacy only | `docker-compose` | `docker-compose` |
| plugin only | `docker compose` | `docker compose` |
| neither | `docker compose` | `docker compose` |

Only the first row changes. The probe adds one `docker compose version` call per
`make` invocation, measured at 30 ms and requiring no daemon connection.


photoprismci:
Same change in the edition sub-repos: photoprism/plus#5, photoprism/pro#11. All three carry an identical block.

PR #5832, opened 2026-09-21
Media: Preserve archived Insta360 captures

### LINKED ISSUE #5829: Media: Keep a 360° picture archived when reconciling an Insta360 lens pair
**As a user who archived a 360° picture, I want it to stay archived after a rescan, so that my decision is not silently undone.**

### Expected Behavior

`reconcileInsta360Photos` (`internal/photoprism/index_insta360.go`) merges the photo rows of a separate-lens Insta360 capture (`_00` / `_10`) into one canonical picture. It combines the flags of all members, taking `photo_favorite`, `photo_private` and `photo_panorama` from any member that has them set.

Archival is combined the other way round: the canonical row keeps its `deleted_at` only when *every* member is archived, and is reset to `NULL` otherwise.

### Observed Behavior

After a merge, all files of the capture point at the canonical photo, so reconciliation is a no-op on later passes - it returns early once fewer than two photo IDs are found for those file names. It becomes active again when a second photo row owns one of those files, which is what a forced rescan can produce while the canonical row is archived. In that state `allArchived` is false and the canonical row's `deleted_at` is cleared, so a picture the user archived is listed again.

A smaller second point in the same function: `photos.photo_quality` is not recomputed on the canonical row after the other members' files move to it. The `File.RegenerateIndex()` call at the end refreshes the file sorting keys (`files.time_index`), not the photo quality.

### Technical Considerations

- The flag merge is deliberate where it ORs. The question is whether archival should follow the same rule as the other flags - stay archived when any member is archived - rather than requiring all of them.
- Reproducing this needs a real separate-lens `.insv` pair; the repository fixtures carry a single `.insv`, so this is a code-reading finding that has not been executed end to end. It was noticed while reviewing #5711 / #5716 and not filed at the time.

### Acceptance Criteria

- [ ] The behavior MUST be confirmed or refuted with a separate-lens capture: archive the merged picture, force a rescan, and check whether it returns.
- [ ] If confirmed, a merged 360° picture MUST stay archived when the user archived it, whatever the number of member rows.
- [ ] The chosen rule SHOULD be stated in a comment beside the flag merge, so its difference from the OR-combined flags is legible.
- [ ] `photos.photo_quality` SHOULD be recomputed on the canonical row after a merge, or the omission recorded as intended.
- [ ] A test SHOULD cover reconciliation of a two-member capture in which one member is archived.

omerdduran:
## Summary

- preserve an Insta360 capture archive timestamp when any reconciled member is archived
- recompute the canonical photo quality from the merged state
- cover canonical and duplicate archive cases with a valid square-video fake INSV lens pair

## Testing

- `make fmt-go`
- `go test ./internal/photoprism -run "^TestReconcileInsta360Photos$" -count=1`
- `golangci-lint run --new-from-rev=HEAD ./internal/photoprism/...`

The full `internal/photoprism` package run still encounters the existing failures in `TestFaces_migrationEmbedder` and `TestFaces_restoreEmbedder/AFailedTargetLeavesTheModelInPlace`.

Fixes #5829


lastzero:
@omerdduran Thanks for working on this! A short status update:

- We reproduced #5829 end to end: a capture file that arrives after the rest of the capture was indexed becomes a separate photo, and the forced rescan that merges the two then removes the picture from the archive. Your change keeps it archived in that case.
- One side effect came up in testing: photos removed automatically (`deleted_at` set, `photo_quality = -1`, e.g. after a purge of missing files) are restored by the indexer once their files come back. Because the quality is now recomputed unconditionally, and such members count as archived, a split capture removed this way stays in the archive instead. #5829 now lists this as an acceptance criterion.
- The root cause, files of one capture having different base names, is tracked in #5839. We plan to add `fs.StackPrefix()` so that late capture files are stacked with the existing picture under the `_00` name already stored in `photo_name`, regardless of the sequence stacking setting. Once that is in, reconciliation, and therefore this PR, only applies to captures that are already split in existing libraries.
- We still need to decide which rule archival should follow for those rows: archived if any member is archived, or decided by the photo with the lowest ID. The updated description of #5829 has the details. We will follow up with specific feedback once that is settled.

